PR #12, opened 2017-12-17
Fix the example 2 double click bug, use !== instead of !=

leeallen337:
## Why?

Issue https://github.com/google/boardgame.io/issues/9

In example 2, you can skip a player's turn by clicking on an already filled cell.

## What?
- I'm assuming that the code in `example-2.html` in the script tag is what is running the game in the example docs and not `tic-tac-toe.js`
  - In the code in the script tag it's not checking if the cell `id` is null or not, it's currently checking only if there is already a winner
  - Updated to check if the current cell is not `null`
- I also updated `!=` to `!==` in `tic-tac-toe.js` to be consistent with the next line

**Note**: I've never used docsify (which I think is what's powering the docs page?) before and I was tinkering around trying to get the docs page to run locally but I was unsuccessful, so I didn't manage to test this fix for myself. However, I did go into the codepen example linked and was able to replicate the behavior noted in the bug by only checking for a winner and not checking if the currently clicked on cell is `null`. I'd be happy to test this myself too if someone could provide some insight on locally running the documentation page 👍 

Thanks!

### DIFF
--- a/docs/react/example-2.html
+++ b/docs/react/example-2.html
@@ -73,10 +73,16 @@ const TicTacToe = BoardgameIO.Game({
 
 class TicTacToeBoard extends React.Component {
   onClick(id) {
-    if (this.props.G.winner == null) {
+    if (this.isActive(id)) {
       this.props.moves.clickCell(id);
       this.props.endTurn();
     }
+  }
+
+  isActive(id) {
+    if (this.props.G.winner !== null) return false;
+    if (this.props.G.cells[id] !== null) return false;
+    return true;
   }
 
   render() {
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -135,10 +135,16 @@ import React from 'react';
 
 class TicTacToeBoard extends React.Component {
   onClick(id) {
-    if (this.props.G.winner == null) {
+    if (this.isActive(id)) {
       this.props.moves.clickCell(id);
       this.props.endTurn();
     }
+  }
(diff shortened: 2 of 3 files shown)


leeallen337:
I signed the CLA.

nicolodavis:
Thanks for the fix! You will need to update the code in `tutorial.md`.

You can test the docs locally by running:

```
docsify serve docs
```

after installing the docsify CLI via

```
npm install -g docsify-cli
```

leeallen337:
@nicolodavis Thanks for pointing me to the `tutorials.md` file. I completely missed it. It's been updated now.

I also appreciate the instructions on running the docs locally so I was able to get that up and running and it worked for me. 👍

What are your thoughts on adding `docsify-cli` to the `devDependencies` in `package.json` and adding a `"docsify": "docsify serve docs"` script in `package.json` and updating the `README.md` with instructions? In doing so, it would also benefit locking in the version number in `package-lock.json`. If you're interested in that route, I'd be happy to make a pull request with the necessary updates:

1. Updating `package.json` and `package-lock.json` with the dependency and the script
2. Updating the README with instructions on how to run the docs locally

If not, I completely understand 👍 

Looking forward to contributing more!

nicolodavis:
That sounds great :) I think that will definitely make things easier for people that want to modify the docs in the future.

Happy to review more Pull Requests. Thanks for getting involved!

Issue #17, opened 2017-12-17
Node cannot find server file from packages


saeidalidadi:
I got `Node` module resolving error as bellow:

```
Error: Cannot find module 'boardgame.io/server'
````

xiaoyuhen:
same problem

nicolodavis:
Actually, this was broken by https://github.com/google/boardgame.io/pull/11 :) Sorry, approved that rather absent mindedly.

nicolodavis:
Should be fixed by #13

PR #20, opened 2017-12-18
fix: soft link packages to modules

saeidalidadi:
This is a fix for linking packages inside package to node_modules.
I succeed to run the project and it is so attractive :100:


saeidalidadi:
FIX: #17

Issue #7, opened 2017-11-30
Add a generic game AI


tolmar:
It's almost possible to write a generic AI using the existing framework, but there are two problems:

1. There's no way to enumerate legal moves. The ability to pass arbitrary arguments to the moves functions makes for some trouble here. Also the Tic Tac Toe example handles illegal moves by doing nothing, which is ergonomic, but the framework doesn't know it happened.
2. The framework doesn't know when a turn ends. The separate move + end turn calls are nice to make it easy to create games where one player's turn is composed of multiple logical moves, but "can I end turn" and "do I have to end turn" are questions only the caller can answer right now.

tolmar:
The easiest way to enumerate would be to add an optional "enumerateMoves" function to the Game constructor but this puts a lot of work on the user's plate. I think some sort of "these are the ranges of legal values for the arguments" system would be easier, along with letting moves return errors for conditionally illegal moves. (Checking a move then returning an error is slower and thus a generic AI will be weaker with this approach, but I think that's acceptable. People shouldn't expect the world of a generic AI.)

I don't have any suggestions for the end of turn problem.

tolmar:
For the actual AI portion I was thinking:

1. Do a generic tree search
2. Break ties randomly
3. Allow an optional heuristic function that scores the value of a non-winning board

It'll be embarrassingly bad if no heuristic function is specified, but still playable enough to debug a game. And there are relatively simple heuristics that can beat most humans at chess with a small search tree.

nicolodavis:
An optional extra function to implement sounds quite reasonable:

Something like:
(gamestate) => list of next move types with arg ranges
perhaps?

In fact, given that END_TURN can just be treated like just another move, it would fit right in.

Example:
(G) => ({
  TAKE_CARD: [list of card ids in G.board],
  PLAY_CARD: [list of card ids in G.hand],
  END_TURN: []
})

The user could optionally provide weights for these as a heuristic, and the AI code could do some reasonable things like making sure at least one move is made before END_TURN.

What do you think?

tolmar:
(G) => List of move types and arg ranges seems reasonable.

Since moves take an argument list, arg ranges need to be a list of lists. Chess example:
(G) => ({
MOVE_PIECE: [["a", 2, "a", 3], ["a", 2, "a", 4], ["b", 2, "b", 3] ... ]
})

[...]

PR #27, opened 2017-12-19
Makes winner evaluation to be part of the Game API instead of being assumed to be in the state.

Felizardo:
This will be needed for implementing miniMax later per issue #7 , given that we need a standard way to know if the game is finished or not, and who won.

### DIFF
--- a/docs/api/Game.md
+++ b/docs/api/Game.md
@@ -80,7 +80,7 @@ var game = Game({
   },
 
   victory: function(G, ctx)  {
-    return IsWinner(G, ctx.currentPlayer)) ? ctx.currentPlayer : null;
+    return IsWinner(G, ctx.currentPlayer) ? ctx.currentPlayer : null;
   },
 
   playerView: function(G, ctx) {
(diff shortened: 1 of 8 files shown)


Felizardo:
Wait, Im also a Google employee, I guess signing the CLA is redundant lol

Felizardo:
Done, signed. I signed it!

Felizardo:
@nicolodavis Let me know what you think about this change. I think there is other ways to do the same, but I think it is clearer to have it being part of the API.

We need some way of knowing that the game is over for the MiniMax, so the search can stop. In this case it would stop if winner() returns something other than null.

Federkun:
Uhm, do you think there are use case where the winner won't be the current player of that turn? If not, we can avoid writing

```
winner: function(G, ctx, id)  {
    if (theOnlyThingThatIAmInterested(G)) {
         return ctx.currentPlayer;
   }
   return null;
 }
```

each time.

nicolodavis:
Thanks for the PR @Felizardo! This looks mostly ready to be merged. I'll make a few changes closer to the weekend and get it in.

Felizardo:
Hey Frederico,

That's definitely true for most of the simple games. However, in more
complex game rules you might do some action that leads someone else to win.
For instance, you might have hidden objectives of eliminating some player
and you win even if someone else does eliminate the player for you.
Leaving the function to decide who wins gives this flexibility.

PR #37, opened 2017-12-20
Add docsify to package.json

leeallen337:
## Why?

- So `docsify-cli` is added when running `npm install`

## What?
- Added `docsify-cli` to the `devDependencies`
- Added a script in `package.json` to be able to run the docs locally
- Updated the `README.md` with instructions on running the docs
- Added `.DS_Store` to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_Store
 *.swp
 npm-debug.log*
 dist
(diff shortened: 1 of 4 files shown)


leeallen337:
If you want me to back out the addition of `.DS_Store` to `.gitignore` let me know 👍

PR #42, opened 2017-12-23
Reset tic tac toe

saeidalidadi:
I added a reset button to example game in order to  warm up myself on using the framework,
You can merge it If you think this feature would be useful.


examples/tic-tac-toe/app.test.js
```diff
@@ -15,6 +15,13 @@ Enzyme.configure({ adapter: new Adapter() });
 
 const Grid = (n) => Array(n).fill(null);
 
+const makeMove = (board, moves) => {
```
nicolodavis: Can you also use this down in line 70?

examples/tic-tac-toe/tic-tac-toe.js
```diff
@@ -51,6 +53,11 @@ export const TicTacToe = Game({
       }
 
       return { ...G, cells };
+    },
+    resetGame(G) {
+      const cells = noFill(9);
+
+      return { ...G, cells }
```
nicolodavis: Need to also reset `log` here.

examples/tic-tac-toe/app.test.js
saeidalidadi: I can use also in line 37

examples/tic-tac-toe/tic-tac-toe.js
saeidalidadi: Ok

saeidalidadi: the `log` list has not any reference inside `moves`

nicolodavis:
This doesn't work correctly once the game has been won.

Also, it doesn't reset the turn (which you would expect to go back to 0). A better way to implement this would be to save the initial state (both G and ctx), and then use the `restore()` action available in the reducer.

nicolodavis:
Going to close this for now given that it's old and out-of-sync. Feel free to re-open if you resume work on this.

PR #43, opened 2017-12-25
Documentation small typos

Ajimi:


docs/README.md
```diff
@@ -11,7 +11,7 @@
 ---
 
 The goal of this framework is to allow a game author to
-essentially translate the rules of a game to a series of
+essentially translate the rules of a game into a series of
```
nicolodavis: Can you also change the same line in the root README.md file?

docs/tutorial.md
```diff
@@ -96,7 +96,7 @@ Let's also prevent players from being able to overwrite cells.
 
 ```js
 function IsVictory(cells) {
-  // Return true if cells is in a winning configuration.
```
nicolodavis: This is actually referring to the object `cells`, so the singular *is* is correct here. Feel free to wrap cells in quotes if that makes it clearer.

PR #44, opened 2017-12-25
Minor fix

kant:


nicolodavis:
!>
is a special sequence interpreted by `docsify` to create the info box here: http://boardgame.io/#/multiplayer

Is there anything to fix?

Issue #48, opened 2017-12-27
UI toolkit


nicolodavis:
Just tracking some ideas planned for the UI toolkit in https://github.com/google/boardgame.io/projects/5

PR's are welcome for items not yet claimed. I'll try to keep the items bite sized (implementable in an hour or less), but some may be larger, and will get broken down once they start getting worked on:

- Generic token (that can be dragged and dropped)
- Generic placeholder (that can receive draggables and fire events)
- Cards (with drag-n-drop)
- Deck (with events to shuffle, draw cards, insert cards etc.)
- Dice
- Hex grid

Felizardo:
Hey @nicolodavis, I started working on chess pieces and board (which might be used by other games like checkers). Would this be welcomed too?

Felizardo:
I started this because I am porting the code I already had done for turnato to this framework.

Felizardo:
I am also thinking about adding chess as an example... It is a classic game and can illustrate some features like AI

nicolodavis:
Hey @Felizardo. Sounds good! Will add it to the list. Let's have just one generic token component (with options to style via props), though.

You can use unicode symbols for chess pieces on this generic token: https://en.wikipedia.org/wiki/Chess_symbols_in_Unicode

Felizardo:
Ok! I will send a PR soon

jchristman:
Philosophically, after we get a decent UI toolkit, we will probably want to split out into separate repos/npm modules? Like boardgame.io/ui-base, boardgame.io/ui-chess, and so on. Once this kit becomes full featured I think it would make sense to not include chess pieces in a checkers game, right? Not saying we should do this now, but possibly down the line...

nicolodavis:
I think so too. I think we should only include generic components that are applicable to a wide variety of games in our distribution:

```
Token
Placeholder
Grid
HexGrid
...
```

These can stay in this repository, but be distributed as `boardgame.io/ui` (just like we split the NPM already at the moment).

Chess is a good way to bootstrap some examples, so I'm not opposed to getting chess pieces in. However, if there is an easier way to implement chess without adding more components (just use unicode symbols perhaps?), I'd be a fan of that! Alternatively, we could move the chess piece SVG foo into the examples directory and make it clear that it's not intended to be distributed via NPM.

```
/* Unicode Chess */

<Token>♔</Token>
```

jchristman:
I’m thinking of a hierarchy of UI components that looks like this:

- GameTable has one or more of
[...]

PR #53, opened 2017-12-28
Create chess pieces.

Felizardo:
As discussed in #48. This PR only creates the chess pieces, I will send another one after this with the board.


nicolodavis:
Great work, @Felizardo! Thanks for getting everything tested as well.

Just want some changes before we can get this in:

#### 1.  The vision for tokens in boardgame.io is to have exactly one `Token` component that can accept other sub-components for rendering purposes:

The sub-component will actually be optional (the token will render as a blank square by default). This will be great for prototyping.

```
<Token render={Knight} />
```

This token component is what we will expand (eventually) to add drag-n-drop, hover animations etc. The `Knight` implementation should be fairly dumb (just rendering without any behavior attached to it).

In your case, I guess just rename `ChessPiece` to `Token`, remove anything chess-specific and move it further up the directory hierarchy?

#### 2. The user should not interact with anything SVG.

This is an implementation detail that the user should not have to worry about. I want us to have our own co-ordinate system so we can implement generic grids and hex grids. The underlying implementation can be SVG or something else, but should be hidden from the user.

```
<Grid rows={8} cols={8}>
   <Token x={0} y={0} render={Knight}/>
</Grid>
```

should create a grid and place the token on the top left square. This should also work seamlessly when using a `HexGrid`

```
<HexGrid ...>
   <Token x={0} y={0} z={0} ...>
</HexGrid>
```

Implementing `HexGrid` is outside the scope of this PR, but perhaps you can create a tiny `Grid` implementation so that you can get rid of SVG parameters from the props everywhere?

src/client/lib/chess/piece.js
```diff
@@ -0,0 +1,148 @@
...
+      x: this.props.x,
+      y: this.props.y
+    });
+  }
+
+  /**
+  * If there is a change in x and y, saves old X, Y, and current time. Starts
+  * animation.
+  */
+  componentWillReceiveProps (nextProps) {
```
nicolodavis: No space before the paren just to be consistent with the code style elsewhere. Can you also add https://eslint.org/docs/rules/space-before-function-paren to the .eslintrc file?

Felizardo:
Oh, I see, the Token component would handle all the behavior then, and the
subcomponents (King and etc) the presentation, right? That makes total
sense.

WRT grid, then I could encapsulate the <svg> element in it, for now? We
probably need to specify some attributes for rendering (width, height) of
the element too, right?

Felizardo:
Is the grid supposed to behave like the CSS grid?
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout

nicolodavis:
The Grid would be much simpler than a CSS grid. Just a matrix with rows and columns of fixed length, and an integer co-ordinate system to identify locations. Tokens map to a single location on the Grid (they cannot span multiple ones).

The Grid can encapsulate SVG elements of course. Just make sure that the external interface does not refer to anything that is specific to SVG (width and height are ok).

jchristman:
One little comment I have for Token is that I think it may be more natural for React developers for it to follow a more typical HoC usage syntax. That is, using it may be more along the lines of:

```
import { Token } from ‘boardgame.io/something’;

const options = {
  // possibly not needed
};

const Knight = (props) => (
  // any special rendering parts here. X and y and other get passed as props
);

export default Token(options)(Knight);
```

Sorry I’m on my phone so may not be entirely clear, but I mostly don’t like passing components as props...

nicolodavis:
The HoC approach sounds fine too. However, Token should be usable by itself. I fear that making a HoC will give the impression that it is always meant to be used as a wrapper around another component.

Felizardo:
Ok, I was going to suggest child-component too :)

Felizardo:
I am closing this pull request, I am going to split this code into two, one for the Token, Grid and one later for Chess example.

PR #55, opened 2017-12-29
Modify 'null' playerID functionality and refactor Tic-Tac-Toe example

jchristman:
See commit comments for more details, but I basically did the following:
- 'null' playerID on multiplayer is now treated as 'view-only' by the server
- 'null' playerID on singleplayer is now treated as 'hotseat'
- Refactored some code on the example to make it more maintainable


nicolodavis:
Nice work, Joshua!

PR #59, opened 2017-12-31
Restructured Examples App

jchristman:
I took the time to rewrite the structure of the examples app, as well as start to restructure the test cases for it. Just running the new is probably the easiest way to explain, but basically it can now have separate dropdown menus for each game in our examples (e.g. Tic Tac Toe and Chess). New directory/route structure should make this much more extensible.

This change forced me to change up the app.test.js to be able to test routes, which required a mock of HashRouter. But everything passes tests now.

I want to (tomorrow) separate out the Tic Tac Toe game testing from the app.test.js and put it somewhere else. I also need to figure out how to run multiple game servers so that we can have a chess server and tic tac toe server simultaneously.

For now though, this new structure is much better. 😄


jchristman:
Sorry about the conflicts. I probably should have rebased before starting work on this....

nicolodavis:
Directory structures looks neat!

About the UI, I think it's still nice to have single click navigation between examples without having to hover over menus. What do you think about a hierarchical menu on the left side just like in the documentation at boardgame.io?

nicolodavis:
I've rebased locally on my client. If you open up write permissions for me on your branch, I can push it back to the PR.

nicolodavis:
The one concern that I have is that the examples folder is where most people will start (I even see some modify the examples into apps they want to create). I want to make sure it stays fairly simple. I think people will find it a bit hard to dismantle the hierarchy here if they want to modify the examples directly.

The routing logic is also a bit hard to follow, and I worry we may lose some of the JS newbies. What do you think? Shall we keep the existing directory structure and do something plain and dumb for routing (just one `routes.js` with everything added there)?

jchristman:
I’m fine with whichever. I personally think this directory structure is easier to follow if we start to open up several different examples because it separates the code a bit cleaner. I admit the touting logic is complicated by the drop down and I’m fine with moving it to flat with a sidebar - should take minor tweaks with CSS to make that happen. I was going for a slightly more elegant solution to routing, but I’m fine with eliminating the transform_routes function and just having developers specify full route...

You’re the boss! Let me know what changes you’d like and I can try to get them done tonight or tomorrow. You should have write access on my branch now.

jchristman:
I also think that routing is one of the first things someone might rip out and change if hey want to build their own site. It’s the main reason app.js is now nothing but routing, so that it’s trivial to change how that works if someone wanted to add their own routing. It may still be a good idea to simplify the routing though if you think it’s a barrier to entry.

Felizardo:
Maybe it is better to avoid creating a single example with all games in it, and instead create one folder per game / example?
This has the advantage of the overhead code (i.e. routing) to be minimal and  simpler to understand for newbies. 

We can also add link on the documentation of other implementations in the wild, which would show show case more complex examples, with routes, user management and etc.

nicolodavis:
Yeah, I think let's keep everything about this PR except let's:
- simplify the routing.
- move the navigation to the left (without hover).

That should retain all the nice properties of the code separation. I personally think the routing is very elegant (and should allow us to add more sub-folders quite easily), but I think it's not worth the additional complexity for newcomers.

Thanks for taking the time on this! I really do like it. Just want to make sure we don't lose the target audience.

Also, I tried pushing my rebase but it's still giving me a "You can't push to ..." error. Hmm, not sure what's going on. For reference, I did an interactive rebase and dropped the first two commits, which cleans things up.

@Felizardo I think a single app is better than separate apps for individual examples. The ability to quickly navigate between them is a plus.

jchristman:
That is weird with the permissions - I’ll see if I can resolve the differences and reapply the diff.

Tracking the “simplify routing” — I’ll try to work on that and the menu repositioning tonight.

jchristman:
alright I got it so it will auto-merge. I'm going to care of the routing complexity now and simplify the nav menus. Hopefully have that done in just a few minutes. 😄

jchristman:
K @nicolodavis -- it's ready for round 2 of your review. I simplified routing and stole the sidebar CSS from docsify. I think it looks pretty swish now. Lemme know if you want any more changes, but it should be good to automerge since I took care of getting the other changes into my fork.

nicolodavis:
Great! Will take a look at this tomorrow.

nicolodavis:
Great work @jchristman :) Looks pretty sweet!

Please double-check if we need attribution for vue.css (and link to it somewhere in the app if necessary).

jchristman:
Will check on that tonight

jchristman:
vue.css seems to be a custom theme from docsify that mimics vuejs.org website. Docsify is MIT licensed so I think we are good...

nicolodavis:
Ok great!

PR #60, opened 2017-12-31
Adding token and grid

Felizardo:
As discussed in #48 and #53 

Creates a `<Token>` that supports animation when changing x and y. It also receives a hidden function from `Grid` that gets the underlying svg coordinates.


nicolodavis:
Looks great @Felizardo :)

PR #64, opened 2018-01-02
Update license year

benhc123:


nicolodavis:
This is unnecessary. The year is the year of first publication.

benhc123:
@nicolodavis, technically it's the year of last modification to the copyrighted article (in this case the repository).

dentemple:
Technically, a copyright notice isn't necessary at all.  Copyright is granted automatically if it's applicable to the work.

benhc123:
@dentemple You're right that copyright is often implied. But if you have a file called `LICENSE`, updating it seems reasonable. In many cases having a notice like this is a representation of the fact that you have made an effort to enforce that copyright.

PR #65, opened 2018-01-02
Change color of box next to move in GameLog 

scheijan:
Change color of box next to move in GameLog based on the player that made the move (#4).

I'm using `numPlayers` from `this.props.initialState.ctx` to get the number of players and assign a player number to each log entry. I'm not sure whether this is the right approach and I had to add an `initialState` to the test to make it work.

I also tried using `numPlayers` from `this.props.store.getState` but that doesn't seem to be available in the test either.

The CSS contains 16 colors (taken from http://clrs.cc/), so it supports up to 16 players before a color is used again.

Is this the right way or am I missing an easier way to get to that number? All feedback and comments are welcome!

### DIFF
--- a/examples/modules/app.css
+++ b/examples/modules/app.css
@@ -6,6 +6,10 @@
  * https://opensource.org/licenses/MIT.
  */
 
+* {
+  box-sizing: content-box;
+}
+
 .client {
   padding: 50px;
 }
--- a/src/client/debug/debug.css
+++ b/src/client/debug/debug.css
@@ -118,6 +118,7 @@
   height: 30px;
   line-height: 30px;
   border: 3px solid #eee;
+  box-sizing: content-box;
 }
 
 .debug-ui .player.active {


nicolodavis:
Thanks for the PR! Will take a look sometime tomorrow.

src/client/log/log.js
```diff
@@ -69,9 +72,11 @@ export class GameLog extends React.Component {
 
     for (let i = 0; i < turns.length; i++) {
       const turn = turns[i];
+      const currentPlayerNumber = i % numPlayers;
```
nicolodavis: I'm actually changing the code in the [phases](https://github.com/google/boardgame.io/issues/51) branch so that the playerID that made the move will be part of the action object. Can you wait till I get that in before we merge this?

The play order will no longer be round-robin (the user can customize it), so this code will break.

scheijan: Sure, I can wait with that. If the playerID becomes a part of the action object (and I assume is then logged with the action, too?) it will be easier to use that.

I didn't want to use the playerID to match to a certain color. That's why I used the round-robin approach. If the order of turns will be customizable I can't just use `modulo`, but with a unique playerID I could still assign colors from a list when a playerID is seen in a log for the first time.

nicolodavis: v0.16 is released, which contains the new `playerID` in each action.

scheijan: Ok, I'll take a look and send another PR.

What did you think about the color-presets in log.css? Should I keep it like that or style things in JS programmatically?

nicolodavis: Color presets in log.css look good! Maybe you can look into putting them inside SCSS variables? But even without that is fine!

scheijan: I've opened a [new PR](https://github.com/google/boardgame.io/pull/70) for this.

PR #66, opened 2018-01-03
Update License file copyright year

phillipalexander:


nicolodavis:
This is unnecessary. The year is the year of first publication.

PR #67, opened 2018-01-05
Chess example

Felizardo:
As discussed on #48 and Gitter.


Felizardo:
checkmate

jchristman:
My only issue with this (no code review yet lol) is that the GIF wastes a move in black's victory!!! The Queen could have won one move earlier.

0/10

Felizardo:
True hahahaha, but we can add to our TODO lista to refactor the GIF

examples/modules/chess/components/checkerboard.js
```diff
@@ -0,0 +1,56 @@
...
+import PropTypes from 'prop-types';
+const NUM_COLS = 8;
+const NUM_ROWS = 8;
+const PRIMARY_COLOR = '#d18b47';
+const SECONDARY_COLOR = '#ffce9e';
+const MOVABLE_COLOR = 'palegreen';
+const SELECTED_COLOR = 'green';
+const COL_NAMES = 'abcdefgh';
+
+class Checkerboard extends React.Component {
```
nicolodavis: This class assumes knowledge of the internals of the `Grid` implementation (`viewBox` specifically). It will break if we change the implementation from SVG to something else.

What do you think about just adding some props to the `Grid` class to do some of the color highlighting you need for the checkerboard (and getting rid of this class entirely)? I think they are general enough that they will be useful to a wide variety of games.

jchristman:
So I was starting to play with the Grid and Token classes last night and I think we can do a lot of style manipulation on behalf of the customer by passing values via props and using reasonable default values. I for one am on board with expanding on Grid to implement a lot of the functionality here.

Maybe Checkerboard can be a component that wraps Grid that we provide for ease and an example though? So keep Checkerboard in boardgame.io/ui and basically just have it render() a Grid with appropriate props?

jchristman:
Would make checkers trivial to implement as another example...

philihp:
@jchristman what is it that does the motion tween of the pieces when they move to their destination?

Felizardo:
@philihp That's <Token> built-in easing if you pass animate={True} to its props. 👍

jchristman:
I think you could maybe separate a bit, but for more advanced UI elements I’d expect someone might roll their own interface. This is really just for rapid prototyping.

So I think if you wanted to make Checkerboard be:

```js
const Checkerboard = (props) => (
    <Grid ...>
        <rect .../>
        ...
    </Grid>
);

To implement the board or put it all in Grid, I’m fine with either. Eventually we can replace rect with a TokenReceptor that will act as dnd destination that is agnostic to shape?

I’m really open to anything. I’m just thinking to abstract away svg know how for prototyping as it can be a bit of a challenge to get the hang of....

**Comment by nicolodavis** [comment]
General plan to wrap `Grid` in the implementation of `Checkerboard` sounds good.

@Felizardo: About what features to migrate to `Grid` specifically, maybe we can have:
- a prop to enable rendering an outline for the grid (like a chessboard where all tiles are white)
- a prop to select squares on the grid (green highlighting)

I think these are generic enough and useful for all sorts of games. For the checkerboard pattern, I don't have a particularly strong opinion.

- You could have a prop to pass in a custom colormap as a matrix to make it a checkerboard.
OR
- You could have all the black tiles actually be immovable tokens.

**Comment by Felizardo** [comment]
Ok, just to keep everybody in the loop, I am developing a Checkerboard component with its own coordinate system (following https://en.wikipedia.org/wiki/Algebraic_notation_(chess) ...) that will work as following:
```
<Checkerboard>
  <Token square="c5">
    <King color="black" />
  </Token>
</Checkerboard>
```
It is almost done, and it will be independent of Grid, although uses the same idea.

**Comment by Felizardo** [comment]
ok, please take a look now :D

File: src/ui/checkerboard.js [hunk 2]
```diff
@@ -0,0 +1,135 @@
...
+  render() {
+    // Children
+    const childrenWithExtraProp = React.Children.map(this.props.children,
+      child => {
+        return React.cloneElement(child, {
+          _coordinateFn: this.algebraicCord.bind(this)
+        });
+      }
+    );
+    return (<svg viewBox={'0 0 ' + this.props.cols + ' ' + this.props.rows}
```
nicolodavis: Can we use a `Grid` here instead of SVG directly?

src/ui/checkerboard.js
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * Usage:
+ *
+ * <Checkerboard>
+ *   <Token square={'c5'}>
+ *     <Knight color='dark' />
+ *   </Token>
+ * </Checkerboard>
+ */
+class Checkerboard extends React.Component {
```
nicolodavis: Can we keep `Checkerboard` in the examples section? I don't think it will be useful to games beyond chess and checkers (i.e. I don't think it should be distributed).

src/ui/checkerboard.js
Felizardo: The issue with using ```Grid``` is that this implementation of Checkerboard has its own coordinate system (implementing _coordinateFn), and ```Grid``` has another implementation for _coordinateFn. Because the coordinate system uses ```<Token/>``` props, we can pass only one or another.

I prefer to create a coordinate system using "a1" notation because it encapsulates the positioning of the elements on the screen totally on Checkerboard, and the board.js does not need to know anything about it (just the position like "c5").

src/ui/checkerboard.js
Felizardo: Sure

src/ui/checkerboard.js
Felizardo: @nicolodavis let me know if you have some idea on how to make this better, im happy to change

nicolodavis:
I added a commit as a suggestion for what I meant by using `Grid` inside `Checkerboard`. Let me know what you think (we can revert if you don't like it).

`Checkerboard` basically adds the pattern and also does the square <=> x,y conversions and `Grid` does the SVG stuff.

PR #71, opened 2018-01-08
update documentations for imports

philihp:
Updates a few of the documentation pages.

* In `Game` has two examples of usage. Merges them into one.
* In `Game`, switch the import from `game` to `core`.
* In `Server`, use an `import` instead of a `require`.
* Removes a link to localhost
* Red PR


nicolodavis:
Thanks :)

PR #74, opened 2018-01-13
Card and Deck 🃏

chrisheninger:
Started playing with adding a deck component to the UI and spruce up the card component.

This is all a work in progress– I was using storybook to start thinking out ideas so I've left tests broken (I think).

Here's a screen capture:
boardgameio-demo


nicolodavis:
Wow, this looks absolutely fantastic!

What do you think about minimizing the logo at the back of the card to just the "IO" inside the hexagon? As much as I'd love the attribution, I feel like it might the first thing that people would want to strip out.

Maybe just have the hexagon with "IO" right in the middle of the card so that it's also nice and symmetric?

chrisheninger:
That sounds good, I got a little carried away. 😅

I'm gonna give this a pass cleaning it up– curious if you have any recommendations for the API. After I sketched this out I realized what you were asking for originally is a component that can be used by any game– so I want it to adjust it to be customizable in that way. Here's what I'm thinking– I'm not set on any of the prop names and would gladly take advice/direction:

```
<Card
  front={<div>Ace ♥</div>} (defaultProp will be the word "card")
  back={<div>Logo</div>} (defaultProp will be IO logo)
  isFaceUp={false}
/>
```

```
<Deck
  cards={[]}
/>
```

How should we handle the card flipping and deck draw– should they be controlled or uncontrolled components? (Do we need to designate a click handler for flipping the card along with passing the value of isFaceUp, or should the card/deck handle that itself?)

nicolodavis:
- `front`
- `back`
- `isFaceUp`

sound good for props to start with! `onClick` as well? (users can use that to flip cards).

About deck draw, I think it's ok if the deck depletes by one when it is clicked. We can also provide an `onClick` handler that receives as an argument the card that was ejected if the user wants to add it to another hand or deck.

chrisheninger:
Made a few more changes and cleaned things up– still need to write some tests for deck and maybe add a few more for card? I'd love your input there– writing tests is not my forte.

bgio-card-and-deck

Give the code a look and let me know what you think. I feel like it's easy to get opinionated about functionality– hoping that by keeping them simple it'll be a good base that people will be able to customize from.

src/ui/deck.js
```diff
@@ -0,0 +1,39 @@
...
+const Deck = ({ cards, ...rest }) => (
+  <div className="bgio-deck" {...rest}>
+    {cards.map((card, i) =>
+      React.cloneElement(card, {
+        key: i,
+        className: i > 0 ? "no-hover" : "",
+        isFaceUp: i === 0, // Only the top card should ever be face up
+        style: {
+          position: i ? "absolute" : "inherit",
+          left: i * 3,
```
nicolodavis: Let's add a `splayWidth` prop to `Deck` to allow the user to adjust this magic number `3` here? That will also allow the user to disable splay altogether by setting it to `0` if they desire viewing just the top card and nothing else.

src/ui/deck.js
```diff
@@ -0,0 +1,39 @@
...
+import React from "react";
+import PropTypes from "prop-types";
+import "./deck.css";
+
+const Deck = ({ cards, ...rest }) => (
+  <div className="bgio-deck" {...rest}>
+    {cards.map((card, i) =>
+      React.cloneElement(card, {
+        key: i,
+        className: i > 0 ? "no-hover" : "",
```
nicolodavis: Let's make `canHover` a prop of `Card` (i.e. don't set class elements on `Card` from `Deck`).

storybook/DeckStory.js
```diff
@@ -0,0 +1,50 @@
...
+    super(props);
+
+    this.state = {
+      deck: standardDeck
+    };
+  }
+
+  onClick = () => {
+    const deck = [...this.state.deck];
+    const card = deck.shift();
```
nicolodavis: I think this behavior should almost certainly be a part of the `Deck` class because it's a very common use case. Maybe we can expose a prop `onClick` in `Deck`? The behavior for when a user clicks on a `Deck` could be:
- remove the top card
- invoke `onClick` and pass the card that was ejected as an argument

chrisheninger: Excellent, I like that. I can make an example of a card being drawn from the deck, as well as being put back into the deck.

chrisheninger:
Hey @nicolodavis I haven't had much time to work on this– I'm starting a new job and giving a couple of meetup talks next week. :)

I pushed what I have– but the deck class is a bit weird still. It currently takes in `cards` when it is instantiated– but there isn't a way to update the cards to add new ones back into the deck or whatever.

I probably won't be able to get back to this for about two weeks– feel free to pick up where I've left off it wanted sooner!

nicolodavis:
Hey @chrisheninger! I'm a bit busy myself, so take your time with it. I'll take it over if I get more time sooner.

Good luck with the new job and talks!

chrisheninger:
Was trying to do some work on this– things got a little out of hand when I merged the latest stuff down.

I'm not sure what I messed up– but my Deck Story has a memory leak and will crash the browser tab. 🙈 I'll have to figure it out next time I have a chance to work on this again.

chrisheninger:
Okay, fixed. 😅 

Turns out that the storybook action logger doesn't handle logging massive objects properly– JSON.stringify did the trick.

I think this should be ready for review again!

nicolodavis:
Great! Will take a look this weekend.

chrisheninger:
👍 I'll be around off-and-on if anything needs updating or fixing! Still open to ideas/changes. :)

nicolodavis:
Great job on this @chrisheninger :)

We can keep tweaking the deck API, but this is a solid start.

nicolodavis:
Also, this is my fault, but do you think the "IO" on the hex looks like a 10? Do you think we should remove it and have a blank hexagon instead?

chrisheninger:
:clap: Thanks for the help! I enjoyed getting some experience contributing to a lib like this– the majority of my time is spent on the app-building side. 🙂

I thought the same thing about the IO / 10– I think using a blank hex is a good idea.

Redmega:
Would it be better to have this api? `<Deck><Card /> <Card /> <Card />...</Deck>`

in other words, use `children` instead of a prop?

Redmega:
Also, I believe a `spread` prop is called for:
`<Deck spread />` to display it spread out like in the storybook. 
The default should be all on top of eachother.

nicolodavis:
@Redmega Is there a reason you think using children is better than a prop? I'm open to either option, as long as we understand what the benefit is.

chrisheninger:
@Redmega regarding the `spread` prop– it exists but is called `splayWidth`

PR #75, opened 2018-01-13
Offer a standard StripSecrets for playerView

philihp:
Adds a game Liar's Dice to the example demos, which shows that PlayerView will strip the secrets from a game's state for a player.

Also, makes a change to the local Client, which automatically applies PlayerView, even if the default `(G) => G` reducer is used. I guess this is a bug, but not really becauase I'd imagine most of the time you only care to do client secret stripping while in development.

Note: I didn't finish implementing the actual Liar's Dice game. It makes for a good example to demo 
https://github.com/google/boardgame.io/issues/72, but would also work for a demo of https://github.com/google/boardgame.io/issues/68 as well, when the interface for stochastic state is figured out.


nicolodavis:
Looks great @philihp! Thanks for the PR :)

PR #81, opened 2018-01-16
Add tests for custom turn order

philihp:
* First test (for Agricola, etc) tests that the framework calls a custom `turnOrder.first` function.
* Second test (for reversing turn order on phases) tests that the framework calls a custom `turnOrder.next` function.


src/core/flow.test.js
```diff
@@ -238,6 +238,155 @@ test('pass', () => {
...
+  let state = {
+    G: {
+      startingPlayerToken: 0
+    },
+    ctx: flow.ctx(numPlayers)
+  }
+
+  expect(state.ctx.currentPlayer).toBe('0')
+
+  state = flow.processGameEvent(state, { type: 'endTurn' });
```
nicolodavis: Please change `{ type: 'endTurn' }` to `gameEvent('endTurn')`. Similarly below.

philihp:
cool, updated. Didn't notice game.test.js, so I moved it into there - testing this feels a lot more fully integrated.

PR #82, opened 2018-01-17
persist game state inside db

saeidalidadi:
This is very simple without the tests, please check and let me know about your consideration. thanks.

### DIFF
--- a/src/server/db.js
+++ b/src/server/db.js
@@ -22,8 +22,11 @@ export class InMemory {
    * @param {string} id - The game id.
    * @param {object} store - A Redux store to persist.
    */
-  set(id, store) {
-    this.games.set(id, store);
+  set(id, state) {
+    return new Promise((resolve) => {
+      const result = this.games.set(id, state);
+      resolve(result);
+    })
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/server/db.js
```diff
@@ -22,8 +22,11 @@ export class InMemory {
    * @param {string} id - The game id.
    * @param {object} store - A Redux store to persist.
    */
-  set(id, store) {
-    this.games.set(id, store);
+  set(id, state) {
```
nicolodavis: Can these be made [async](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) functions to avoid having to manually create promises?

src/server/index.js
```diff
@@ -29,57 +29,61 @@ function Server({ games, db }) {
...
-      socket.on('action', (action, stateID, gameID, playerID) => {
-        const store = db.get(gameID);
 
-        if (store === undefined) {
-          return { error: 'game not found' };
-        }
-
-        const state = store.getState();
+      socket.on('action', (action, stateID, gameID, playerID, numPlayers) => {
+        db.get(gameID).then(state => {
```
nicolodavis: Change this to `await` after making `db.get` an async function?

src/server/db.js
saeidalidadi: I did this with `async` but I got an issue with class methods, I should should find the `babel` config for `async` methods

nicolodavis: Alternatively, don't use classes at all.

PR #86, opened 2018-01-18
make db interface async

saeidalidadi:


nicolodavis:
Sorry, I haven't had time to look more closely at this. Are you stuck trying to fix the tests?

src/server/index.test.js
```diff
@@ -67,19 +68,21 @@ test('sync', () => {
...
   io.socket.receive('sync', 'gameID');
+
   expect(io.socket.emit).toHaveBeenCalledTimes(2);
   expect(spy).not.toHaveBeenCalled();
 
   spy.mockRestore();
 });
 
-test('action', () => {
+test('action', async () => {
```
akaguny: async not needed, because not await exist

saeidalidadi: Agree, but I forgot to remove it after I clean `await` s when it didn't work

src/server/index.test.js
```diff
@@ -67,19 +68,21 @@ test('sync', () => {
   // Sync causes the server to respond.
   expect(io.socket.emit).toHaveBeenCalledTimes(0);
   io.socket.receive('sync', 'gameID');
-  expect(io.socket.emit).toHaveBeenCalledTimes(1);
```
saeidalidadi: Wrapping this assertion inside a timeout works fine but I think there should be a better solution for `async` behavior of `socket` callbacks

nicolodavis: Just use `await` in the tests. The trick is to make the function explicitly have a `return` statement. Then you can use `await` to wait for its execution and then test what it did. In this case, I wrapped `receive` in this way to make the tests to work.

package.json
```diff
@@ -62,6 +62,7 @@
     "babel-plugin-external-helpers": "^6.22.0",
     "babel-plugin-module-resolver": "^3.0.0",
     "babel-plugin-transform-es2015-modules-commonjs": "^6.26.0",
+    "babel-polyfill": "^6.26.0",
```
nicolodavis: Is it possible to do this without `babel-polyfill`? Modern versions of Node support async / await out of the box.

saeidalidadi: After running examples I got `regeneratorRuntime` error.
I tried to solve it by testing all options from [babel docs ](https://babeljs.io/docs/plugins/transform-regenerator/) and finally this one worked.

nicolodavis: babel-polyfill shouldn't be used in a library, so we should find another way to do this. Have you tried babel-preset-env?

nicolodavis: The other option would be to not use Babel on anything in src/server and write it without ES5 features. I can take a look at this later.

nicolodavis:
I think this is working correctly now!

I want to also tackle #101 either as part of this PR or get that in first before I merge this.

PR #88, opened 2018-01-19
Automatically ending turn when game ends, fixing #80

### LINKED ISSUE #80: Last turn not captured in log
Suggestions:
- Also end the turn when `endGameIf` returns true.
OR
- Make `GameLog` add an entry on `gameover` in addition to the turn end.

Felizardo:
Fixes #80


src/core/flow.js
```diff
@@ -471,10 +478,6 @@ export function FlowWithPhases({
       state = dispatch(state, { type: 'endPhase', args: [end], playerID: action.playerID });
```
Felizardo: I wasn't sure if (for the sake of consistency) we should end the phase too... Let me know if you think it is a good idea, I can change the code.

nicolodavis:
Thanks for the fix, @Felizardo!

PR #89, opened 2018-01-22
Logging x, y, z click on the HexGrid storybook example.

Felizardo:
Im trying to learn this x, y, z for hex coordinate system and It helps to see the coordinates when you click :).

### DIFF
--- a/storybook/hex.js
+++ b/storybook/hex.js
@@ -8,6 +8,7 @@
 
 import React from 'react';
 import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
 import { withKnobs, boolean, number } from '@storybook/addon-knobs/react';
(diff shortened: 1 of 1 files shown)


Felizardo:
Done

Felizardo:
Btw, because it is a object, you need to click to expand it (the one-line summary only shows [Object]) on Storybook. :man_shrugging:

nicolodavis:
Feel free to change it back to the string if you think it will be easier to read than the object.

Felizardo:
I think it would be easier to read, but at the same time more confusing, because that wouldn't be the actual argument of the onClick function. So Im ok leaving the way it is.

nicolodavis:
Also, you can learn more about the co-ordinate system here: https://www.redblobgames.com/grids/hexagons/

Felizardo:
Ok, agreed. I will add JSON.stringify then.

PR #90, opened 2018-01-22
resolve #85 add option in Client config to specify socket server

### LINKED ISSUE #85: Add option in Client config to specify socket server
Currently you can only wire it to the same server that serves up the webpage.

An option like:
```
Client({
   game: TicTacToe,
   multiplayer: { server: 'hostname:port' },
})
```

would allow serving the page and game logic from different servers. Might also help the `create-react-app` users.

aidenbenner:
Adds an option to the client constructor to specify the port used for multiplayer games. Option works just like the example in #85 e.g. 
```
Client({
   game: TicTacToe,
   multiplayer: { server: 'hostname:port' },
})
```
First PR with a real repo open to feedback.


nicolodavis:
This looks great! Will take a closer look near the weekend. Thanks for the PR.

Issue #91, opened 2018-01-26
Prevent endTurn until movesPerTurn is satisfied


nicolodavis:
There is a mechanism to end the turn automatically after a number of moves have been made: https://github.com/google/boardgame.io/blob/master/src/core/flow.js#L83

We'd like to extend this so that the turn cannot be ended earlier (via `endTurn`) until these many moves have indeed been made.

akaguny:
You whant to add any conditional to https://github.com/google/boardgame.io/blob/master/src/core/flow.js#L131?
any :
```javascript
let acceptToEndTurn,
     G;
acceptToEndTurn = movesPerTurn && ctx.currentPlayerMoves >= movesPerTurn;
if(acceptToEndTurn !== false){
  G  = onTurnEnd(state.G, state.ctx);
}
```

karel1980:
A practical solution is to just disable endTurn calls, and implement endTurn as a move, which sets a flag on the game state. The flag would be picked up by endTurnIf

nicolodavis:
@akaguny Yes, something similar to that.

@karel1980 Disabling state changing signals like `endTurn` and `endPhase` is a good idea too. Perhaps we could have an option like `allowedSignals` (similar to `allowedMoves`) to specify which in `endTurn` and `endPhase` is allowed.

No reason not to have both! Would you guys be interested in sending me PR's for your respective suggestions? I'll eventually get down to them if you haven't got the time.

akaguny:
@nicolodavis i want to make PR if @karel1980 don't worry for this, because it`s him bug)

karel1980:
@nicolodavis you argument makes sense. I don't mind sending a PR either.

karel1980:
@nicolodavis @akaguny Please see comments on the PR, there's still a few concerns, but it's a step in the right direction I think.

akaguny:
@nicolodavis, i think that idea about allowedSignals not good because programmist must think and prevent manual execution of the endturn signals by heself, and if endTurn can be execute by player he must disable this feature in interface

nicolodavis:
@akaguny I disagree. I personally think that users will find it surprising that they need to enable a feature like ending a turn before being able to use it (and will probably trip a lot of people learning the library). It feels more like a feature for an advanced user, so I'm leaning toward making it opt-out (i.e. all events are enabled by default, and we provide a mechanism to disable them).

akaguny:
[...]

PR #94, opened 2018-01-27
Added test for allowEndTurnIf on SimpleFlow

karel1980:
A first stab at #91 to see if you think I'm going in the right direction.
- I'm only testing SimpleFlow. The fact that it has to be implemented both in simple flow and in phase flow is a code smell. Isn't simple flow just a special case of phase flow (where there is only one fixed phase)? Room for simplification here.

- I added a simple check to tic-tac-toe to force at least one move. The flow tests seem very low level. I had to resort to manually testing with tic tac toe to be confident that it worked.

- This is not covering everything yet. You can still specify an invalid move (e.g. clickCell(999) in tic-tac-toe, and endTurnIf will automatically end the turn.

- How should the system behave if endTurnIf and allowEndTurn contradict each other? (maybe that's what happening in tic-tac-toe right now.

### DIFF
--- a/src/core/flow.js
+++ b/src/core/flow.js
@@ -101,6 +101,8 @@ export function Flow({ ctx, events, init, validator, processMove }) {
  *                               Whenever `condition` is true the `action` is run.
  *                               Triggers are processed one after the other in the
  *                               order they are defined at the end of each move.
+ * @param {...object} allowEndTurnIf - Turn not allowed to end if this returns true.
+ *                                     (G, ctx) => boolean
  */
 export function SimpleFlow({
   movesPerTurn,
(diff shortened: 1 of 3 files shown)


karel1980:
I'll get the CLA in order

karel1980:
I signed it!

src/core/flow.js
```diff
@@ -101,18 +101,22 @@ export function Flow({ ctx, events, init, validator, processMove }) {
  *                               Whenever `condition` is true the `action` is run.
  *                               Triggers are processed one after the other in the
  *                               order they are defined at the end of each move.
+ * @param {...object} allowEndTurnIf - Turn not allowed to end if this returns true.
```
nicolodavis: Hmm, I think this is going to make the semantics a bit confusing. Like you mention, what happens if this conflicts with `endTurnIf`?

How about just using a simple whitelist similar to `allowedMoves`. In the flow section, you can have:

```
flow: {
  // default value if not provided.
  allowedEvents: [ 'endTurn', 'endPhase', 'pass'],
}
```

karel1980: Yes the conflict with endTurnIf is a problem. Simply whitelisting would work.
For my use case the 'endTurn' event is only allowed after specific moves have been made,
but I can make it work by (a) disabling it using allowedEvents on flow and (b) implementing a move which sets a flag on game and (c) endTurnIf picks up the flag.

nicolodavis:
@karel1980 As part of getting rid of `SimpleFlow` (and make it a special case of `FlowWithPhases`), I actually had to implement the event whitelist as well: https://github.com/google/boardgame.io/commit/1abc042de2ff9192a54ab15b0d1fdb8ebf3948ee

Closing this PR if that commit subsumes this (I'll release it in the next version of the NPM soon). Feel free to re-open if it doesn't fix your problem.

PR #100, opened 2018-02-07
Allow auth creation of game rooms. 

JoshuaStrunk:
This adds a rest like API call to the server app for creating game rooms. Post:'/api/createGame'

This does not actually feel like the best way to get this functionality into this project, but wanting to open up and formalize the discussion on how it could be. One major drawback is this implementation requires playerIDs to be unique across all games and rooms. 

A better approach would be having some system for developers to script this sort of createRoom and validation check logic themselves. I am not sure how that would look though.


JoshuaStrunk:
I signed it!

nicolodavis:
Sorry about the delay getting to this! In the middle of moving. Will send you some comments in the next couple of days.

src/server/index.js
```diff
@@ -58,8 +72,10 @@ function Server({ games, db }) {
           const state = store.getState();
 
           // Get clients connected to this current game.
-          const roomClients = roomInfo.get(gameID);
-          for (const client of roomClients.values()) {
+          const roomClients = Array.from(roomInfo.get(gameID))
+            .map(playerID => playerInfo.get(playerID))
+            .filter(socketID => socketID !== undefined);
```
nicolodavis: Checking for `undefined` is probably unnecessary after https://github.com/google/boardgame.io/commit/a768f1f8a8775360b354f8475da77b96ca029535

src/server/index.js
```diff
@@ -80,16 +96,26 @@ function Server({ games, db }) {
       });
 
       socket.on('sync', (gameID, playerID, numPlayers) => {
```
nicolodavis: I think we should use some sort of `userID` here instead of `playerID`. `playerID` should typically be '0', '1', '2' and so on (some of the turn order logic even relies on this, but will be made more generic eventually).

Each game has a fixed set of `playerID`'s that don't change. The model I have in mind is:
- `userID`'s can be associated with these `playerID`'s for a particular game in `createRoom`. This mapping can even be dynamic (users dropping in and out of games, or being replaced by bots).
- `userID` can be associated with a client similar to the way `playerID` currently is [here](https://github.com/google/boardgame.io/blob/master/src/client/client.js#L66). You probably need to send a secret along as well so that we can be sure that that client is indeed the `userID` it claims to be.
- Unauthenticated games can just send a `playerID` and things should work as before.

src/server/index.js
JoshuaStrunk: I will look at that commit more, but my quick glance confirms this check is now unnecessary.

src/server/index.js
JoshuaStrunk: This sounds good. Just got back from a trip so will dig more into these changes later this week.

JoshuaStrunk:
One other change I think would be good for is removing all of the HTTP request handling code from server/index.js and instead, have server expose a createGame function. Developers could then use whatever method they want to create authenticated games. Any thought on how to go about this? Attaching it to the returned Koa app seems wrong. Would attaching it to an object passed in be bad form? Should the server be refactored to be its own object or can a blank object be passed in and the server interface attached to it?

nicolodavis:
In general, I'd like the server code to be opinionated (zero config), but with an easy mechanism to override our defaults. I think attaching to the Koa app is fine as long as we allow the user to override that. For example, we can accept a second argument that (if `undefined`) can be the object that the handlers are attached to instead. I don't think we need a full blown server interface yet.

nicolodavis:
Closing this due to lack of activity. Feel free to reopen if you pick it up again.

PR #103, opened 2018-02-08
Random API

Stefan-Hanke:
I'm using TODO comments to show that the code at hand needs more consideration. I hope this is OK.


Stefan-Hanke:
I've signed the CLA.

src/server/index.js
```diff
@@ -65,8 +65,11 @@ function Server({ games, db }) {
           const roomClients = roomInfo.get(gameID);
           for (const client of roomClients.values()) {
             const playerID = clientInfo.get(client).playerID;
+
+            let newCtx = { ...state.ctx, seed: undefined };
```
nicolodavis: just inline this below?

src/server/index.js
```diff
@@ -99,8 +102,10 @@ function Server({ games, db }) {
           await db.set(gameID, state);
         }
 
+        let newCtx = { ...state.ctx, seed: undefined };
```
nicolodavis: same here

src/server/index.test.js
```diff
@@ -57,7 +57,11 @@ jest.mock('koa-socket', () => {
   return MockIO;
 });
 
-const game = Game({});
+const game = Game({
+  setup: () => {
+    return { seed: 0 };
```
nicolodavis: We should probably pass seed through a separate option in `Game` rather than through `G`.

src/core/random.js
```diff
@@ -0,0 +1,105 @@
...
+          // TODO ignore? log?
+          break;
+      }
+    });
+  }
+
+  return { randomresults, ctx: ctx2 };
+}
+
+function runrandom(G, ctx) {
```
nicolodavis: You will need to call this somewhere in `flow.js` right after a move is processed.

src/core/random.js
```diff
@@ -0,0 +1,105 @@
...
+ * Please note this is an in-place shuffle. It will modify the given array.
+ * If you want to retain the original order, clone the array before.
+ *
+ * This code is taken directly from Mike Bostock, with a few slight modifications
+ * https://bost.ocks.org/mike/shuffle/
+ *
+ * @param {...object} ctx - The context object keeping the seed and current state.
+ * @param {any[]} array - The array to sort. Will be modified in-place.
+ */
+function shuffle(ctx, array) {
```
nicolodavis: If this is meant to be a public API, then it probably should look similar to `rolldie` below (i.e. only place a request to shuffle in `G`).

src/server/index.js
Stefan-Hanke: `newCtx` is used twice, since calling into `playerView` also needs the `ctx`. I've changed the code to use the old `state.ctx` to use in `playerView`, and inlined the code.

src/server/index.test.js
Stefan-Hanke: I was unsure about this one. The `seed` is now passed as separate option.

src/core/random.js
Stefan-Hanke: I've currently dropped shuffle for now, since this is something that could be done via `random` + a kind-of _seeded_ shuffle implementation.

src/core/random.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Add a request to Roll a D6 die.
+ *
+ * The result is put into the given fieldname and is available
+ * on the next move.
+ *
+ * @param {...object} G - Game instance.
+ * @param {*} fieldname - Fieldname to put the random result into.
+ */
+export function rolldie(G, fieldname) {
```
nicolodavis: You'll need to import these in `packages/core.js` and `packages/main.js`. I have a mild preference to have these capitalized and camel-cased like `RollDie` and `Random`.

src/core/randomeval.js
```diff
@@ -0,0 +1,62 @@
...
+
+export function evaluaterandomops(G, ctx) {
+  let randomresults = {};
+  let ctx2 = ctx;
+
+  // some flow tests run without a defined G
+  if (G && G._randomOps !== undefined) {
+    G._randomOps.forEach(r => {
+      switch (r.op) {
+        case 'D6': {
```
nicolodavis: Use a constant named `RANDOM_D6` here?

src/core/randomeval.js
```diff
@@ -0,0 +1,62 @@
...
+          const { ctx: ctx3, randomnumber } = randomctx(ctx2);
+          ctx2 = ctx3;
+          const d6 = Math.floor(randomnumber * 6) + 1;
+          randomresults[r.fieldname] = d6;
+          break;
+        }
+        case 'R': {
```
nicolodavis: Use a constant named `RANDOM_R` here?

src/core/reducer.js
```diff
@@ -44,6 +44,12 @@ export function createGameReducer({ game, numPlayers, multiplayer }) {
   };
 
   const state = game.flow.init({ G: initial.G, ctx: initial.ctx });
+
+  // either take the provided seed, or use seedrandom to create one.
+  // Math.seedrandom returns an autoseed with local entropy only.
+  const seed = game.seed === undefined ? Math.seedrandom() : game.seed;
```
nicolodavis: I personally think it's better to pass `seed` down to `FlowWithPhases` in `game.js` and have it initialize `ctx` properly.

src/core/random.js
nicolodavis: Also, can you make this accept a second parameter for the die dimension? We can use a D6 if nothing is passed, but it would be nice to allow the user to override it.

src/core/randomeval.js
Stefan-Hanke: I've added a `SpotValue` object containing all common dice values from D4 to D20.

src/core/randomeval.js
Stefan-Hanke: see above.

PR #104, opened 2018-02-09
fix formatting of changelog

sladwig:
the changelog is not correctly formated for the new versions, also found a small typo (should be 0.17.1) and made all the commits clickable...


sladwig:
I signed it!

nicolodavis:
Thanks! :)

PR #105, opened 2018-02-10
Dice: Adding Dice to UI Package

weatherfordmat:
As part of the UI Toolkit project, I am adding dice as a basic component.


src/ui/dice.test.js
```diff
@@ -0,0 +1,65 @@
...
+    expect(dice.html()).toContain('custom');
+  }
+});
+
+test('state change', () => {
+  {
+    const dice = Enzyme.mount(<Dice dotStyle={false} faceValue={4} />);
+    expect(dice.state().faceValue).toBe(4);
+    dice.setProps({ faceValue: 3 });
+    expect(dice.state().faceValue).toBe(3);
```
akaguny: it's tests for tests?) I think that only one exept is allowed in this test

src/ui/dice.test.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+  {
+    const dice = Enzyme.shallow(<Dice className="custom" />);
+    expect(dice.html()).toContain('custom');
+  }
+});
+
+test('state change', () => {
+  {
+    const dice = Enzyme.mount(<Dice dotStyle={false} faceValue={4} />);
```
akaguny: you send `dotStyle={false}` how you expect effect?

src/ui/dice.test.js
```diff
@@ -22,14 +22,15 @@ test('basic', () => {
     const dice = Enzyme.shallow(<Dice className="custom" />);
     expect(dice.html()).toContain('custom');
   }
+  {
+    const dice = Enzyme.shallow(<Dice dotStyle={false} faceValue={1} />); // renders plain text instead of styled dots
+    expect(dice.html()).toContain(1);
```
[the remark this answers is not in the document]
weatherfordmat: Good catch.

src/ui/dice.test.js
```diff
@@ -0,0 +1,66 @@
...
+    expect(dice.state().faceValue).toBe(1);
+    dice.setProps({ faceValue: 1 });
+    expect(dice.state().faceValue).toBe(1);
+  }
+});
+
+test('onClick', () => {
+  {
+    const dice = Enzyme.mount(<Dice faceValue={4} />);
+    expect(dice.state().faceValue).toBe(4);
```
akaguny: it's not onclick test

src/ui/dice.test.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+  {
+    const onClick = jest.fn();
+    const dice = Enzyme.mount(<Dice onClick={onClick} faceValue={4} />);
+    dice.simulate('click');
+    expect(onClick).toHaveBeenCalled();
+    expect(dice.state().faceValue).not.toBe(0);
+  }
+  {
+    const onClick = jest.fn();
```
akaguny: you may create 
```js
describe('onClick, ...
let onClick;
beforeEach(()=>{
 onClick = jest.fn()
}
test('',...

```

src/ui/dice.test.js
```diff
@@ -0,0 +1,66 @@
...
+    const dice = Enzyme.mount(<Dice onClick={null} faceValue={4} />);
+    dice.simulate('click');
+    expect(onClick).not.toHaveBeenCalled();
+    expect(dice.state().faceValue).not.toBe(0);
+  }
+  {
+    const onClick = jest.fn();
+    const dice = Enzyme.mount(<Dice onClick={onClick} faceValue={4} />);
+    dice.simulate('click');
+    expect(dice.state().faceValue).not.toBe(0);
```
akaguny: maybe add expect `expect(onClick).toHaveBeenCalled();`

src/ui/dice.test.js
akaguny: when you use tree hierarchy of tests you can describe what cases you testing and it the good way for cut duplication of code

src/ui/dice.test.js
```diff
@@ -0,0 +1,66 @@
...
+    expect(dice.text()).toContain(1);
+  }
+});
+
+test('state change', () => {
+  {
+    const dice = Enzyme.mount(<Dice faceValue={4} />);
+    dice.setProps({ faceValue: 1 });
+    expect(dice.state().faceValue).toBe(1);
+    dice.setProps({ faceValue: 1 });
```
akaguny: it's duplicate?

src/ui/dice.test.js
```diff
@@ -0,0 +1,66 @@
...
+Enzyme.configure({ adapter: new Adapter() });
+
+test('basic', () => {
+  {
+    const dice = Enzyme.shallow(<Dice />);
+    expect(dice.html()).toContain('div');
+  }
+  {
+    const dice = Enzyme.shallow(<Dice className="custom" />);
+    expect(dice.html()).toContain('custom');
```
akaguny: `div class="custom"`
you should catch any code mutation in tests for safe refactoring of tested code

src/ui/dice.test.js
weatherfordmat: It seems so, but that tests the if branch in the componentWillReceiveProps. Otherwise, line #36 in dice.js would remain untested since we couldn't tell if it would still change the state.

src/ui/dice.test.js
weatherfordmat: Not sure I understand.

src/ui/dice.test.js
weatherfordmat: I can add that, but not sure it's necessary.

src/ui/dice.test.js
weatherfordmat: Okay. Moved to "State" tests.

src/ui/dice.test.js
weatherfordmat: I'm not sure in this case it helps too much. Additionally, it follows the pattern in the rest of the testing files so that would be part of a larger refactoring.

src/ui/dice.test.js
akaguny: expect(dice.html()).toContain('custom');
<custom class="div"> will be passed

weatherfordmat: Correct, that's meant to pass? I'm missing something

src/ui/dice.test.js
akaguny: ok

src/ui/dice.test.js
akaguny: yes it's duplicate code, sorry.
what diff between your tests at L46 and L60?

weatherfordmat: Okay so, let's cut out L.60+ to combine the mount functions and have both expect statements.

src/ui/dice.test.js
weatherfordmat: Okay. I see. So we should use something like hasClass?

akaguny: yes)

nicolodavis:
This is great, but will need to use the API being built in https://github.com/google/boardgame.io/issues/68, so maybe you and Stefan can coordinate efforts on this?

weatherfordmat:
I'm a less familiar with the inner mechanics of the core package so I may leave this to Stefan and follow the discussion to see when the PR is merged.

Stefan-Hanke:
This code currently does not use the library at all - it is just the UI side of rolling a die. It is decoupled by introducing an `onClick` prop.

It is not the dice to generate the random value - drop that code. The random value is going to be _injected_ into the die (presumably via a prop), since in a multiplayer game it will be generated on the server and passed down to the client.

If we want to use this as an _example_ of how to use randomness and dice together, I'd suggest you create a real one below a _examples/components/randomdice/_ directory. I'd use an exiting component to get started.

weatherfordmat:
@Stefan-Hanke, yes, we are waiting for the randomness discussion to close before continuing with this.

The second point, I am a little less clear on. This is not meant to be primarily an example, but a basic UI component to use along with Cards, Deck, etc. It will be added to the storyboard as an example as well, but I think I'm missing what you mean by using an existing component.

Stefan-Hanke:
Either I don't understand the goal of this PR, or we are talking at cross purposes.

Why exactly are you waiting for randomness to be finished? Where do you see the connection between both? In the end, it will be the game programmer to connect the random value provided by the server to the corresponding UI.

Alright, _component_ was the wrong term, the code calls it _module_. See e.g. the [tic-tac-toe](https://github.com/google/boardgame.io/tree/master/examples/modules/tic-tac-toe) module. I was talking about another one of those that showcases randomness together with the dice UI implementation here. If you wanted to implement say a game using dice, or just a tech demo, then you'd need to wait for the other PR to be finished.

weatherfordmat:
Okay. my original thinking was more in line with what you are saying, but I think I took @nicolodavis to mean that the method would be called inside the dice class. 

This means that possibly that the dice UI should be a stateless functional component instead. We could merge the UI part and wait for the other PR to be finished before providing an example.

nicolodavis:
I've changed the component to be functional (sorry, I managed to confuse myself earlier when I said it should use the Random API).

Any change you can make the D6 values resemble their real life counterparts? For example, a 5 usually contains a dot at the very center (your component renders a row of three dots followed by a row of two).

nicolodavis:
You can test the component visually by running: `npm run storybook`

nicolodavis:
Closing this for lack of activity. Feel free to re-open if you pick it up again.

Issue #99, opened 2018-02-06
OnTurnBegin


Tacohej:
Hi!

Is there a reason for onPhaseBegin, onPhaseEnd, onTurnEnd being implemented but not onTurnBegin? Otherwise,  I would gladly give it a go.

nicolodavis:
Go for it :) I didn't think of `onTurnBegin`, but I think it would be good for uniformity.

PR #109, opened 2018-02-12
On turn begin

Tacohej:
This is my first pull request to this repo based on Issue #99 (by me). Hopefully I got everything right.

I made the decision that the first `onPhaseBegin `will always trigger before the first `onTurnBegin`. To trigger the `onTurnBegin` first would I guess be easy using a "dummy phase.


Tacohej:
CLA signed!

PR #110, opened 2018-02-16
Creating AI section of Game object

Felizardo:
Following what was discussed in #7 and [here](https://docs.google.com/document/d/1RXr12MMOiB6GDw09_h4pfcItZ8t7ocRPxQ0_-etQIY8/edit).

Could you please create a new branch for the AI section? We still need to implement:
- Client (and server) API for adding bots and executing them.
- AI algorithms (bots).
- Documentation for everything
- Examples

But because this is getting large, I would like to validate each part with distinct PRs... So it would be good to have an AI branch that I can send those PRs to.


Felizardo:
I was just re-reading the code I submitted, and I forgot the comments of ai.js... I will do it tomorrow or sunday.

nicolodavis:
Added an `ai` branch.

Felizardo:
The comments are done :)

src/core/ai.js
```diff
@@ -0,0 +1,156 @@
...
+ * Sets up the necessary information for running bots in a game.
+ *
+ * There are two parts that need to be described:
+ * - What are the next possible moves given a particular game state.
+ * - What is the evaluation function (score) for a particular state.
+ *
+ * A bot might use this information to maximize its score and minimize
+ * other player's scores (i.e. MiniMax and similar algorithms).
+ *
+ * @param {...object} possibleMoves - Function with signature (G, ctx) => moves
```
nicolodavis: We may have different ways of wanting to expand `possibleMoves`. This function provides one such way.

I think this interface will be cleaner and more extensible if we ask the user to provide:

```
ai: {
   possibleMoves: () => [...]
   score: () => {}
}
```

in raw form (the return value of this function, basically), and then we provide different ways to generate `possibleMoves` and `score` independently.

```
ai: {
   possibleMoves: MoveRange(...),
   score: TaskPriority(...),
}
```

Each of these generator functions can accept whatever it wants as long as it generates the raw form correctly.

src/core/ai.js
```diff
@@ -0,0 +1,156 @@
...
+ *                                    OR
+ *
+ *                                    Object with two keys:
+ *                                    {
+ *                                      ranges:
+ *                                        Object with key as move names
+ *                                        and values with arguments ranges. Each
+ *                                        range can be an object with {min, max, step}
+ *                                        or a list of possible values. Ex:
+ *                                        { clickCell: [{min: 0, max: 8}] }
```
nicolodavis: What about moves with multiple arguments?

src/core/ai.js
Felizardo: Ok! That was actually my first idea but I was having a hard time to find a place to add this "MoveRange" or "TaskPriority" function. But I think they can sit on ai.js and be part of the exports = {}. 
I will change this later today (or maybe tomorrow, I am kind of sick)

nicolodavis: Get better soon!

src/core/ai.js
Felizardo: This works with moves with multiple arguments. Each argument can have its own range object or array of possible values. See for instance this test with two arguments:
https://github.com/Felizardo/boardgame.io/blob/master/src/core/ai.test.js#L99

Felizardo: I think the comment is missing the point here, it should say that it is an array of objects. I will fix that.

src/core/ai.js
Felizardo: Ok this is done now :)

nicolodavis:
Merged!

Felizardo:
Cool! I will start working on the interface to bots and their implementation now :D

PR #111, opened 2018-02-17
SVG token templates

Stefan-Hanke:
This PR adds SVG models and a React adapter called `SvgComponent`. An SVG model is a group that renders into a specific shape and must not be modified. The `SvgComponent` adds a mean to translate it in x and y directions; it does not define the `<svg>` top level element.

Since I don't know whether this moves into the right direction, I've only added a handful components. By decoupling the SVG models from react, these should be reusable across UI frameworks.

One does not need to touch the story book for new SVG models, it suffices to add a new model to _svg.js_ and export it.

The SVG models themselves are written to render themselves uniformly into the rectangle (0,0) and (1,1). The intention is to make it predictable when applying transformations or other operations.


nicolodavis:
Can you merge this with `Token`? Something like:

```
<Token template="meeple"/>
```

The current plan is to have `Token` render as a:
- generic square inside `Grid`
- generic hexagon inside `HexGrid`
- a circle (?) elsewhere

We can then provide a prop like `template` to customize this. `Token` already handles interactions with the grids (and drag-n-drop eventually), so I'd prefer to keep everything in this one class.

Stefan-Hanke:
I'd try to make Token independent of where it is used. I would drop `_inHexGrid and `_inGrid` and use only the new `template` prop. The hex grid and normal grid could just inject the corresponding props. Ok?

nicolodavis:
You mean use `template` to indicate whether it is a hexagon or square when used in a grid? As long as we can preserve the existing behavior of the token taking on the shape of the grid it is in automatically, I'm ok retiring props that are superfluous.

Stefan-Hanke:
I've pushed the calculation of `center`, `width`, and `height` to `HexGrid` and pass it down. Maybe it needs a bit more cleanup. I hope this did not get too far into lands where we don't want to be.

There is a React warning about invalid children of `HexGrid` in the second tab on the `HexGrid` story. I don't know how to fix that. I've tried to use object spread but it looks like that operator is not supported there? Looks strange.

src/ui/svgcomponent.js
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Use of this source code is governed by a MIT-style
+ * license that can be found in the LICENSE file or at
+ * https://opensource.org/licenses/MIT.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export class SvgComponent extends React.Component {
```
nicolodavis: I like that everything in `svg.js` can be easily reused across UI frameworks. How about we auto-generate `SvgComponent` wrappers for them here so that users don't have to be aware of this class (it just remains as an internal implementation adapter for React users).

src/ui/token.js
```diff
@@ -51,13 +49,15 @@ class Token extends React.Component {
     x: PropTypes.number,
     y: PropTypes.number,
     z: PropTypes.number,
+    template: PropTypes.func,
+    center: PropTypes.object,
```
nicolodavis: I don't think we want users to manipulate `center` directly. Let's rename it to `_center` and make it an internal prop.

src/ui/token.js
```diff
@@ -158,30 +158,20 @@ class Token extends React.Component {
...
-          style={this.props.style}
-          onClick={this.props.onClick}
-        >
-          {this.props.children}
-        </Square>
+    if (this.props.template) {
+      return React.createElement(
+        this.props.template,
+        {
+          x: this.state.x,
```
nicolodavis: Now that we're calculating `center`, we no longer need to pass `x, y and z` to the template class, correct?

Stefan-Hanke:
I need to take a step back and rethink this. For me, `center` is something that could be hidden inside the respective grid itself...

src/ui/svgcomponent.js
Stefan-Hanke: Can you elaborate on this? Where do you think the wrapper component be created? For reuse of the models, I suggest we extract the models themselves and use an UI-agnostic representation (an array somewhere).

src/ui/token.js
Stefan-Hanke: The code now passes a function to recalculate graphical properties that are dependent on x, y, z, and the cellSize. Otherwise, the current animation logic does not work since it sets the coordinates on Token, and thus Token must be able to recalculate e.g. the center.

src/ui/token.js
Stefan-Hanke: No - this breaks the animation code. The `onClick` handler passes the cube coordinates up the hierarchy, and in the case of the `Runner` (inside storybook) then sets the state, which then gets interpolated inside Token.

Stefan-Hanke:
I don't know - my gut feeling tells me to abandon that branch, and just reuse parts of it for a new branch. I don't know whether it was the best decision to try to handle all graphics stuff uniformly - I think it would have been wiser to leave that `insideHex` and `insideGrid` stuff in there for the time being.

What do you think? You're ok with that?

nicolodavis:
OK, regardless of which field goes where, let's look at the higher-level picture:

### Token
- The only non-container class that I'm hoping we will need.
- Takes `template` to determine what shape it takes.
- Takes `x, y, z` to determine grid position (when used inside a grid).
- Can accept other props (say `cx, cy` for the co-ordinates of the center in SVG space) to determine positioning and dimensions when used outside a grid.

### HexGrid / Grid
- Translate `x, y, z` to `cx, cy`.
- Inject a default `template` if one is not provided (`Hex` in the case of a `HexGrid`).

I think we can arrange it such that `Token` has no knowledge of it being inside a particular grid (the grid can just do the translation from grid co-ordinates to absolute co-ordinates).

Is that what you had in mind as well?

src/ui/svgcomponent.js
nicolodavis: If I'm reading this correctly, it sounds like the user needs to actually be aware of this class in order to use the SVG models. Is this the intended usage?

```
const wheat = <SvgComponent component={Wheat}/>;
const token = <Token template={wheat}/>;
```

If so, I'm suggesting that we could just auto-generate wrappers like:

```
const WheatSVG = (props) => <SvgComponent component={Wheat} ...props/>;
const CornSVG = (props) => <SvgComponent component={Corn} ...props/>;
...
```

and use them like:

```
const token = <Token template={WheatSVG}/>
```

(swapping the names of `Wheat` and `WheatSVG` perhaps in the actual implementation).

Stefan-Hanke: OK, this makes sense.

Stefan-Hanke:
Yes.

Basically, I did not know about how the animation actually worked, it broke, and I got confused. I have worked on the branch in the last days but did not push those changes. I'll check whether the code matches our expectations, change it, and push it.

For the animation to work, the template itself will need to know where it is located - the grid backgrounds do not use Token, but rather `Hex` and `Square` directly. We could use Token here as well, that way the templates could stay clean of all coordinates.

Stefan-Hanke:
Argh, I'm very unconvinced about the changes made here.

I'm too unexperienced with React to use it effectively. Take e.g. the MeepleSVG component which I wanted to parameterize with the color. It looks like a mess :(

`Token` takes x,y,z coordinates even within the square grid. I'd like to just pass "coordinates" and hide the actual structure.

nicolodavis:
I'll take a look later today and add some commits to help you with this.

src/ui/svgcomponent.js
```diff
@@ -0,0 +1,58 @@
...
+    let tx = 0,
+      ty = 0;
+    if (this.props._center !== undefined) {
+      tx = this.props._center.x;
+      ty = this.props._center.y;
+    }
+
+    return (
+      <g onClick={this.props.onClick} transform={`translate(${tx}, ${ty})`}>
+        {this.props.component}
```
jstacoder: i think here you want to actually render the passed component? you could set it to a variable above the return statement ( and just make sure its uppercase like a react component should be) like `const SvgComponent = this.props.component` then on this line render it `<SvgComponent />`

nicolodavis:
I'm going to break this up into a few changes to make it more manageable.

src/ui/svgcomponent.js
nicolodavis: Shouldn't it render the component even as the code stands now?

jstacoder: The way it stands the component is just being referenced here, an array of components can be rendered like that ie: `{this.props.children}` but I’m pretty sure that a single component needs to be rendered like `<Component />` while currently it’s just a reference `Component`, I only mentioned this because I thought it wasn’t working yet? If it is working currently then then I guess i am probably Mistaken.

nicolodavis:
Going to close this now that we've implemented some of the ideas from this PR.

PR #112, opened 2018-02-18
Hover debug pane on top of game

philihp:
* Changes the styling of the debug pane so it doesn't reposition elements on the screen, and instead floats on top of them.

* Adds `jest-cli`, which was necessary to commit and run tests -- but i'm guessing everyone probably globally installed it, so it was never an issue.

## Before 
before mp4

## After
after mp4


Stefan-Hanke:
How did you run the tests? Using `npm run test`, it should just work, since `jest` is available as command inside `node_modules/.bin`.

Felizardo:
That looks good :)

nicolodavis:
Looks great! Tests should run fine without `jest-cli` (if you use `npm test`), so can you check if that works and just include the CSS change in this PR?

philihp:
Hmmm, jest-cli thing won't happen again after cleaning node_modules. Removing...

PR #113, opened 2018-02-18
drop seedrandom by inlining the alea PRNG

Stefan-Hanke:
This PR drops seedrandom that was introduced during implementation of the random API.
The alea code currently is inlined in _random.js_.
If preferred we could just as well have a _random.alea.js_ to contain that code.

Please note:
* Code coverage dropped since an if path inside alea is not taken. 
* I've commented out function definitions that were not used.


src/core/random.js
```diff
@@ -6,7 +6,87 @@
...
+  return mash;
+}
+
+// this is the function that uses Alea to create a variety of methods.
+// the commented-out code was not used and lowered code coverage.
+function impl(seed, opts) {
+  var xg = new Alea(seed),
+    state = opts && opts.state,
+    prng = xg.next;
+  // prng.int32 = function() { return (xg.next() * 0x100000000) | 0; }
```
nicolodavis: If we don't envision ourselves using these in the future, let's just remove them rather than leaving them commented out. Alternatively, you could use `prng.int32` in the die roll code below with the modulus operator.

Stefan-Hanke: I've dropped them - I think one primitive to generate random numbers inside the library is sufficient.

nicolodavis:
I split it into a separate file and excluded it from test coverage (we're not testing Alea here, so it doesn't make sense to include it in the test coverage score).

PR #114, opened 2018-02-18
Add random to Liar's Dice example

philihp:
I started to add the rolling of random dice that @Stefan-Hanke just added to the Liar's Dice example. Two thoughts:

* First, it seems random dice don't currently play well with secret info in a multiplayer context. I had to disable the multiplayer view's STRIP_SECRETS for this.

* Second, is it intentional that we can't use Random during game state setup? It seems as though either we should have that, or that every game should include a "setup" phase, where card decks are shuffled, etc.

### DIFF
--- a/examples/modules/liars-dice/game.js
+++ b/examples/modules/liars-dice/game.js
@@ -6,8 +6,7 @@
  * https://opensource.org/licenses/MIT.
  */
 
-import { Game, Random } from 'boardgame.io/core';
-// import { PlayerView } from 'boardgame.io/core';
+import { Game, Random, PlayerView } from 'boardgame.io/core';
 
 const LiarsDice = Game({
   name: 'liars-dice',
@@ -72,7 +71,7 @@ const LiarsDice = Game({
     ],
   },
 
-  // playerView: PlayerView.STRIP_SECRETS,
+  playerView: PlayerView.STRIP_SECRETS,
 });
 
 export default LiarsDice;
--- a/examples/modules/liars-dice/game.js
+++ b/examples/modules/liars-dice/game.js
@@ -6,7 +6,7 @@
  * https://opensource.org/licenses/MIT.
  */
 
-import { Game, Random, PlayerView } from 'boardgame.io/core';
+import { Game, Random } from 'boardgame.io/core';
 
 const LiarsDice = Game({
   name: 'liars-dice',
@@ -22,27 +22,25 @@ const LiarsDice = Game({
   }),
 
   moves: {
-    bid: (G, ctx, id) => {
-      const cells = [...G.cells];
-      if (cells[id] === null) {
-        cells[id] = ctx.currentPlayer;
-      }
-      return { ...G, cells };
-    },
+    bid: G => G,
     challenge: G => G, // TODO
-    roll: (G, ctx) => ({
-      ...G,
-      players: {
-        ...G.players,
-        [ctx.currentPlayer]: [
-          Random.D6(),
-          Random.D6(),
-          Random.D6(),
-          Random.D6(),
-          Random.D6(),
-        ],
-      },
-    }),
+    roll: (G, ctx) => {
+      const newG = {
+        ...G,
+        players: {
+          ...G.players,
+          [ctx.currentPlayer]: [
+            Random.D6(),
+            Random.D6(),
+            Random.D6(),
+            Random.D6(),
+            Random.D6(),
+          ],
+        },
+      };
+      console.log(newG);
+      return newG;
+    },
   },
 
   flow: {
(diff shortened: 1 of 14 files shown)


Stefan-Hanke:
Looks like the player view currently is called with `pid===null`. This leads to all players being stripped. Try using:

```js
  playerView: (g, ctx, pid) => {
    if(pid === null) {
      return g;
    }
    return PlayerView.STRIP_SECRETS(g, ctx, pid);
  },
```

nicolodavis:
About calling `playerView` when `multiplayer == false`, I think we added that so a developer can see the effect when prototyping locally. I think we should call it with `currentPlayer` if `playerID` is unavailable.

About calling `Random` during setup, you can already create a dummy initial phase that accomplishes exactly this. However, that's a bit clunky and we should probably have better support for this during setup. Perhaps we can modify `Random.Die` and `Random.Number` to:
- add a random op request when called inside `moves`, and
- directly return a random result when called inside `flow` or `setup`.

Thoughts?

nicolodavis:
@Stefan-Hanke About the naming of `multiplayer`, perhaps we should change it to `online` or `local` (where `multiplayer` would be equivalent to `!local`).

nicolodavis:
Also, I noticed that we don't strip `prngstate` (only `seed`). I'm going to move both these keys inside a single one called `random` that gets stripped.

nicolodavis:
Another thing we probably should do is enable allowing the user to click on different player IDs in the Debug pane and be able to see those respective respective views (as simple as setting playerID in the client component).

Stefan-Hanke:
@nicolodavis On the naming issue: Both `local` and `online` do decouple from `numPlayers`. I'd personally would go with something like `usingServerAt` to be as expressive as possible, but I'm ok with the former two also.

On the setup issue: Well, I don't like APIs that behave differently depending on where calls come from. This is just one more thing to keep in mind when doing game development. In effect, it would encode the knowledge that `move` and `setup` is always called on the server (given that a server is used), and `flow` is not.

We could e.g. run random ops directly after calling setup inside [reducer.js, l.45](https://github.com/google/boardgame.io/blob/master/src/core/reducer.js#L45)? I'm unsure.

nicolodavis:
Running the random ops right after setup sounds good!

Separately, I think we still have an annoying quirk with the API. What happens if someone calls it inside `onPhaseBegin` or some other `flow` code? Currently, it won't be processed until the next move is made, which is not desirable. I think we'll need to call `RunRandom` in a few more places in `flow.js`.

Stefan-Hanke:
Yes, it all looks and feels like it's not the final state. I don't think it's a good idea to spread those calls around though. I'd like to have it at 2-3 places most, and also support custom flows that a game developer might use.

nicolodavis:
https://github.com/google/boardgame.io/pull/119 should take care of most of the remaining concerns we have with the Random API.

examples/modules/liars-dice/game.js
```diff
@@ -6,32 +6,65 @@
...
-      return { ...G, cells };
+    bid: G => G,
+    challenge: G => G, // TODO
+    roll: (G, ctx) => {
+      const newG = {
+        ...G,
+        players: {
+          ...G.players,
+          [ctx.currentPlayer]: [
+            Random.D6(),
```
nicolodavis: You should use the new ability to roll multiple dice at the same time.

examples/modules/liars-dice/game.js
```diff
@@ -6,32 +6,65 @@
...
+    ],
+  },
+
+  playerView: (G, ctx, pid) => {
+    console.log(G, ctx, pid);
+    // if (pid === null) {
+    //   return G;
+    // }
+    return G;
+    // return PlayerView.STRIP_SECRETS(G, ctx, pid);
```
nicolodavis: Is `STRIP_SECRETS` not behaving as you would expect?

philihp: @nicolodavis not locally, when viewing through the multiview :\ running this branch's examples shows this.

nicolodavis:
A bunch of stuff has changed that may have likely also fixed your problem. Can you sync and let me know if this works?

philihp:
I noticed :) I'll sync this up and see if it's resolved.

philihp:
^feel free to close this -- i have a lot of momentum in my own game for the moment and i don't wanna context switch out :) or leave it open, i have every intention of getting to this, but i can just make another PR when the time comes.

nicolodavis:
Closing this for now. I think I want to strip down the Liar's Dice example into just the constituent API's actually (how to roll dice, how to display secret state etc. with some toy examples).

philihp:
SGTM

Issue #68, opened 2018-01-07
Random API


nicolodavis:
Dice, shuffling etc. all require randomness, but having impure reducers (using `Math.random()`, for example) breaks the ability to view earlier game states (and is an anti-pattern in Redux). We should use a seed in `ctx` (that's stripped away before sending it to the client) that is used to predictably generate random numbers.

Plan:
- [x] API to generate random numbers.
- [x] API for various die rolls.
- [x] API to shuffle an array.
- [x] Allow calling the Random API in `flow` code (only really supported in `moves` at the moment).

scheijan:
I've used https://github.com/davidbau/seedrandom before to solve a similar problem. If you think that could be a solution here as well, I could take a stab...

nicolodavis:
Using seedrandom sounds good!

Felizardo:
I just want to note that this will also be a problem for when developing a generic AI, as the algorithm will need to know what is the probability of each state happening for maximizing the expected value of each action.
 
If we just use Math.random() or any similar approach, it can lead to almost infinite amount of distinct states. Maybe we need to develop an API that will allow us to tell how many possible states it can happen from there -- ie throw a 6-sided dice API. 
But to be completely honest I still don't have a good solution for this. I want to focus on having a simple AI for deterministic games first, and then maybe we can revisit this.

philihp:
I think keeping a seed in the ctx is the best way to make sure the rescuers stay deterministic.

If a bot/AI were to search the state tree with random elements would probably want an MCTS, since something like a draw from a 6-deck shoe in blackjack can have 300+ branches per level of depth.

There will need to be a way to tell Game to use a new seed, though, because the AI will want to sample what “could” happen, but if it knows what “will” happen with the actual seed, that wouldn’t be fair to the humans.

Felizardo:
Yes, again, I think keeping the seed is a good solution, I just wanted to start thinking about how to solve this for AI as it is related. We are going to need to have more information for sure, after all we are going to need to generate a graph to run Markov decision process algorithms...

scheijan:
@nicolodavis where would like to you see functions like `shuffle()` or `random()`? Directly on `ctx`? As a method of the GameBoard? A new module?

nicolodavis:
`ctx` should be a plain object, so let's put them in a separate module?

nicolodavis:
@scheijan Haven't heard from you in a while. Are you still working on this?

scheijan:
[...]

PR #116, opened 2018-02-19
Server-side array shuffling.

philihp:
Adds a method for performing a deterministic random shuffle for the 3rd point in #68. The game logic requests that one of the attributes in G be shuffled. The server then assumes it's an array and does a shallow shuffle of it.

To keep the interface simple, only arrays at the root of G are able to be shuffled. If the user wants to shuffle a nested array, they'll have to do it themselves.

This should be rebased on top of https://github.com/google/boardgame.io/pull/113 prior to merging.

@nicolodavis @Stefan-Hanke


nicolodavis:
Why not allow nested fields as well? You can just split on `.` and use the index operator to retrieve deeper fields without changing the interface.

nicolodavis:
Philihp, can you merge `master` in? There are some conflicting changes.

philihp:
@nicolodavis I started to do nested deep fields but it got tricky with having to clone every level of the object to preserve immutability of the source `G` for Redux, and the added complexity for little gain made me think I was overengineering it. Here's the test I was trying to satisfy:

```
test('Random.Shuffle works on a nested attribute', () => {
  let ctx = { random: { seed: 'some_predetermined_seed' } };
  const tiles = ['A', 'B', 'C', 'D', 'E']
  let G = {
    players: {
      0: tiles,
      1: tiles
    }
  }

  let G2 = Random.Shuffle(G, 'players.1')
  let { G: G3, ctx: ctx2 } = RunRandom(G2, ctx)
  expect(G.players['0']).toMatchObject(tiles)
  expect(G.players['1']).toMatchObject(tiles) // this was the tricky one :(
  expect(G3.players['0']).toMatchObject(tiles)
  expect(G3.players['1']).not.toMatchObject(tiles)
})
```

Stefan-Hanke:
Ah, there's nothing like a small little recursion problem. :)
Warning: I've omitted error handling.

```js
function fisheryates_shuffle(array) {
  var currentIndex = array.length, temporaryValue, randomIndex;
  while (0 !== currentIndex) {
    randomIndex = Math.floor(Math.random() * currentIndex);
    currentIndex -= 1;
    temporaryValue = array[currentIndex];
    array[currentIndex] = array[randomIndex];
    array[randomIndex] = temporaryValue;
  }
  return array;
}

const tiles = ['A', 'B', 'C', 'D', 'E'];
let G = {
  players: {
    0: tiles,
    1: tiles
  }
}

function shuffle(datastructure, structuredname) {
    // recursion end-case
    if(structuredname.indexOf(".") < 0) {
        let data = fisheryates_shuffle([...datastructure[structuredname]]);
        return { ...datastructure, [structuredname]: data };
    }

    // strip one level, recurse
    const [car, ...cdr] = structuredname.split(".")
    let data = datastructure[car];
    let shuffleddata = shuffle(data, cdr.join("."));

    // and merge
    return { ...datastructure, [car]: shuffleddata};
}

let G2 = shuffle(G, "players.1");

console.log(JSON.stringify(G, undefined, 3));
console.log(JSON.stringify(G2, undefined, 3));
```

philihp:
Yeah, doing that only works when there's only one Random op. In RunRandom, [this line](https://github.com/google/boardgame.io/blob/master/src/core/random.js#L80) also doesn't work with deep attributes... I think it's a lot of complexity we can live without for the time being.

nicolodavis:
I think it's ok if we leave this PR to just handle top level field names.

We'll eventually need to support deeper fields before we can properly release the Random API, though (I think it's too restrictive to game designers to have all dice and card decks at the top level). For example, some game designers will prefer the following model for a card game (I personally think this is quite natural too):

```
Game({
  setup: numPlayers => ({
    players: {
      '0': { hand: [...], discard: [...] },
      '1': { hand: [...], discard: [...] },
       ...
      '<numPlayer-1>': { hand: [...], discard: [...] },
    }
  })
})
```

If we don't support nested fields, everyone will have to flatten their states:

```
Game({
  setup: numPlayers => ({
    '0-hand': [...],
    '1-hand': [...],
    ...
    
    '0-discard': [...],
    '1-discard: [...],
    ...
  })
})
```

So, maybe let's just get the shuffling logic in this PR and one of us can follow up with a broader change to support nested fields across the entire Random API later?

philihp:
SGTM. Good point, shuffling a discard is a common thing. For Liar's dice, it's already awkward to roll into secret player info.

Stefan-Hanke:
@philihp I agree. If you have suggestions for making the API more dev-friendly, feel free to propose. An extension of the _random language_ e.g. would be to allow terms like `5D6` which would reduce the boiler plate in Liar's Dice. Or even `4D6 2D12 1D20`.

If you mean by _awkward_ the nature of the API itself, that random values are not just returned, we'd need to change where game moves are executed - in case a server is present, it needs to be restricted to running only there.

nicolodavis:
Now that I think about it, this is actually possible!

Context: We want moves to execute on both client and server so that there is no lag (optimistic update). The server state is still authoritative and the client state is overwritten eventually (if it differs).

We could disable optimistic updates for random calls. For example, if G contains a random op, we could have the client just wait for the server update.

This way the random API can just return values directly. I think this will simplify things quite a bit, including our problem of not being able to call random code in the flow section at the moment.

philihp:
@Stefan-Hanke by awkward, I mean to your first point, if the user wants 5d6, they have to call Random.D6 5 times ([e.g.](https://github.com/google/boardgame.io/pull/114/files#diff-01733ebd7dc868f9954930479f02243fR35))... then they show up at the root of `G`, and have to be moved out during onMove ([e.g.](https://github.com/google/boardgame.io/pull/114/files#diff-01733ebd7dc868f9954930479f02243fR50))

PR #118, opened 2018-02-20
Random deep field

### LINKED ISSUE #117: Nested fields in the Random API
Just following up on some discussion at https://github.com/google/boardgame.io/pull/116.

Currently, the Random API only allows retrieving random results into fields at the top level of `G`. We should support deeper levels using a syntax like:

```
G = Random.Number(G, 'a.b.c');
G = Random.Shuffle(G, 'players.0.hand');
```

@Stefan-Hanke has some example code in https://github.com/google/boardgame.io/pull/116 that might serve as a good starting point.

Stefan-Hanke:
Fixes #117 

* Accessing an object is now done inside two functions `deep_set` and `deep_get`.
* I've added @philihp s test unmodified
* Random numbers and dice support setting deep fields also.

### DIFF
--- a/src/core/random.js
+++ b/src/core/random.js
@@ -70,7 +70,8 @@ export function evaluaterandomops(G, ctx) {
 
         case SHUFFLE: {
           const rng = alea(randomnumber);
-          G[r.fieldname] = shuffle(G[r.fieldname], rng);
+          const field = deep_get(G, r.fieldname);
+          randomresults[r.fieldname] = shuffle(field, rng);
           break;
         }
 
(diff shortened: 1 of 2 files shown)


src/core/random.js
```diff
@@ -83,10 +84,34 @@ export function evaluaterandomops(G, ctx) {
   return { randomresults, ctx: ctx2 };
 }
 
+function deep_set(G, structuredname, newvalue) {
```
akaguny: why no jsdoc?

src/core/random.test.js
```diff
@@ -8,6 +8,12 @@
 
 import { randomctx, RunRandom, addrandomop, DICE, Random } from './random';
 
+function checkrandom(value, min, max) {
```
akaguny: why not lower camelCase?

src/core/random.js
```diff
@@ -83,10 +84,34 @@ export function evaluaterandomops(G, ctx) {
...
+  });
+  return ctx;
+}
+
 export function RunRandom(G, ctx) {
   let { randomresults, ctx: ctx2 } = evaluaterandomops(G, ctx);
-  const G2 = { ...G, ...randomresults, _randomOps: undefined };
-  return { G: G2, ctx: ctx2 };
+
+  let G2 = G;
```
akaguny: what does it mean G2, G3, GN?

philihp:
Feel free to `git cherry-pick 64d7748f7726605e5a446f571018cc621be1f2e8` into your branch for some more tests https://github.com/google/boardgame.io/commit/64d7748f7726605e5a446f571018cc621be1f2e8

I looked into using `object-path-immutable` since it's supposed to be safe for use with Redux stores, but doing a Get from it felt strange (see [philihp:deep-set](https://github.com/google/boardgame.io/compare/master...philihp:deep-set?expand=1))

src/core/random.js
philihp: @akaguny they're like second and third generations of the G state.

G2 is the G with the random ops added
G3 is the G2 returned from the server

Stefan-Hanke:
@philihp nicolo proposed using  https://github.com/debitoor/dot-prop-immutable which IMO has a nice API. No strange `get` code there.

philihp:
that package looks great!

nicolodavis:
Given that `deep_get` and `deep_set` is only 10 lines long, if we think we won't need the additional features of `dot-prop-immutable`, let's keep things as is to avoid the extra dependency.

nicolodavis:
Closing this now that it is superseded by #119

PR #119, opened 2018-02-21
Simplify Random API

nicolodavis:
* Run random ops directly (instead of queuing them and executing them in two steps).
* Introduce a mechanism to disable optimistic updates on the client.
* Discard updates on the client that use randomness using this mechanism (just wait for the server instead).

This gets rid of some quirks with the Random API while still retaining the ability to run moves optimistically on the client.


nicolodavis:
@Stefan-Hanke @philihp Can you take a look?

I'll follow up with updating the documentation and bringing the test coverage back up to 100%.

Stefan-Hanke:
Overall: looks good to me. I'd wait with more docs until the stuff is merged and we have a proof-by-example that it works as we envisioned.

_Note: I am prefixing numbers to make referencing easier._

(1) What API would you prefer?
* `G = Random.D6(G, 'diceValue');`
* `const { G: G2, diceValue } = Random.D6(G);`

Before settling on the previous solution, I'd rather have this be a deliberate decision.
The first one was chosen because the random value was generated on the server and had to be transported back. The second one will make the library a bit smaller since we can just omit the necessity of deep state updates altogether.

I think a functional approach would try to hide the monadic `G`. However, I don't want any of this kind of abstraction forced onto a game developer.

I'd personally go with the second one, but I'm also ok with the first.

(2) The previous API was written having the seed on `ctx`. Now, both seed and `prngstate` are on `G`.  It feels like not _really_ belonging to `G`, but then, I'm ok with this.

(3) For me, the core part is the change in the reducer (and the other code to make it work). We currently need to do the state update client-side, since the attached middleware will send the change to the server. I quite like the name `optimisticUpdate`, since it does express the behaviour: Trying to update the state, but you used random => rollback.

(4) We still need to make the API a bit more expressive, however as done with the last PRs, let's do this separately.

(5) This makes #118 obsolete, however we could take `deep_get` and `deep_set` when we decide to use the first solution in (1). #114 would need to be updated, however I'd wait before the API is more expressive.

nicolodavis:
Regarding (1), I feel like the second approach puts more burden on the user to maintain the PRNG state correctly (i.e. they need to be consciously aware of it). That being said, incorrect code is possible in both approaches:

**Example 1**
```
const { diceValue } = Random.D6(G, 'diceValue');
return { ...G, field: derivative(diceValue) };
```

**Example 2**
```
const { diceValue } = Random.D6(G);
return { ...G, field: derivative(diceValue) };
```

I'm open to either approach. I think not having to worry about nested fields etc. would be a nice benefit of using the second approach.

nicolodavis:
Another approach (that purists will dislike) would be something like:

```
const dieValue = Random.D6(G);
return { ...G, dieValue };
```

The `Random` call modifies `G` and keeps the PRNG state updated. The only failure mode here would be if the user returns this `G` directly without creating a copy. However, given that returning a copy at
the end of a move is such a common idiom, I am not particularly worried about this.

The resulting code for the user is certainly much prettier. Compare:

```
    shuffleDeck(G) {
      const deck = Random.Shuffle(G, G.deck);
      return { ...G, deck };
    }
```

with

```
    shuffleDeck(G) {
      const { G: G2, result: deck } = Random.Shuffle(G, G.deck);
      return { ...G2, deck };
    }
```

What do you think?

Stefan-Hanke:
Yes, I think this is the way to go. For me, the pure approach using monads just doesn't cut it in languages where there is no explicit support, like `do` in Haskell. And all it does is hiding "something".

If we allow to destructively modify `G` we could inject the object directly into `Random`, and just have random calls like `const shuffledDeck = Random.Shuffle(G.deck)`. For that injection, the developer e.g. would have to create `Random` via a create function (`const Random = genrandom(G)`), or we pass `Random` via `ctx`?

When the PRNG state is hidden, we could eliminate the failure mode where that state is not correctly returned; the library could just take the one it passed in.

nicolodavis:
Updated the PR to support the following neat syntax:

```
moves: {
  rollDice: G => ({ ...G, dice: Random.D6() }),
}
```

The `Random` package initializes itself with data from `ctx` prior to each move, and updates `ctx` at the end of each move with the PRNG state.

philihp:
I like that this doesn't require an onMove handler, but it felt like that might have been a more appropriate place. This LGTM, I'd be down to merge and see how it works in the Liar's Dice demo.

nicolodavis:
Merged. Let me know once you've merged it with the Liar's Dice demo and I'll take a look.

Also, let's continue the discussion at https://github.com/google/boardgame.io/issues/68 about other features we may want to improve expressiveness of the API (or if the current feature set is good enough).

PR #123, opened 2018-02-24
Debug UI: fixes related to errors in arguments

Mrmaxmeier:
Hey,
the behaviour of the debug UI is pretty confusing.
Calling a move/event function with `[1, 2, 3]` results in a call with arguments `undefined, 2, undefined`.

This changes the arguments to all be evaluated in one call, thus `argText.split(',')` is no longer needed. As a result the test `3, unknown, 4` will error instead of evaluating to `3, undefined, 4`. I'd say this behaviour is sensible. I've changed the test accordingly.

This also implements basic error reporting.

### DIFF
--- a/src/client/debug/debug.css
+++ b/src/client/debug/debug.css
@@ -47,6 +47,11 @@
   font-weight: bold;
 }
 
+.debug-ui .move-error {
+  color: #a00;
+  font-weight: bold;
+}
+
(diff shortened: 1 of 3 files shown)


Mrmaxmeier:
@googlebot I signed it!

nicolodavis:
Looks great. Thanks for the fix!

PR #124, opened 2018-02-25
preset-env

philihp:
The tests on my machine have always failed when I tried to push
```
 server              |        0 |        0 |        0 |        0 |                |
  db.js              |        0 |        0 |        0 |        0 |... 26,36,44,45 |
  index.js           |        0 |        0 |        0 |        0 |... 130,132,133 |
```

With an error that was pointing at this
```
  */test('basic', async function () {var db
                  ^^^^^
```

On https://facebook.github.io/jest/docs/en/tutorial-async.html, it says to add the `env` preset to use async/await in tests. Adding it fixed this for me. It should be innocuous for others where this test already ran fine.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,6 @@
 {
   "presets": [
     ["env", { "modules": false }],
-    ["es2015", { "modules": false }],
     "react",
   ],
   "env": {
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,6 @@
 {
   "presets": [
+    ["env", { "modules": false }],
     ["es2015", { "modules": false }],
     "react",
   ],
--- a/package-lock.json
+++ b/package-lock.json
@@ -2408,65 +2408,27 @@
         "babel-plugin-transform-es2015-unicode-regex": "6.24.1",
         "babel-plugin-transform-exponentiation-operator": "6.24.1",
         "babel-plugin-transform-regenerator": "6.26.0",
-        "browserslist": "2.11.3",
+        "browserslist": "2.9.1",
         "invariant": "2.2.2",
         "semver": "5.4.1"
       },
       "dependencies": {
         "browserslist": {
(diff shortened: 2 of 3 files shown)


Stefan-Hanke:
I would be interested in knowing the root cause. `npm t` resulted in the mentioned error? 

`npm t` just runs `jest`. This in turn uses the configured babel stuff since `jest-babel` and `regenerator-runtime` modules are installed (see [here](https://facebook.github.io/jest/docs/en/getting-started.html)). This in turn uses the configuration found in `.babelrc`. The [es2015 preset](https://babeljs.io/docs/plugins/preset-es2015/) does not contain a transform for `async`.

Could you try adding the [async transform plugin](https://babeljs.io/docs/plugins/transform-async-to-generator/) to the test env?

Why does it work for others?

philihp:
@Stefan-Hanke it looks like the config of node that Husky 0.14 was using used a Node which didn't support it without this PR. The tests only failed during a push. Running a lone `node test` and `node test:coverage` worked fine.

I was also able to fix this by upgrading Husky to 0.15.

philihp:
Also, adding that plugin also did the trick.

https://github.com/philihp/boardgame.io/tree/transform-async-to-generator

Any of these three changes would be cool with me.

nicolodavis:
Changing to preset `env` is a good thing to do regardless of this problem, so I'm happy to get that in. You should also uninstall the `es2015-preset`.

nicolodavis:
Ugh, looks like preset `env` still tries to transform `async` even on Node versions that support it. Need to do a bit more research before migrating over to it. I reverted this and updated Husky instead.

https://github.com/google/boardgame.io/commit/a5fc52cb71483de94354ceff311c32727585ef35
https://github.com/google/boardgame.io/commit/8a79c745eb9a9d293d0aa9f0dec57edbecc9887d

PR #125, opened 2018-02-27
Fix a typo in example

SBeator:
The description in the example for player 2 should be 
```
<App playerID="2"/>
```
Instead of 
```
<App playerID="1"/>
```


nicolodavis:
Thanks!

PR #127, opened 2018-02-27
default reverse order

rzulian:
simple implementation of the reverse order that you can need in some phase of your game

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "babel-watch": "^2.0.7",
     "chess.js": "^0.10.2",
     "coveralls": "^3.0.0",
-    "cross-env": "^5.1.1",
+    "cross-env": "^5.1.3",
     "css-loader": "^0.28.7",
     "docsify-cli": "^4.1.12",
     "enzyme": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "babel-watch": "^2.0.7",
     "chess.js": "^0.10.2",
     "coveralls": "^3.0.0",
-    "cross-env": "^5.1.3",
+    "cross-env": "^5.1.1",
     "css-loader": "^0.28.7",
     "docsify-cli": "^4.1.12",
     "enzyme": "^3.0.0",
--- a/src/core/turn-order.js
+++ b/src/core/turn-order.js
@@ -55,13 +55,7 @@ export const TurnOrder = {
    */
   DEFAULT_REVERSE: {
     first: (G, ctx) => ctx.numPlayers -1 + '',
-    next: (G, ctx) => {
-      if ((ctx.currentPlayer == 0)) {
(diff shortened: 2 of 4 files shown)


rzulian:
I signed it!

src/core/turn-order.js
```diff
@@ -48,6 +48,22 @@ export const TurnOrder = {
...
   },
 
+    /**
+   * DEFAULT-REVERSE
+   *
+   * The default round-robin turn order starting from the last player.
+   */
+  DEFAULT_REVERSE: {
+    first: (G, ctx) => ctx.numPlayers -1 + '',
+    next: (G, ctx) => {
```
nicolodavis: This should also work:
(+ctx.currentPlayer + ctx.numPlayers - 1) % ctx.numPlayers + ''

nicolodavis:
Also, you should use the label 'ptal', not add it to your Pull Request description.

src/core/turn-order.js
rzulian: yes, this is cleaner

rzulian:
Sorry for the label. This is my first pull request ever!

nicolodavis:
Yeah, I realized that just now. Removing the recommendation to add labels from the guidelines.

Also, looks like you have a lint error: https://travis-ci.org/google/boardgame.io/builds/346812962?utm_source=github_status&utm_medium=notification

src/core/turn-order.js
```diff
@@ -48,6 +48,16 @@ export const TurnOrder = {
     next: (G, ctx) => (+ctx.currentPlayer + 1) % ctx.numPlayers + '',
   },
 
+    /**
+   * DEFAULT-REVERSE
+   *
+   * The default round-robin turn order starting from the last player.
+   */
+  DEFAULT_REVERSE: {
```
nicolodavis: Rename this to just `REVERSE`.

rzulian:
@nicolodavis I thinking to extend the TurnOrder to manage custom order, which is very common in boardgames. Problem is that current implementation is assuming player order starting from currentPlayer='0'.  What if I need to start a new phase or round base on the pass order of the previous phase? 
What do you think?

nicolodavis:
Yes, you can use `G` to store the starting player and use a custom turn order to read that field from `G`.

package.json
```diff
@@ -72,7 +72,7 @@
     "babel-watch": "^2.0.7",
     "chess.js": "^0.10.2",
     "coveralls": "^3.0.0",
-    "cross-env": "^5.1.1",
+    "cross-env": "^5.1.3",
```
nicolodavis: Please revert this and the change to package-lock.json.

nicolodavis:
I'm assuming that you want to close this in favor of https://github.com/google/boardgame.io/pull/130.

PR #130, opened 2018-03-01
Custom turn order

rzulian:
Current turnOrder implementation is assuming the order of the players for the round-robin [0 , 1, 2, 3, .]. Many board games are changing the player order in each phase, based for example on the pass order of the previous phase, or on the victory points at the end of the phase. 
I used G.playOrder to keep track of the real turn order. In Turnorder.order you can set the playOrder, or you can, for example on the OnPhaseBegin, set G.playOrder to your specific order.


src/core/turn-order.js
```diff
@@ -44,8 +44,34 @@ export const TurnOrder = {
    * The default round-robin turn order.
    */
   DEFAULT: {
-    first: (G, ctx) => ctx.currentPlayer,
-    next: (G, ctx) => (+ctx.currentPlayer + 1) % ctx.numPlayers + '',
+    order: (G, ctx) => {
```
nicolodavis: Let's not change `DEFAULT`, or any of the other turn orders in this file. How about creating a new one:

```
CUSTOM: {
  first: G => G.playOrder[0] + '',
  next: (G, ctx) => ...
}
```

I don't understand the purpose of `order`.

Also, we should probably make the fieldname `playOrder` customizable. Maybe make `CUSTOM` accept a parameter as the field name and have it return the turn order implementation?

```
CUSTOM: (fieldname = 'turnOrder') => ({
  first: G => G[fieldname][0] + '',
  next: (G, ctx) => ...
})
```

which can be used like:

```
Game({
  moves: { ... }
  flow: {
    turnOrder: TurnOrder.CUSTOM(),
  }
})
```

rzulian:
Hi Nicolo, 
I read your suggestions and I'm proposing I different solution.
I created a new `ctx.playerOrder ` which is populating the array `onPhaseBegin`. TurnOrder.DEFAULT and TurnOrder.SKIP are using playOrder instead of the ctx.currentPlayer. If you don't change playOrder the result is the same, but if you need a different order based on the result of the previous phase or based on a draft, you can use DEFAUL and SKIP to go through your new order.
I've created a `TurnOrder.ONETIME`, which is useful when you need to do one turn only per player and the terminate the phase.
I've also modified the `passMap` to an array which is storing the pass order. this array can be used on a new phase to determine the new playOrder.
The being said, I'm still missing a nice way to manipulate the ctx.playOrder. It would be nice that you can use onPhaseBegin and onPhaseEnd.

rzulian:
@nicolodavis : What id we set the `currentPlayer = 'none'` at the end of ANY and ONESHOT, and we automatically use this as a condition to end the Phase in flow.js.375     `
// End phase if condition is met.
    const end = conf.endPhaseIf(G, ctx);
`?

rzulian:
@nicolodavis Is there anything I can do to help you in this pull request?

nicolodavis:
Nothing for now. Sorry about the delay. I'll only have time next week to take a look.

src/core/game.test.js
```diff
@@ -100,23 +101,20 @@ test('serpentine setup phases', () => {
...
+          turnOrder: TurnOrder.ONETIME,
         },
         {
           name: 'second setup round',
-          turnOrder: {
-            first: (G, ctx) => ctx.numPlayers - 1 + '',
-            next: (G, ctx) => (+ctx.currentPlayer - 1) % ctx.numPlayers + '',
+          turnOrder: TurnOrder.ONETIME,
+          onPhaseBegin: (G, ctx) => {
+            ctx.playerOrder = [3, 2, 1, 0];
```
nicolodavis: All the move and flow functions are meant to just return a new version of `G`. `ctx` is a read-only object.

We should find another way to update the play-order here. One way might be to create a new event (similar to `endTurn`) whose only purpose is to update `playOrder`.

src/core/flow.js
```diff
@@ -326,7 +333,7 @@ export function FlowWithPhases({
 
     // End the new phase automatically if necessary.
     // In order to avoid infinite loops, this is called
-    // a finite number of times.
+    // a finite :number of times.
```
nicolodavis: typo?

src/core/turn-order.js
```diff
@@ -37,15 +37,53 @@ export const Pass = (G, ctx) => {
...
+
+  /**
+   * ONETIME
+   *
+   * The default round-robin turn order only one time.
+   */
+  ONETIME: {
+    first: (G, ctx) => ctx.playerOrder[0] + '',
+    next: (G, ctx) => {
+      if (playerOrderPos(ctx) < ctx.numPlayers - 1) {
```
nicolodavis: What happens when this function returns `undefined` (which happens when this `if` condition is false)?

src/core/game.test.js
nicolodavis: But that's not ideal either because you can't call it from within a move.

Overall, I still think my suggestion above for a `TurnOrder.CUSTOM` is simpler and doesn't add any additional complexity to `flow.js` which is already getting a bit hard to follow. Is there any type of game that you want to implement that you can't using that?

rzulian: A lot of games are using a simple round-robin turn order. Sometimes the only requirement is to rotate the first user from phase to phase, but always with the same default order (i.e 0,1,2,3), which you already covered with the startingPlayerToken example in game.test.js. But this is always assuming that the player order is the joining order. 

Terra Mystica, Gaia Project and others have a different turn order mechanism. Usually the setup phase is requiring something like 0,1,2,3,3,2,1,0 , or if some races are in play 0,1,2,2,1,0,0,3.
Then during the play phases, the pass order will determine the new play order.
Some boardgames are basing the new turn order on the victory points or the resources on the previous phase.

When you are playing tournaments, usually you have an initial draft phase, where you bet victory points to select your turn order.

The only option I'm seeing, if we want to keep the current implementation and reject this pull request, is to store the playOrder in G and use it in a TurnOrder.CUSTOM. What do you prefer?

nicolodavis: Right.

I guess my question is does `TurnOrder.CUSTOM` cover all these cases? If it does, I think I would prefer that because it doesn't touch anything in `flow.js`.

rzulian: I guess it does. I will remove playOrder from ctx and implement it as a G property

rzulian: What if we use G.playerOrder and keep using the TurnOrder.DEFAULT ,  TurnOrder.REVERSE, TurnOrder.SKIP for the turn behaviour ?  Something like 
 ````
  DEFAULT: {
    first: (G, ctx) => {
      if (G.playerOrder !== undefined && G.playerOrder.length > 0) {
        return G.playerOrder[0] + '';
      } else {
        return ctx.currentPlayer;
      }
    },
````

Comment (nicolodavis) [reply src/core/game.test.js]: I think it's better to be more explicit with a new turn order rather than have the existing turn orders be affected by a field in G.

Comment (rzulian) [reply src/core/game.test.js]: Yes, I agree with you.My only concern is that we will end up adding `TurnOrder.CUSTOM_DEFAULT`, `TurnOrder.CUSTOM_SKIP` and the `TurnOrder.CUSTOM_ONETIME`

Comment (nicolodavis) [reply src/core/game.test.js]: Oh, I see.

In that case, how about we parameterize `DEFAULT`, `SKIP` and `ONETIME`?

In the case that no additional argument is provided, they behave as they do today. In case an argument is passed, a field in `G` by the same name is used to determine the turn order instead.

**Comment by nicolodavis** [comment]
Also just wanted to bring #141 to your attention. Let's make sure that whatever design we pick here will allow someone to implement that in the future.

File: src/core/game.test.js (hunk 2 above)
Comment (nicolodavis) [reply src/core/game.test.js]: Actually, after thinking about this a bit more, how about we use something like the proposal in #142 in order to set `playerOrder` in `ctx` (through an event called `Event.setPlayerOrder()`), and then use it like you are currently?

That way, we don't leverage anything in `G` while still providing the developer with a way to manipulate the turn order (which is stored entirely in `ctx`) from within a move.

File: src/core/turn-order.js [hunk 5]
```diff
@@ -78,4 +88,23 @@ export const TurnOrder = {
...
+    // -        for (let i = 0; i < ctx.numPlayers; i++) {
+    // -          playOrder.push(i);
+    // -        }
+    // -        G = { ...G, playOrder };
+    // -      }
+    // -      return G;
+    first: G => G.playerOrder[0] + '',
+    next: (G, ctx) => {
+      let playerOrderPos = G.playerOrder.findIndex(
+        nextPlayer => nextPlayer == ctx.currentPlayer
```
akaguny: why not strict equal??

rzulian: currentPlayer is a string (usually), passOrder and playerOrder a array of integer

rzulian:
@nicolodavis reverted to playerOrder in ctx. Should be compatible with the new Events. converted passMap to passOrder as array

nicolodavis:
Great! Will take a look soon.

src/core/turn-order.test.js
```diff
@@ -144,3 +147,29 @@ test('override', () => {
...
   expect(state.ctx.currentPlayer).toBe('5');
 });
+
+test('custom order', () => {
+  const flow = FlowWithPhases({
+    phases: [
+      {
+        name: 'A',
+        turnOrder: TurnOrder.DEFAULT,
+        onPhaseBegin: (G, ctx) => (ctx.playerOrder = [2, 0, 1]),
```
nicolodavis: Can you set `state.ctx.playerOrder` down in line 167 (instead of passing `undefined`) and remove `onPhaseBegin`?

src/core/turn-order.js
```diff
@@ -44,8 +44,11 @@ export const TurnOrder = {
    * The default round-robin turn order.
    */
   DEFAULT: {
-    first: (G, ctx) => ctx.currentPlayer,
-    next: (G, ctx) => (+ctx.currentPlayer + 1) % ctx.numPlayers + '',
+    first: (G, ctx) => ctx.playerOrder[0] + '',
+    next: (G, ctx) => {
+      let playerPos = ctx.playerOrder.findIndex(x => x == ctx.currentPlayer);
```
nicolodavis: This will break if we want to support duplicate entries in `playerOrder`. Instead, why don't we store `playerOrder` and `_playerOrderIndex` in `ctx`? `_playerOrderIndex` is just an integer that's incremented normally that's an index into the `playerOrder` array.

Also, I have a mild preference to name it `playOrder` rather than `playerOrder`.

src/core/turn-order.js
```diff
@@ -10,19 +10,19 @@
...
+    passOrder = G.passOrder;
   }
   const playerID =
     ctx.currentPlayer === 'any' ? ctx.playerID : ctx.currentPlayer;
-  passMap[playerID] = true;
-  G = { ...G, passMap };
-  if (Object.keys(passMap).length >= ctx.numPlayers) {
+  passOrder.push(playerID);
+  G = { ...G, passOrder };
+  if (passOrder.length >= ctx.numPlayers) {
```
nicolodavis: This should probably be changed to `passOrder.length >= ctx.playerOrder.length` so that it doesn't break when `playerOrder` contains duplicate entries.

src/core/turn-order.js
rzulian: @nicolodavis OK. Is there any particular reason to stick with currentPlayer as string instead of an integer?

src/core/turn-order.js
rzulian: passOrder should be the same size of numPlayers, when all players passed

src/core/turn-order.test.js
rzulian: I've tried but wasn't able to move this. What if we wait for events to manage this test properly?

src/core/turn-order.js
nicolodavis: So that we can change to actual string ID's (like "Bob" or "Alice") in the future easily.

src/core/turn-order.test.js
nicolodavis: I fixed this for you.

PR #139, opened 2018-03-11
allow optional redux enhancer

philihp:
Adds an optional param to a Client which allows the user to hook into the Redux store with an enhancer or middleware.

The main motivation for this is to hook into it with [`redux-logger`](https://github.com/evgenyrodionov/redux-logger) or [Redux Devtools](http://extension.remotedev.io/), and is documented in an additional page.

#### Checklist

### DIFF
--- a/src/client/react.js
+++ b/src/client/react.js
@@ -23,13 +23,21 @@ import { Client as RawClient } from './client';
  *                                  to make a multiplayer client. The second
  *                                  syntax specifies a non-default socket server.
  * @param {...object} debug - Enables the Debug UI.
+ * @param {...object} enhancer - Optional enhancer to send to the Redux store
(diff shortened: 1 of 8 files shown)


src/client/react.js
```diff
@@ -65,6 +73,7 @@ export function Client({ game, numPlayers, board, multiplayer, debug }) {
         multiplayer,
         gameID: props.gameID,
         playerID: props.playerID,
+        enhancer: enhancer,
```
nicolodavis: Can just omit the value (similar to some of the fields above).

philihp: oh totally, sure

philihp: @nicolodavis fixed

PR #140, opened 2018-03-12
Enabling server-side-rendering for boardgame.io (fixes #22)

### LINKED ISSUE #22: SSR: Server-Side Rendering
Hello,

I'm currently trying to use boardgame.io together with next.js.
Because the `Server` part exports a Koa server, I went with this example to create a basic project:
https://github.com/zeit/next.js/tree/canary/examples/custom-server-koa

However, when I run the code, I get a `window is not defined` error.
So I did some research and it's of course because Next.js is SSR and the `Client` component requires `window` to be present. So I made it like this:
```javascript
const App = Client({
    game: Game,
    multiplayer: true,
});

export default class Index extends React.Component {
    state = {
        isClient: false,
    };

    componentDidMount() {
        this.setState({
            isClient: true,
        });
    }

    render() {
        if (this.state.isClient) {
            return <App />;
        }

        return 'Loading...';
    }
}
```

According to next.js, the client side code only runs, after `componentDidMount` has been triggered, so obviously `window` should be available, but somehow it isn't. Something inside `Client` seems to run before that happens.

Felizardo:
Fixes #22 !!!

- Creates test for making sure we are not throwing errors when doing SSR. (Cool stuff that jest can emulate node!)
- Replaces react-json-view with simple \<pre\> and JSON (react-json-view throws "window is not defined")
- Avoid establishing socket.io connection to the server if you are, er, executing on the server.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2715,11 +2715,6 @@
       "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
       "dev": true
     },
-    "base16": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/base16/-/base16-1.0.0.tgz",
-      "integrity": "sha1-4pf2DX7BAUp6lxo568ipjAtoHnA="
-    },
     "base64-arraybuffer": {
       "version": "0.1.5",
       "resolved": "https://registry.npmjs.org/base64-arraybuffer/-/base64-arraybuffer-0.1.5.tgz",
@@ -3421,6 +3416,8 @@
       "version": "1.7.1",
       "resolved": "https://registry.npmjs.org/clipboard/-/clipboard-1.7.1.tgz",
       "integrity": "sha1-Ng1taUbpmnof7zleQrqStem1oWs=",
+      "dev": true,
+      "optional": true,
       "requires": {
         "good-listener": "1.2.2",
         "select": "1.1.2",
@@ -4218,7 +4215,9 @@
     "delegate": {
       "version": "3.1.3",
       "resolved": "https://registry.npmjs.org/delegate/-/delegate-3.1.3.tgz",
-      "integrity": "sha1-moJRp3fXAl+qVXN7w7BxdCEnqf0="
+      "integrity": "sha1-moJRp3fXAl+qVXN7w7BxdCEnqf0=",
+      "dev": true,
+      "optional": true
     },
     "delegates": {
       "version": "1.0.0",
@@ -5582,14 +5581,6 @@
         "bser": "2.0.0"
       }
     },
-    "fbemitter": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/fbemitter/-/fbemitter-2.1.1.tgz",
-      "integrity": "sha1-Uj4U/a9SSIBbsC9i78M75wP1GGU=",
-      "requires": {
-        "fbjs": "0.8.15"
-      }
-    },
     "fbjs": {
       "version": "0.8.15",
       "resolved": "https://registry.npmjs.org/fbjs/-/fbjs-0.8.15.tgz",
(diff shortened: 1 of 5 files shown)


Stefan-Hanke:
Why is that node environment necessary? Jest doesn't use node?

Felizardo:
@Stefan-Hanke Jest by default uses a browser-like environment with JSDOM, which means that it has a window object available as global. See here: https://facebook.github.io/jest/docs/en/configuration.html#testenvironment-string  

With this test, we want to specifically catch those usages of window or other global objects that won't run on a node.js server. More specifically, react-json-view using the window object and trying to connect to the server using socket io were the culprits of boardgame.io not working for SSR before.

nicolodavis:
Awesome!

Will take a closer look soon. In the meantime, did you check that Mousetrap doesn't cause any problems for SSR?

Felizardo:
@nicolodavis It is passing the test with the debug view enabled, so it does not seem to be an issue :man_shrugging:

nicolodavis:
Then I'm curious why `react-json-view` was the problem? I'm OK getting rid of the dependency, but I'm just wondering.

Felizardo:
They are applying their styles using window object, and there is other people with the same issue:
https://github.com/mac-s-g/react-json-view/issues/121

Another option would be to help them making their library work with SSR too, but I am kind of prioritizing the AI stuff now :P

Felizardo:
Btw, this PR is already live in my website:
`curl https://turnato.com`
`curl https://turnato.com/g/chess`

Felizardo:
@nicolodavis Another option would be to render an empty debug view for SSR, so it does not render react-json-view and maybe does not cause this issue. Let me try it here :)

nicolodavis:
That's what I tried in https://github.com/google/boardgame.io/commit/252ab4055dd92ca0a8ca02da919f255a8786995b (now reverted).

I think I was running into some Rollup issues as well. You can verify that by running:

```
$ npm run prepare
$ npm run docsify:update
$ npm run docsify
```

and then navigating to the tutorial page to see that the examples work properly.

Felizardo:
No luck with this approach indeed. I changed the code to be like this:
```javascript
    let gJSON = (<pre>JSON.stringify(this.props.gamestate.G, null, 2)</pre>);
    let ctxJSON = (<pre>JSON.stringify(this.props.gamestate.ctx, null, 2)</pre>);
    if (typeof window !== 'undefined') {
      gJSON = (<ReactJson
                src={this.props.gamestate.G}
                name="G"
                enableClipboard={false}
                displayDataTypes={false}
               />);
      ctxJSON = (<ReactJson
                  src={this.props.gamestate.ctx}
                  name="ctx"
                  enableClipboard={false}
                  displayDataTypes={false}
                 />);
    }
```
And that's what I got:
```
 FAIL  src/client/react.ssr.test.js
  ● Test suite failed to run

    ReferenceError: window is not defined
      
      at node_modules/react-json-view/dist/main.js:1:150634
      at node_modules/react-json-view/dist/main.js:1:150596
      at e.exports (node_modules/react-json-view/dist/main.js:1:151098)
      at Object.<anonymous> (node_modules/react-json-view/dist/main.js:1:146290)
      at t (node_modules/react-json-view/dist/main.js:1:466)
      at Object.defineProperty.value (node_modules/react-json-view/dist/main.js:1:41419)
      at t (node_modules/react-json-view/dist/main.js:1:466)
      at Object.defineProperty.value (node_modules/react-json-view/dist/main.js:1:40080)
      at t (node_modules/react-json-view/dist/main.js:1:466)
      at t.exports (node_modules/react-json-view/dist/main.js:1:823)
```

Felizardo:
@nicolodavis I did run `npm run prepare` and `npm run prepublishOnly` with no problems after these changes to generate the content of this branch https://github.com/Felizardo/boardgame.io/tree/package that I use as dependency on https://github.com/Felizardo/turnato

Stefan-Hanke:
@Felizardo The people over at react-json-view got it working by [delay loading it](https://github.com/mac-s-g/react-json-view/issues/121#issuecomment-352346007).

Felizardo:
@Stefan-Hanke I saw that, but I think it is hack-ish and also I am not sure if it will work well with our rollup setup (the rollup project says we should avoid mixing require() and import statements)

nicolodavis:
OK, let's just remove `react-json-view` for now then. It's also a fairly large dependency. We can replace it with a SSR-compatible renderer in the future (or just write some code ourselves to enable collapsing / expanding of keys).

PR #144, opened 2018-03-12
Run endPhase event (analogue to endTurn) when game ends

Stefan-Hanke:
This PR changes `FlowWithPhases` `processMove` by also running the _endPhase_ event when the game ends. I've run across this while reading the code.

The intention is to streamline the flow of events. Currently, when the game ends, only the _endTurn_ event is fired. However, I see no reason not to run the _endPhase_ event also.

Also, since I was on it, I've added some assertions to bring coverage to full 100%. If that was really necessary or not, I don't know. 

#### Checklist

* [ x] Use a separate branch in your local repo (not `master`).
* [ x] Test coverage is 100% (or you have a story for why it's ok).

### DIFF
--- a/src/client/multiplayer/multiplayer.test.js
+++ b/src/client/multiplayer/multiplayer.test.js
@@ -26,6 +26,11 @@ class MockSocket {
   }
 }
 
+test('Multiplayer defaults', () => {
+  const m = new Multiplayer();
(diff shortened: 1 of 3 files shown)


nicolodavis:
Wonderful. Thanks!

PR #143, opened 2018-03-12
move player check from server to flow

Stefan-Hanke:
This PR will move the active player check from the server to the flow implementation.

This PR is part of #141.

#### Checklist

* [x] Use a separate branch in your local repo (not `master`).
* [x] Test coverage is 100% (or you have a story for why it's ok).

Issue #141, opened 2018-03-12
Ability to change currentPlayer without ending turn


nicolodavis:
There is currently no way to change the current player without ending the turn. However, sometimes you might need other players to execute an action as part of **your** turn.

janosimas:
Maybe create the concept of a sub-turn, phases could have possible sub-turns as they have moves

nicolodavis:
If we can do it without an additional concept, that would be even better, though. It's less cognitive overload when you only have to think about phases, turns and moves. A move is already essentially a sub-turn. Right now all the moves in a turn have to be made by the same player. This feature request will enable a turn to contain moves from different players.

janosimas:
This way it will also need a way of blocking and requesting moves from other players.

Stefan-Hanke:
We could have an _active player_ that is the owner of the current turn, and a _action player_ that is supposed to make a move. It would be beneficial to allow multiple _action players_. The logic for this should reside in `Flow` or the `Game` itself s.t. games can change this behaviour.

nicolodavis:
I like that idea.

nicolodavis:
I also imagine that #142 will be relevant here.

Stefan-Hanke:
We'd have to modifiy [the server](https://github.com/google/boardgame.io/blob/master/src/server/index.js#L61), where the behaviour (`any` player etc.) is currently hard coded. At that point, `game` is available, and thus the `flow`, so we could move that behaviour to a flow method?

philihp:
So the `currentPlayer` and the `activePlayer` are different things, yeah? And by default they're always the same, but the `currentPlayer` is the owner of the current turn, while the `activePlayer` is the person with the initiative, and the game is awaiting their move?

I like it.

Stefan-Hanke:
I'd like to rename _currentPlayer_ to _turnOwner_. The intention is to reduce potential confusion with the player that is supposed to make a move. Since this is a breaking change, we could defer that for now. Thoughts?

@philihp good idea about the default state of _activePlayer_. As stated above, when modifying this, we could just as well allow multiple active players, to enable games that require action from multiple "other" players. The default then is an array with only the _currentPlayer_.

nicolodavis:
Yes, let's add no additional burden to users that don't care about separating these two concepts, i.e. *activePlayer* should default to an array with *currentPlayer* (*turnOwner*) in it.

[...]

PR #145, opened 2018-03-12
Add notion of actionPlayers

Stefan-Hanke:
This PR adds the notion of _actionPlayers_ to separate the current player from those who can take actions. It subsumes PR #143.

For discussion, see #141.

I'm unsure about the coverage results - somehow, the graphical result was some characters off. I'll fix when the coverage drops.

#### Checklist


src/core/flow.js
```diff
@@ -44,12 +47,23 @@ export function Flow({
...
   optimisticUpdate,
+  canMakeMove,
 }) {
   if (!ctx) ctx = () => ({});
   if (!events) events = {};
   if (!init) init = state => state;
   if (!validator) validator = () => true;
   if (!processMove) processMove = state => state;
+  if (!canMakeMove) {
+    canMakeMove = (G, ctx, playerID) => {
```
nicolodavis: `validator` above does a similar thing (deny a move if it's not valid in the current phase). How about we combine the two into a single `canMakeMove` function (also call this in reducer.js)?

src/core/flow.js
```diff
@@ -44,12 +47,23 @@ export function Flow({
...
+      const actionPlayers = ctx.actionPlayers || [];
+      return (
+        playerID === ctx.currentPlayer ||
```
nicolodavis: Sometimes you may not want the *currentPlayer* to play (even if they own the turn). Let's just use *actionPlayers* here.

src/core/flow.js
```diff
@@ -44,12 +47,23 @@ export function Flow({
...
+        actionPlayers.includes(playerID) ||
+        actionPlayers.includes('any')
```
nicolodavis: How about we get rid of the concept of `any` (it's a bit of a hack)? We can simulate it by just adding all possible playerID's to `actionPlayers`.

src/core/flow.js
```diff
@@ -418,12 +437,15 @@ export function FlowWithPhases({
...
 
   function processMove(state, action, dispatch) {
-    // Update currentPlayerMoves.
-    const currentPlayerMoves = state.ctx.currentPlayerMoves + 1;
-    state = { ...state, ctx: { ...state.ctx, currentPlayerMoves } };
-
     const conf = phaseMap[state.ctx.phase];
 
+    const currentPlayerMoves = state.ctx.currentPlayerMoves + 1;
+    const actionPlayers = conf.turnOrder.actionPlayers(G, state.ctx);
```
nicolodavis: I would imagine that the `actionPlayer` would change in response to some event (someone played the Militia card in Dominion, for example). Rather than have that logic sit inside a `TurnOrder`, I think it's more natural inside a move:

```
playMilitia(G, ctx) {
   Events.setActionPlayers(...);
}
```

I'm not asking you to implement https://github.com/google/boardgame.io/issues/142 in this PR (you can if you like), but let's make sure that this PR is compatible with that approach.

nicolodavis: Actually, this might also suggest not retiring 'any' so that it's easy for the user. At the very least, we should convert it to an enum (constant) rather than a hardcoded string.

nicolodavis:
@rzulian FYI, just to make sure you guys don't step over each other since you are modifying turn order code concurrently.

Stefan-Hanke:
@rzulian This PR will only add the notion of _actionPlayers_, nothing else. I'm going to reverse the changes in `TurnOrder` in favor of the yet-to-come `Events` API, allowing direct control of _actionPlayers_ inside moves.

src/core/flow.js
Stefan-Hanke: OK.

src/core/flow.js
Stefan-Hanke: OK.

src/core/flow.js
```diff
@@ -9,6 +9,36 @@
...
+ * @param {object} ctx   - The ctx instance
+ * @param {object} opts  - Options - used here to transport the playerID.
+ */
+function canMakeMoveDefault(G, ctx, opts) {
+  const { playerID } = opts || {};
+
+  // In multiplayer mode, the default playerID is null, which corresponds
+  // to a spectator that can't make moves.
+  if (playerID === null) {
+    return true;
```
Golmote: Hello! I was reading through the codebase and stubbled across this line. The `true` value here seems in contradiction with the comment right before. If the spectator can _not_ make moves, shouldn't this return `false` instead?

nicolodavis: This is now fixed. Thanks for the report.

PR #146, opened 2018-03-13
Make Random package not depend on global state

nicolodavis:
#### We have the following problem at the moment:
The `Random` package is bundled into `react.js`, `core.js` and `client.js`. When the user calls the API, they use the version in `core.js`. However, when the framework tries to persist the PRNG state, it uses the version in `react.js` (if using a React client). This loses the "global" state, and makes the Random package always generate the same numbers, regardless of which seed is used.

#### Fix

This PR injects the Random API into `ctx` instead, so you would use it like:

```
moves: {
  rollDie: (G, ctx) => ({ ...G, die: ctx.random.D6() })
}
```

This also has the nice property that the individual move sub-reducers are now pure.

#### Checklist


nicolodavis:
@Stefan-Hanke @sladwig Can you take a look?

src/core/reducer.js
```diff
@@ -84,11 +86,12 @@ export function createGameReducer({ game, numPlayers, multiplayer }) {
         }
 
         // Init PRNG state.
```
Stefan-Hanke: That comment is out of date.

src/core/flow.js
```diff
@@ -278,7 +278,8 @@ export function FlowWithPhases({
   const startTurn = function(state, config) {
     const ctx = { ...state.ctx };
     const G = config.onTurnBegin(state.G, ctx);
-    const _undo = [{ G, ctx }];
+    const { random, ...ctxWithoutAPI } = ctx; // eslint-disable-line no-unused-vars
```
Stefan-Hanke: I'd like to hide how `random` is attached to `ctx`. How about a function inside _random.js_ that performs that action?

How does this scale to, say, the Events API? Does this have to be attached/stripped likewise?

Stefan-Hanke:
Looks ok to me.

I had expected you'd need to change the JSON inside _index.test.js_ also, but then, why should it change; the `seed` needs to be stored anyway...

sladwig:
Looks goot to me to, will test it later in game.

src/core/flow.js
nicolodavis: Done. The Events API will probably have a similar `attach / detach` mechanism for ctx.

src/core/reducer.js
nicolodavis: Fixed.

PR #148, opened 2018-03-15
Correct import typo in multiplayer docs

zwliew:
#### Checklist

### DIFF
--- a/docs/multiplayer.md
+++ b/docs/multiplayer.md
@@ -9,7 +9,7 @@ sync in realtime (without any browser refreshes needed).
 ```js
 // src/App.js
 
-import Client from 'boardgame.io/react';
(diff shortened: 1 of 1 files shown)


nicolodavis:
Thanks!

nicolodavis:
Just curious, was everything else working for you? There is a discussion at https://github.com/google/boardgame.io/issues/138 about some parts of the docs that aren't working for some people.

zwliew:
Yep, it all works fine for me. Admittedly, I set up and configured Babel a little differently and I hadn't tried running the Client and the Server on the same Node server.

zwliew:
I do think that the very first part (adding 'multiplayer: true') should be placed after the part on setting up a separate server though.

Issue #95, opened 2018-02-04
Undo / Redo


nicolodavis:
Features planned:
- [x] Undo moves in the current turn.
- [x] Redo moves in the current turn.
- [x] Mark certain moves as not undo-able (to prevent peeking at a card, for example).
- [x] Reset the entire game state.
- [ ] Move Undo/Redo out of game events (make them analogous to reset #155).

karel1980:
Other possibly related features:
Other feature ideas (inspired by online-go.com):
- forking (creating a new game based on a specific game state)
- reviewing: Commenting on moves after the fact
- analysis: Simulate moves to inspect the outcome (not very useful if there is randomness involved)
- planning conditional moves (if x happens, y in my next move - useful for slow paced, correspondence style games)

karel1980:
Just to be clear: these are just brainstorming suggestions, not features I really want or need right now. There's no need for implementation, but maybe they are things which can be taken into account when designing the undo/redo functionality.

sladwig:
Since most (all?) of the attributes you can add to the flow object, you can also add to the individual phases, my first attemt was to add `undo: true` to the phase where undo makes sense and so only 'applying' undo to the moves in that phase. I was a bit surprised to find out that I could only apply undo to the whole game, thus making undo not yet usable out of the box for me.

I do however like the idea of allowing undo only on specific moves, being more ganular and probably more useful for other games, where you can do undoable and not undoable moves in one phase. I don't have a specific case in mind, but guess there are games out there which do. 

My idea would be to use an array similar to 'allowedMoves' e.g. 'undoableMoves', which could be added generally to the flow object or to be more specific to each phase, allowing moves to be undoable in one phase but not in another. For some games it might even be interesting for undoableMoves to take a function like `(G, ctc) => [undoable Moves]` to decide undoable moves based on the gamestate, but not sure if this is really necessary... So what do you think?

I do have some time on my hands in the next couple of weeks, so could work on this.

Stefan-Hanke:
I like how the current Dominion implementation handles Undo: It asks all other players for permission. If everyone is OK with the Undo, it is performed.

sladwig:
Sounds good, definitely a good scenario to keep in mind. How I see it we shouldn't impose a specific way to undo moves though. 
At Backgammon it happens sometimes, that a player moves a stone and then sees another move makes more sense and as long as you are not finished with both moves, you are free to change the first one as you please. So the first move is undoable the second is not...
[...]

PR #149, opened 2018-03-15
undoableMoves implementation

sladwig:
This PR adds the ability to specify only specific moves to be undoable in undoableMoves.

See discussion in #95 

#### Checklist


src/core/flow.js
```diff
@@ -207,9 +210,12 @@ export function FlowWithPhases({
   if (endTurn === undefined) {
     endTurn = true;
   }
-  if (undo === undefined) {
+  if (undo === undefined && undoableMoves === undefined) {
```
nicolodavis: How about we just get rid of the `undo` boolean? We can use the following semantics instead:
- To make all moves undoable:
`undoableMoves === undefined`

- To make no move undoable:
`undoableMoves === []`

- To make some moves undoable:
`undoableMoves === [ ... ]`

nicolodavis: Once we adopt these semantics, I think that will also pave the way to having consistency between `undoableMoves` and `allowedMoves`. For example, we can implement both at a global and per-phase level easily.

src/core/flow.js
```diff
@@ -455,9 +466,13 @@ export function FlowWithPhases({
     // Update undo / redo state.
     if (!endTurn) {
       const undo = state._undo || [];
+      const moveType =
+        action !== undefined && action.payload !== undefined
```
nicolodavis: When is `action` ever `undefined`?

sladwig: Basically both checks are there because of failing tests, for me it looks like this wouldn't happen 'in production', but I am not sure and didn't want to mess with other tests.
Should I update the failing test(s) instead?

    FAIL  src/core/flow.test.js
    ● onMove
    
    TypeError: Cannot read property 'payload' of undefined
      
      at processMove (src/core/flow.js:469:29)
      at Object.processMove (src/core/flow.js:85:14)
      at Object.<anonymous> (src/core/flow.test.js:190:18)
          at new Promise (<anonymous>)
          at <anonymous>

src/core/flow.js
sladwig: Yes I like it a lot, but was a bit reluctant to remove ‘undo‘ without consent. 
So basically we come with undo/redo out of the box and a developer would have to restrict it, if it's not needed. Sounds great to me!

src/core/flow.js
sladwig: so since [#151](https://github.com/google/boardgame.io/pull/151/files) also assumes action and payload to be there, I just fixed the test, by giving them what they needed. If it's not what you're expecting, I will change it...

PR #155, opened 2018-03-18
Reset Game

sladwig:
This PR adds a new resetGame event, which as the name may suggest resets the game.

See discussion in #95 

I was not sure where to add documentation for this, will update the docs when I know where.

#### Checklist


src/core/reducer.js
```diff
@@ -106,6 +106,10 @@ export function createGameReducer({ game, numPlayers, multiplayer }) {
         // Detach Events API from ctx.
         ctx = Events.detach(ctx);
 
+        if (action.payload.type === 'resetGame') {
```
nicolodavis: Given that this event needs special treatment here, why not just make it analogous to `RESTORE` below?

sladwig: Since it's a `gameEvent` I've treated it as such to avoid reimplementing the whole event/random setup (and everything else regading events) and so haven't given the `RESTORE` option much thought. I will try implementing it analogous to `RESTORE` and see how it looks.

sladwig: I had a quick look, my first impression is that it's far more special treatment, when we don't treat `resetGame` as a normal `gameEvent`. As I understand it `RESTORE` is used for the debug component and is not really a `gameEvent` like `endGame`, thus not (directly) available from within the game, when the debug component is deactivated. So if I understand correctly I would have to add extra logic to the flow object (and introducing API and stuff) just to support 'another type' of game event just to reset the game. I am not sure if we want this (correct me if I'm wrong or misunderstood what you mean with analogous to `RESTORE`).  

One simple solution could be to leave the log and stateID intact, this way we don't have a special treatment and also don't lose the history of the log and so the information that the game has been reset. Also stateID would be 'unique', so stateID 1 would be always the first move of the first game and never the state after a reset.

nicolodavis: Actually, not being available from within the game is a feature. A game reset would typically be invoked by some menu outside the game. The ability to reset the game from within a move is not desirable.

Here is what I had in mind:

- A simple one liner in the reducer similar to RESTORE. Just return `initial`. You don't need to do any more.

- A reset method in Client (the raw js client) that dispatches this. Again, a one liner.

- Since the React client is an actual component, it's harder to support a knob that calls reset. One option might be to add a field to the config that accepts a function that when called calls client.reset().

nicolodavis: You could argue that undo and redo ought to be implemented this way too instead of being game events since they are not game events, but one level removed.

sladwig: I see your point, but just for my peace of mind let me explain how I see it, to be sure we support this (or something similar) for the future:

It is true that, most of the games probably have their resetGame option somewhere in the menu, but when I think about playing boardgames with friends, redoing moves or reseting the game are pretty elementary "events" when playing, not much different than making a move or rolling a dice. 
Next there are games, where you play the game in matches, here a match consist of many rounds of the game, where you also maybe want to save the points from each round or something like that. Or games where the next round beginner is based on the previous round loser/winner. So it could aswell be somewhat "in the game" or at least let's say it could also need the state from the game and maybe returns a new state to set up the next round.
Then I can think of events like 'givingUp', which might be implemented by the developer directly for a specific game or games which want to implement some events e.g. undo differently (allowing just a number of redo or something like that). 
Sure one could implement all of this 'outside' of the game and introduce another concept, for the 'mostlyInTheMenuEvents', but how it is implemented now I can easily see adding a `events` property, similar to `moves` and have the developer add their custom events or own implementation of events and have them treatet like all the other events in the game (which are basically functions which take a state and return a state, for which we already have a concept).

So that said, I'm happy to implement it the way you suggest, just wanted to make sure there went some thought into it or undertand the big picture (and I don't like that special treatment part in there myself that much)...

nicolodavis: #### About resets being as common as other events
The experience to the game developer is not going to change very much (the prop is going to move from `events` to the root, that's all). We're talking about an internal implementation detail here. The end user is not going to see any change.

#### About rounds / matches
Here again, you want to make the distinction of whether you need continuity or not. For example, does the log need to be reset? Just resetting G to the initial value is not what this mechanism is for. That's trivial to do like this:

```
const initG = ...

Game({
  setup: () => initG,
  moves: {
    reset: () => initG
  }
})
```

nicolodavis: The existing event mechanism is there so that it ties things with the flow and also updates logs and such. We don't need any of those things for a reset. This is also why the restore code is so simple.

sladwig: I see now what you mean, the touching only G/ctx was the missing piece of my puzzle.

sladwig:
where should I add documentation for reset?

nicolodavis:
To the doc being written [here](https://github.com/google/boardgame.io/pull/159) (once it's submitted). We could probably expand it to explain the different types of events the framework provides (including reset and restore).

sladwig:
I like how it turns out and thanks for the discussion!

PR #159, opened 2018-03-20
Event docs

jstacoder:
#### Checklist

this is just my first take at some event specific docs, i really don't feel its done, i defiantly need to flesh out the examples more, but some early feedback would be great


jstacoder:
i signed it

docs/events.md
```diff
@@ -0,0 +1,42 @@
...
+* endTurn(G, ctx) - provided by default, to disable add `endTurn: false` to a custom flow config
+* endPhase(G, ctx) - provided by default when using phases, to disable add `endPhase: false` to a custom flow config
+* endGame(G, ctx) - only provided when `endGame: true` in a custom flow config
+* undo(G, ctx) - provided when given a list of `undoableMoves` in flow config of game
+* redo(G, ctx) - provided when given a list of `undoableMoves` in flow config of game
+* changeActionPlayers(G, ctx) - provided by default when more than one player available to change players who can currently make a move
+
+### triggering events during gameplay
+
+to trigger game events from inside of your game, your Board component is passed an object of event functions as a prop called events
```
nicolodavis: Can you capitalize these sentences and also end them with a period?

jstacoder: no prob, all done, also added a bit more content, and fixed the function signatures in commit 
 ac832a2

docs/events.md
```diff
@@ -1,42 +1,58 @@
...
-to use just call the event you want to trigger
-
-```
-this.props.events.endTurn()
+* endTurn() - Provided by default, to disable add `endTurn: false` to a custom flow config.
+* endPhase(nextPhase) - Provided by default when using phases,  
+  to disable add `endPhase: false` to a custom flow config. \* nextPhase - optional phase to transition to, if not provided defaults to next phase defined in config
+* endGame(arg) - This is only provided when setting `endGame: true` in a custom flow config.
+  * arg - This optional argument when provided will be available as ctx.gameOver.
+* undo() - Provided when given a list of `undoableMoves` in flow config of game.
```
sladwig: `undo` and `redo` is actually turned on by default, if `undoableMoves` is an empty array, then `undo` and `redo` is not provided, if it has content it restricts `undo` and `redo` to those moves provided.

sladwig: maybe also refer to the the [undo/redo](http://boardgame.io/#/undo) page for more information.

jstacoder: perfect, i guess i was thinking of the old api where if you didnt provide `undo: true` you didnt get the events, that was a big reason i wanted to do this, to clairify weird little parts of the api like that, thanks @sladwig

Issue #154, opened 2018-03-18
Turn Order Customizations


nicolodavis:
- [ ] `setActionPlayers` option that allows others to play, but following a turn order (not simultaneously).
- [ ] Retire the Pass move. It's better to just document a pattern on how to implement passing or convert it to an event.

Stefan-Hanke:
OK, let's assume we've got a Dominion implementation and want to make Militia work. For this, we'd need the first bullet point. As far as I've understood the new API, we'd need an event that must currently be directly implemented in `FlowWithPhases`, since there is currently no possibility to have events defined by a Game.

If that event existed, a game could dispatch a call to it which would be triggered during the `update` call. That event (_changeActionPlayers_ or something like that) would need the actual players that ought to perform a move.

Did I get anything wrong?

nicolodavis:
Yes, that's correct.

Stefan-Hanke:
Since this allows concurrent state modifications, we'd need a way to reconcile state on a client with one that is `sync`ed. I don't know whether it's a good idea to let the game handle that process?

_Sidenote about order of players to take actions_: Actually, Dominion must enforce an order of action players since in case of that _Militia_, dropping cards could trigger e.g. _Marktplatz_ (don't know its english name), which may give you a _Gold_, which you don't when the pile is empty. So it's not the best example to choose... sigh.

nicolodavis:
What concurrent state modification are you referring to?

Stefan-Hanke:
The kind of modification where two states need to be merged somehow. E.g. two players simultaneously click on _Drop Card_, or one player is selecting cards to drop whereas the other one already has selected etc.

nicolodavis:
That's already taken care of via the stateID mechanism is it not?

nicolodavis:
An unsynced client cannot make a move basically.

Stefan-Hanke:
Ah, maybe I don't know the specific multiplayer implementation. I'd like to avoid the case where one player loses part of his move because another is faster and comitted his...

nicolodavis:
That will only happen if the user plays before the update caused by the other players reaches them, i.e. will only be seen if there is noticeable network lag. Maybe we can build a mechanism in the future to perform a true merge, so that an out of date client can be usable in some cases, but this is what we have right now.

nicolodavis:
[...]

PR #162, opened 2018-03-20
Advanced Turn Order Example

Stefan-Hanke:
This PR adds an example for showcasing advanced turn order options.
Currently, it adds the option to change the action players.

This is NOT finished, I've pushed and created this PR in order to facilitate discussion. #154 

#### Checklist

* [x]  Test coverage is 100% (or you have a story for why it's ok).


nicolodavis:
How about we have the fields displayed just once (rather than once per player), and have the players be rendered as just boxes (with some color codes for the activePlayers and the currentPlayer)?

Stefan-Hanke:
You ok with the buttons per players? My goal is to allow to manually play through that _Militia_ example.

nicolodavis:
Sure! I was just looking for a more visual cue for the rest (rather than just writing down the fields).

Changing color, emboldening etc are all fine ways to use visual cues.

Stefan-Hanke:
Somehow, using `events` and `moves` simultaneously does not work. I'd like to have the _Play Militia_ button to reduce actions to 0, and to modify `actionPlayers`.

nicolodavis:
That's because **changeActionPlayers** loses the `stateID` [here](https://github.com/google/boardgame.io/blob/master/src/core/flow.js#L572). I have a fix for this that I will push soon.

Stefan-Hanke:
The example is now working as intended.

PR #164, opened 2018-03-21
Allowed moves as function

jstacoder:
#### Checklist

here it is only on the phase level, im having trouble getting it working on the top level in flow, i want to return all move names by default, or should it just stay undefined by default at the top level? any suggestions would be very appreicated


src/core/flow.js
```diff
@@ -217,8 +217,9 @@ export function Flow({
  *   // A phase-specific movesPerTurn.
  *   movesPerTurn: integer,
  *
- *   // List of moves that are allowed in this phase.
- *   allowedMoves: ['moveA', ...],
+ *   // List of moves or a function that returns a list of moves
+ *   // that are allowed in this phase.
+ *   allowedMoves: (G, ctx) => ['moveA', ...],
```
nicolodavis: Let's also make this available on a per-game basis (not just on a per-phase basis)?

nicolodavis:
Hijacking this thread for a moment to discuss something broader. We don't need to implement any of the following in this PR. One of us will follow up later:

We now have a few different mechanisms to disallow moves:

1. allowedMoves
2. canMakeMove
3. Returning undefined in a move function.

1 is designed to handle disabling *types* of moves. 2 and 3 are designed to disable specific instances of a move (with a particular argument combination for example). Both are valuable, but I would recommend that we only use one out of 2 and 3 to do the latter. I'm leaning toward retiring 2 (i.e. not allowing the user to override canMakeMove).

Also, we should probably expose 1 through ctx so that it is a helpful hint to the UI to figure out what moves to highlight.

@Stefan-Hanke, what do you think? 

@Felizardo, this might also be useful for the AI code. Let's consolidate what we have there as well.

jstacoder:
I think exposing it to ctx is perfect, exactly what’s needed to make it work at the top level, and I like returning undefined from a move to specifically disallow it, but I was wondering if at the top level of the config should it maybe be a `disallowedMoves` array? Which is then checked, possibly overridden in the phase level `allowedMoves`?  It just seems like the use case would be to disallow one or two moves, which would then conditionally be allowed in certain phases or specific parts of turns, but I can also see how maybe having the top level value and phase level value being different (disallowedMoves vs allowedMoves) could be confusing as well. Ideas?

nicolodavis:
Let's just use allowedMoves everywhere. It's much less confusing that way. About the semantics of the override, let's just have the phase level config override the global config whenever it is defined (don't check for undefined, but for the presence of the field, which could be set to undefined itself).

src/core/flow.js
nicolodavis: Oops, forgot to wait for this. No worries, I'll just add this myself.

Felizardo:
@nicolodavis Sorry I've been a little bit busy IRL, so I lost a little bit the track of the changes, I will try to catch up this weekend

Issue #92, opened 2018-01-26
Timer


nicolodavis:
This has come up a few times, so I think we should prioritize this.

Suggestions:
- Some timers in `ctx` that capture the time that has elapsed since the beginning of the phase and since the current turn started.
- Some hooks to automatically end the turn / phase when some amount of time has elapsed.

akaguny:
Suggestions:
* opportunity to pause a timer

akaguny:
it's can be inplemented with simple timeout or react use specific as $timeout in angular?

nicolodavis:
Yes, a simple timeout would also work (probably not necessary to record anything in `ctx`).

akaguny:
@nicolodavis what api timer should take?
when i create game i was need pause func(tur on/off activity of timer) and state variable, that take me: How much seconds left. But my time leav in React component and store state in this component
https://github.com/akaguny/alias-boardgame/blob/master/src/App.js#L54

akaguny:
>  I think we just need a simple setTimeout that is called at the beginning of the turn that ends the turn automatically when the callback is invoked. We can use a similar one for phases as well.
@nicolodavis i'm agree that we should make a simple one timer for first time. But when we configure it? Configure like maxMoves?

akaguny:
i started work on this feature, please set label tacked or some familiar

nicolodavis:
@akaguny Yes, let's start simple with something similar to `movesPerTurn` and expand it later. Will take a look at your PR. Thanks for working on this!

sedobrengocce:
Greetings, someone is woriking on this issue?
If not I can take it

nicolodavis:
@akaguny is busy right now, so feel free to take it.

nicolodavis:
My thoughts on how this should be implemented:

- At the beginning of the turn, call a `setTimeout` that results in the server sending an `endTurn` event to itself. This event should be routed through the transport interface (the socket.io handler).

I think it makes sense for this to wait for #251 so that we can create some hooks to send synthetic events that appear to come from a client, but are actually initiated from the server itself.

lw1990:
The way heroiclabs.com handles this is by using a customizable 'tick rate', as explained: "Tick rate¶
The server will periodically call the match loop function even when there is no input waiting to be processed. The logic is able to advance the game state as needed. It can also validate incoming input and kick players who've been inactive for some time.

[...]

PR #166, opened 2018-03-21
Sedobrengocce/feature.timer

sedobrengocce:
#### Checklist

* [ x ] Use a separate branch in your local repo (not `master`).
* [ x ] Test coverage is 100% (or you have a story for why it's ok).


src/core/timer.test.js
```diff
@@ -0,0 +1,52 @@
...
+    expect(functionToCallWhenTimeIsOut).not.toHaveBeenCalled();
+    jest.advanceTimersByTime(1000);
+    expect(functionToCallWhenTimeIsOut).not.toHaveBeenCalled();
+    jest.advanceTimersByTime(1000);
+    expect(functionToCallWhenTimeIsOut).toHaveBeenCalled();
+  });
+});
+
+describe('timer not initialized', () => {
+  let timer = new Timer();
```
akaguny: move create instance in beforeEach because it must be dedicated and autonomy

akaguny: move create instance in beforeEach because it must be dedicated and autonomy

src/core/flow.js
```diff
@@ -329,9 +341,23 @@ export function FlowWithPhases({
     return playOrder[playOrderPos] + '';
   };
 
+  const onPhaseBeginWrap = (state, phaseConfig) => {
```
nicolodavis: Let's just inline this inside `startPhase`.

src/core/flow.js
```diff
@@ -342,8 +368,20 @@ export function FlowWithPhases({
     return { ...state, G, ctx };
   };
 
+  const onTurnBeginWrap = (state, config) => {
```
nicolodavis: Let's just inline this inside `startTurn`.

src/core/flow.js
```diff
@@ -329,9 +341,23 @@ export function FlowWithPhases({
     return playOrder[playOrderPos] + '';
   };
 
+  const onPhaseBeginWrap = (state, phaseConfig) => {
+    if (phaseConfig.secondsPerPhase) {
+      phaseTimer.pause();
+      phaseTimer.reset();
+      phaseTimer.Routine = () => {
```
nicolodavis: I'm not sure why we need a full blown Timer class with pause and reset. Can we just use `setTimeout` here? You can use the object that it returns to erase previously running timers.

sedobrengocce: At the moment, probably you don't need a full timer class, but in future maybe you want to store some info in timer.
If you prefer i can make it a simple timeout

src/core/flow.js
sedobrengocce: ok, i'll do it, for phase and turn

nicolodavis:
This looks great!

However, I'm having trouble running the upgraded Jest properly. It seems to generate some warnings.

nicolodavis:
> What kind of warnings?

You can see them [here](https://travis-ci.org/google/boardgame.io/builds/358042955?utm_source=github_status&utm_medium=notification).

- Warning: The tag <path> is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter.

- Failed to write coverage reports:        ERROR: TypeError: metrics.isEmpty is not a function

src/core/flow.js
```diff
@@ -352,9 +360,18 @@ export function FlowWithPhases({
...
   const startPhase = function(state, config) {
-    const G = config.onPhaseBegin(state.G, ctx);
-
     const ctx = { ...state.ctx };
+    if (config.secondsPerPhase) {
+      ctx.phaseElapsedTime = 0;
+      const interval = setInterval(() => {
+        if (++ctx.phaseElapsedTime >= config.secondsPerPhase) {
+          clearInterval(interval);
+          state = endPhaseEvent(state);
```
nicolodavis: The only clean way to do this would be to dispatch the `endPhase` event to the store directly. Right now, I don't think this change would be propagated to clients correctly.

I would suggest that you revert the most recent commit about storing elapsed time to keep things simple (I don't think we need to support that; it's easy enough to have a client side timer to show the remaining time).

sedobrengocce:
About that warnings,

I agree with the first one, but know nothing about the second one...

nicolodavis:
One option would be to mock out `setTimeout` in the test so that you don't need Jest 22.

sedobrengocce:
about the empty error, i still have it, but i foud we can fix it using this
"istanbul-api": "1.2.2",
"istanbul-reports": "1.1.4",
in package.json, it works

src/core/flow.js
```diff
@@ -373,13 +385,17 @@ export function FlowWithPhases({
   };
 
   const startTurn = function(state, config) {
-    const G = config.onTurnBegin(state.G, state.ctx);
+    if (config.secondsPerTurn) {
+      setTimeout(() => {
+        endTurnEvent(state);
```
lw1990: On line 374 it reads `state = endPhaseEvent(state);`
But here on line 390, it should be doing a similar thing but it reads
`endTurnEvent(state);`

So one of these is different than the other (one is assigning to the `state` variable and one isn't), this could be causing problems?

nicolodavis:
Let's rework this since it has been a while. We can continue discussion on #92.

PR #176, opened 2018-03-27
Fix bug that was ending phase incorrectly

janosimas:
#### Checklist

#### Test

When a move is made it checks if the turn should end and if the phase should end.
In the situation of `phase A` ending the turn and the phase, it should start `phase B` end check `phase B` `endPhaseIf`.

Expected:
`phase A` - call onPhaseEnd 1 time
`phase b` - don't call onPhaseEnd

### DIFF
--- a/src/core/flow.js
+++ b/src/core/flow.js
@@ -335,7 +335,12 @@ export function FlowWithPhases({
     }
   }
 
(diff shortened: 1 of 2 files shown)


janosimas:
I signed it!

PR #184, opened 2018-04-18
Allow enhancers for react-native clients

### LINKED ISSUE #185: Enhancers don't work for React Native clients
If you use the new React Native client and pass it an enhancer to debug your local redux actions, it won't work as currently the RN client doesn't accept one to pass along to its RawClient.

scally:
The existing implementation doesn't support enhancers for react-native. This prevents redux debugging from working for RN clients.

Fixes #185

#### Checklist

### DIFF
--- a/src/client/react-native.js
+++ b/src/client/react-native.js
@@ -27,7 +27,7 @@ import { Client as RawClient } from './client';
  *   API through props for it to interact with the framework
  *   and dispatch actions such as MAKE_MOVE.
  */
-export function Client({ game, numPlayers, board, multiplayer }) {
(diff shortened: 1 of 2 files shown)


scally:
CLA has been signed.

scally:
The test is a little awkward, and might be unnecessary. I used it to TDD the feature, but having to mount the component just to ensure the enhancer is called might be too finicky. I'm happy either way.

scally:
I've also manually acceptance tested this via pointing a local example repo at these changes and verified that the RN debugger now sees actions.

Issue #171, opened 2018-03-25
Game Creation / Lobby API


scally:
I've been trying to debug an issue with a multiplayer game this weekend, and encountered the following misunderstanding:

When you provide a playerID to the server today, it *must* be a 0-indexed number corresponding to the number of players -- [0, 1] for 2 players, for instance, is the only possible space. This appears to be due to this: https://github.com/google/boardgame.io/blob/bb9ff3b6dd6c9f8ea5314d3712d2741b5d8992bc/src/core/flow.js#L563

So, if you provide a playerID outside of the range [0,1], no player can make a move. 

I can imagine a few possible solutions:

- Document that playerID must be [0,1] for 2 player games, etc. 
- Change the turn ordering code to use the turn cursor to index into the actual playerIDs
- Assign unique numerical playerIDs to connected clients after they provide a unique identifier of some sort -- username, token, whatever else. 

Otherwise, in a true non-demo multiplayer environment, how would you know which player is 0 and which is 1 without some kind of server-side arbitration?

Stefan-Hanke:
The framework needs to assume something about players in order to allow working with them. I didn't find anything in [multiplayer.md](https://github.com/google/boardgame.io/blob/master/docs/multiplayer.md), so I'd imagine we could add a parapraph about this.

There is currently no notion of an _game instance id_ that comes from the server, associating players to a specific running game. Instead, the client is assumed to provide a `gameID` which fulfills this purpose. There is a paragraph about that in the aforementioned docs. In true non-demo, the client currently is assumed to generate a unique `gameID` for each separate game instance.

Take this with a grain of salt, I didn't experiment with the code to get this knowledge, I've only read it.

nicolodavis:
I'll make the documentation clearer about this, but all the playerID/gameID management is currently the responsibility of the user. I'd like to eventually ship a lobby implementation that will do all this for you, but we aren't there yet.

playerID's are '0', '1', etc. as you rightly point out, and deliberately so. Figuring out how to map your notion of a user to these ID's is entirely up to you.

- For example, you might have your own user authentication system where the user is identified by something like "Bob" or "Alice". You need to build some sort of game creation interface where users are assigned to specific seats ("0", "1", "2" etc.) in a game. This ID can then be stored in a cookie or whatever mechanism you choose and passed via the `playerID` prop.

[...]

PR #189, opened 2018-04-25
lobby API support

scally:
This is the initial implementation of lobby API support #171 

It has two new APIs available by default on `WebSocket port + 1`.

**`/games/:id/create`**: creates a new authenticated game for a game named `id`.

Accepts one parameter: `numPlayers`, which is required & indicates how many credentials to create.

Returns `gameID`, which is the ID of the newly created game instance.

**`/game_instances/:id/join`**: allows a player to join the given game instance `id`.

Accepts two parameters, both required:
`playerID`: the ordinal player in the game that is being joined (0, 1...)
`playerName`: the display name of the player joining the game.

Returns `playerCredentials` which is the token this player will require to authenticate their actions in the future.

All actions for an authenticated game require an additional payload field: `credentials`, which must be the given secret associated to the player.

A future PR will create a simple website to handle calling out to this API for projects that would prefer a zero-config solution.

#### Things still needed
* [x] Accepting a prop on the Client object to send a secret with all actions.
* [x] New example app.
* [x] Test coverage
* [x] Cleanup/refactoring
* [x] API documentation

#### Checklist


MrRhodes:
Any reason why the lobby should run separate to the game servers?

scally:
Good question. I had assumed using the same port as the websocket listener would be a blocker. Was that assumption wrong ? Or did you mean something else?

scally:
@MrRhodes 

> Is it relating to the game name being the root of the reducer? so we could have clashes if any game is ever called "lobby" (if that was what we call the lobby)..

Well, my thinking was, the lobby is separate from any particular game, so I don't want it in the game reducer, as it seems to be created for each individual game and then stored away by key.

scally:
I'm growing less convinced that I should be saving the server data in the redux store. I'll probably finish out this PR using that approach and then consider refactoring to do something simpler unless it's overwhelmingly obvious that this is the better move.

scally:
@nicolodavis Have you had a chance to check this out yet? I'd like to get some feedback for some of the stuff I mentioned in the PR notes to make sure I'm not building something too far off base here. 😄

nicolodavis:
Sorry about the delay. Been a bit busy. Will likely get some time to look later tonight or tomorrow.

src/server/index.js
```diff
@@ -9,11 +9,76 @@
...
+  const app = new Koa();
+  const router = new Router();
+
+  router.post('/games/create', koaBody(), async ctx => {
+    // TODO: protect this key somehow -- anyone could use it for a gameID
+    const state = await db.get('serverData');
+
+    const store = Redux.createStore(serverReducer, state);
+
+    store.dispatch({
```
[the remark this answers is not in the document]
scally: Yeah, I for whatever reason it made sense to me initially to tie together retrieving data from the existing state via selector, but the more I wrote the more I realized that I wouldn't really have that use case. I'll change it soon.

scally:
@nicolodavis I've made changes based on your feedback. It's not done-done as I still need to refactor this mess and add some test coverage, but is this more on the right track before I proceed further?

src/server/index.js
```diff
@@ -9,11 +9,124 @@
...
 const PING_TIMEOUT = 20 * 1e3;
 const PING_INTERVAL = 10 * 1e3;
 
+const isActionFromAuthenticPlayer = async ({
+  action,
+  db,
+  gameID,
+  playerID,
+}) => {
+  const gameMetadata = await db.get(`${gameID}:metadata`);
```
scally: I really don't like the duplication of creating the string metadata key in multiple places, so I need to pull that out.

src/server/index.js
```diff
@@ -9,11 +9,124 @@
...
+  if (action.payload.secret !== gameMetadata.players[playerID].secret) {
+    return false;
+  }
+
+  return true;
+};
+
+const createSecret = async () => {
+  return new Promise((resolve, reject) => {
+    Crypto.randomBytes(32, (err, buf) => {
```
scally: I believe 32 bytes is the size of a typical oauth2 client secret

src/server/index.js
```diff
@@ -9,11 +9,124 @@
...
+        reject(err);
+      } else {
+        resolve(buf.toString());
+      }
+    });
+  });
+};
+
+const getNewGameInstanceID = ({ gameName }) => {
+  // TODO: Return a new instance ID for this game
```
scally: This needs to generate a random instance ID. UUID generation would probably be sufficient, if our gameIDs are strings throughout.

src/server/index.js
```diff
@@ -9,11 +9,124 @@
...
+  router.post('/games/:id/create', koaBody(), async ctx => {
+    const gameName = ctx.params.id;
+    const numPlayers = ctx.request.body.numPlayers;
+
+    const newMetaDataState = {
+      players: {},
+    };
+
+    const reducer = createGameReducer({
+      game: games[gameName],
```
scally: This is not right, IIRC games is just an array, not an object. We should probably make it an object and require unique names.

src/server/index.js
```diff
@@ -9,11 +9,124 @@
...
+      numPlayers,
+    });
+    const store = Redux.createStore(reducer);
```
scally: This code is copied from the adhoc `sync` game creation. I  left the sync game creation code there because it needs to be there to support non-authed games, but we could probably refactor up the duplication here.

src/server/index.js
```diff
@@ -9,11 +9,124 @@
...
+
+    await db.set(`${gameID}:metadata`, newMetaDataState);
+    await db.set(gameID, state);
+
+    ctx.body = {
+      gameID,
+    };
+  });
+  // TODO: Is game_instances the right resource here?
+  router.patch('/game_instances/:id/join', koaBody(), async ctx => {
```
scally: I originally called this route `games/` also, but I don't think that's quite right. One resource takes a game NAME and creates an INSTANCE. This other one takes a game INSTANCE as an ID

src/server/index.js
```diff
@@ -58,6 +173,11 @@ export function Server({ games, db, _clientInfo, _roomInfo }) {
           return;
         }
 
+        if (!isActionFromAuthenticPlayer({ action, db, gameID, playerID })) {
+          console.log('inauthentic!!!');
```
scally: Need to fix this for the final revision

scally:
Updated PR description with some basic documentation.

nicolodavis:
Looks on the right track! Don't forget to also remove the code in `sync` that currently creates a game.

scally:
@nicolodavis I left an (outdated) comment on this topic that I'd like your opinion on:

> This code is copied from the adhoc sync game creation. I left the sync game creation code there because it needs to be there to support non-authed games, but we could probably refactor up the duplication here.

nicolodavis:
Ah, ok let's just leave the code in `sync` for now then (no need to refactor).

scally:
OK, the authed multiplayer example is up and working (but ugly)

src/client/multiplayer/multiplayer.js
```diff
@@ -139,6 +139,10 @@ export class Multiplayer {
   updateGameID(id) {
     this.gameID = this.gameName + ':' + id;
 
+    const action = ActionCreators.reset();
```
scally: This was added because changing a player or game should reset the internal state and re-sync with the server

src/core/action-creators.js
```diff
@@ -14,10 +14,11 @@ import * as Actions from './action-types';
  * @param {string} type - The move type.
  * @param {Array}  args - Additional arguments.
  * @param {string}  playerID - The ID of the player making this action.
+ * @param {string}  secret - (optional) The secret for the player making this action.
  */
-export const makeMove = (type, args, playerID) => ({
+export const makeMove = (type, args, playerID, secret) => ({
```
nicolodavis: Let's call this credentials instead of secret?

scally: In this particular case, or everywhere? If everywhere, sure, happy to do it. If just in the one case, it feels a little inconsistent.

scally: @nicolodavis In this particular case, or everywhere? If everywhere, sure, happy to do it. If just in the one case, it feels a little inconsistent.

nicolodavis: Just want to avoid confusion with the game secrets that we strip elsewhere for secret state in games. Changing it everywhere in this PR sounds good.

scally:
@nicolodavis OK, I've rebased, added test coverage and documentation, and squashed. It should be ready for your review now.

scally:
@nicolodavis I also have an additional question on this PR that I don't have a good answer to. What prevents Alice from calling the join API and saying she's Bob (playerID 1), then using his token to send moves on his behalf. Right now, if you have the game instance ID, you can be either player. That puts us right back at square 1.

nicolodavis:
The lobby API just allows players to associate themselves with a particular slot in a game (player 0, 1, 2 etc.), and this PR prevents moves from being spoofed while the game is being played. The API doesn't care if it is Alice or Bob that occupies a particular slot (that is under the purview of the user authentication system, which is built above this API). It just cares that it is the same player that continues to occupy the slot for the rest of the game.

scally:
@nicolodavis 

> this PR prevents moves from being spoofed while the game is being played

I'm not sure that it does. Imagine this line of calls:

1. Alice (via the lobby server, elsewhere) requests to create a new game.

2. The lobby server calls the createGame API and creates a game instance.

3. Alice (via lobby server) requests to join the game. 

4. The lobby server calls joinGame and returns to Alice's client the player ID, game instance ID and player token required to provide to `<Client />` to connect to the game and send moves.

5. Alice calls joinGame (without using the lobby server) as Bob. - she knows she is playerID `0`, and Bob must be `1`. 

Anyone looking at the Client's traffic can see that it's opening a websocket on 8080 to communicate the moves, and a port scan can reveal the well-documented endpoint on 8081. 

The API on port 8081 doesn't know if its caller is Alice, or the lobby server, and it obligingly sends Bob's token when she calls the `joinGame` API with the game Instance ID she knows, and Bob's playerID which she also knows.

6. Alice can now send moves as Bob. 

Do you see what I mean? Do I fundamentally misunderstand some capability in this example? (it's quite possible.) 😄 )

nicolodavis:
Step [5] wouldn't be allowed:
- once the game has started.
- it's the responsibility of the lobby implementation to call it (we can disallow calling **create game** and **join game** directly unless you are the lobby), and it will ensure that Alice can't request to play as Bob.

About snooping traffic, I think the bigger issues is that the credentials are also sent over the wire and can be sniffed. We'll have to think about these things going forward (I'm hoping that people like you come up with clever solutions in this space), but this is already a good first step (I'm OK if we get this PR in and work on the improvements in the future).

scally:
@nicolodavis I'm OK to get this PR in if you're OK with it. 

The two protections you mention above will help (can't join once game has started, can't call API unless you can prove you're lobby somehow), but my point was they're not yet implemented so if this PR goes through in its current state, those cases aren't yet handled. We can defer if that's your preference.

nicolodavis:
If we can get those things in this PR in as well, that would be even better!

I just didn't want to block you in case you wanted to get this in first. Up to you :)

nicolodavis:
OK, sounds good. Will review this sometime before the end of the week. Thanks for the great work!

src/server/index.js
```diff
@@ -33,6 +36,8 @@ export function Server({ games, db, _clientInfo, _roomInfo }) {
     }
   }
 
+  const apiApp = createApiServer({ db, games });
```
nicolodavis: What do you think about decoupling the lobby API and game servers? That will give people more flexibility to choose to host them independently. The caller can then look something like:

```
const gameServer =  GameServer({ games });
const lobbyServer = LobbyServer({ games });
Run([
  { server: gameServer, port: 8000 },
  { server: lobbyServer, port: 8001 },
])
```

src/server/api-server.js
```diff
@@ -0,0 +1,112 @@
...
+
+  return true;
+};
+
+export const createApiServer = ({ db, games }) => {
+  const app = new Koa();
+  const router = new Router();
+
+  router.post('/games/:id/create', koaBody(), async ctx => {
+    const gameName = ctx.params.id;
```
nicolodavis: rename `id` to `name` to avoid confusion with `gameID`?

src/server/api-server.js
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  return true;
+};
+
+export const createApiServer = ({ db, games }) => {
+  const app = new Koa();
+  const router = new Router();
+
+  router.post('/games/:id/create', koaBody(), async ctx => {
```
nicolodavis: This call should fail if the game already exists.

src/server/api-server.js
```diff
@@ -0,0 +1,112 @@
...
+  });
+
+  router.patch('/game_instances/:id/join', koaBody(), async ctx => {
+    const gameID = ctx.params.id;
+    const gameName = ctx.request.body.gameName;
+    const playerID = ctx.request.body.playerID;
+    const playerName = ctx.request.body.playerName;
+
+    const namespacedGameID = `${gameName}:${gameID}`;
+    const gameMetadata = await db.get(getGameMetadataKey(namespacedGameID));
```
nicolodavis: Need to handle case where game does not exist.

examples/react/modules/tic-tac-toe/components/authenticated.js
```diff
@@ -0,0 +1,123 @@
...
+import PropTypes from 'prop-types';
+
+const App = Client({
+  game: TicTacToe,
+  board: Board,
+  debug: false,
+  multiplayer: true,
+});
+
+class AuthenticatedClient extends React.Component {
```
nicolodavis: Looks like the game server this is talking to still uses an unauthenticated game, right? I think it would be more useful to manually create an authenticated game using `db.set` in `examples/react/server.js` to demonstrate that this works (i.e. changing the credentials should result in the move not working; at the moment it seems to be a no-op).

src/client/multiplayer/multiplayer.js
nicolodavis: Shouldn't the `this.socket.emit` below take care of this?

src/server/index.js
scally: It's a little weird to me that the game and lobby server would take separate arguments for their games. I get the flexibility argument, but at the same time I don't know if we want to guard for cases like "game server was started with different set of games from lobby server".

examples/react/modules/tic-tac-toe/components/authenticated.js
scally: I'll look into that. It worked for my manual local testing but there was a lot of tribal knowledge in how to use the two together that I could bake into the tests better.

src/server/api-server.js
scally: @nicolodavis I don't think it's possible in this PR to create a game that already exists. 

A game has a name in its definition. The `create` endpoint takes that name, and creates a new instance of the game with that name, with a new game instance ID. 

There could be the unlikely event of a instance ID collision, but we could catch that internally and wouldn't need to fail the call.

src/client/multiplayer/multiplayer.js
scally: @nicolodavis in this instance, when the sync got sent back from the server to the client, the client rejects the sync because the incoming stateID is often seen as earlier/lower than the existing state's stateID, so it's discarded. That appears to be a bug, because we shouldn't compare to an existing stateID if we have switched games, as the new game's state might be totally different, so this comparison no longer makes sense. 

The first thing that came to mind is that we should flush state when we change players or games because in either case the existing state we have shouldn't be used anymore. We could do something different, like change the sync comparison, but I wasn't sure if that was the right move here. What do you think?

scally:
@nicolodavis OK, I've addressed all the things you mentioned in the past few commits. There's a couple of things I commented about inline that I'm still unsure about -- need your feedback whenever you have some time.

src/server/index.js
nicolodavis: But that's actually a feature, isn't it? For example, you could have one lobby server that's the entry point that can handle all games, and the admin creates one game server per game (for better resource management).

nicolodavis:
OK, will take a look soon. Thanks @scally!

scally:
@nicolodavis Saw you added some comments. Is there anything else you want to see changed at this point or are we good to go on this one?

scally:
@nicolodavis ping 😀do you need anything else on this PR? can we merge it soon?

nicolodavis:
Sorry about the delay! I've been a bit busy. Will likely only be able to look at this in a couple of days. I'll make any small changes that are needed and merge it in.

src/client/multiplayer/multiplayer.js
nicolodavis: Yeah, we should change the sync comparison instead. I can make that change.

examples/react/modules/tic-tac-toe/components/authenticated.js
nicolodavis: I'm still not sure what the expected behavior you had in mind here is. Right now the game plays just fine regardless of what's in the credentials text field. I would expect that no moves are allowed until you enter the proper credentials in those fields.

nicolodavis:
This is ready to be merged (all looks good to me), except for the example, which still doesn't work as expected. I'll merge it in once we resolve that.

examples/react/modules/tic-tac-toe/components/authenticated.js
scally: @nicolodavis This is due to the way that we allow unauthenticated games as the default. 

Once you put the gameID emitted when you start up the examples, then that particular game won't accept moves unless you have the correct credentials.

Until you enter a gameID, you're still in the "default" mode which is fully unauthenticated.

We might be able to move the "create authenticated game" code into the authenticated example itself, which might allow us to prepopulate it.

scally: @nicolodavis pushed a change to move authenticated game creation into the authenticated example, which I think will address your concern. Let me know. :)

nicolodavis:
@scally Do we have a flaky test here? https://travis-ci.org/google/boardgame.io/builds/385591396

scally:
Entirely possible. Wondering if it's an ordering-based failure somehow -- some kind of dependent setup?

nicolodavis:
I just restarted the build to see if it passes now (sorry, didn't realize it would lose the log of the previous build). The failure was something along the lines of:

.createApiServer: async callback wasn't called within the timeout

scally:
It looks like it happened again. I suppose it's also possible some other commit has caused a problem here as well, I haven't checked the commit logs yet.

nicolodavis:
It's only been doc file changes after your PR's, so it can't be.

scally:
Will try it a few times locally

scally:
@nicolodavis I've tried this 10x locally, can't duplicate. Maybe it's something in the environment. Do the tests pass for you locally on loop?

nicolodavis:
Yeah, I haven't had trouble running them locally. On the CI, you're running in a more restricted environment, so timeouts are more likely to occur compared to a local run.

I suspect that there is a bug somewhere, but it's not getting triggered on our machines with ample resources.

nicolodavis:
Here is another recent failure: https://travis-ci.org/google/boardgame.io/builds/387813663?utm_source=github_status&utm_medium=notification

scally:
How about cranking up the default timeout interval to see if that fixes the problem on CI? Not ideal, but would get us building again.

nicolodavis:
I added a `done()` at the end of the test. Not sure if it actually helped, but I don't seem to encounter the error anymore, so I guess that's good enough for now.

Issue #198, opened 2018-05-16
Protect lobby API


scally:
Followup for #189 , as it stands today it's possible for anyone to call the lobby API. 

Some ideas:

1. Shared token between lobby webserver & lobby API
2. ???

nicolodavis:
Shared token sounds good to me.

Another option would be to expose **create game** and **join game** as just normal JS functions. A lobby implementation can then just use them directly as wrappers over `db.set` to create games. The ACL's are now provided at the DB level to prevent the outside world from calling them directly.

We can still expose them through a RESTful API like #189 does, but maybe we can also refactor the POST handler logic into simple functions so that they're reusable directly as well?

scally:
I'll start with the token idea, I'm not sure I fully understand the other option atm.

scally:
@nicolodavis a PR is up for this at #204 . It's branched from #189, so it will require a rebase whenever #189 lands, but if you want to see the real diffs for it for now, it's in the last commit.

PR #204, opened 2018-05-28
Allow the lobby API to require a secret for all requests

scally:
If an environment variable `LOBBY_SECRET` is provided when the API server is started, then all API requests will require that secret in a header named `lobby-secret`. (Issue #198)

This check is provided as a middleware for the API koa server.

#### Checklist


scally:
This is branched from #189 , so will require a rebase when it lands.

PR #223, opened 2018-06-17
Firebase integration

bennygenel:
#### Checklist

I created a new class for Firebase Real Time Database and Firestore as I mentioned on https://github.com/google/boardgame.io/issues/6#issuecomment-397820990. It is basically same as MongoDB. 

The problem is I have so little experience _(practically not at all)_ with testing. Because of this I was not able to keep up with the test coverage percentage. Like I did on Firebase class, I basically copy the work done for MongoDB tests for RTDB and Firestore. I really appreciate some help about tests.


nicolodavis:
Thanks @bennygenel! I can definitely help with the tests.

db.js looks a bit crowded now. Do you think it makes sense to now split it into the following?

```
src/server/db/memory.js
src/server/db/mongo.js
src/server/db/firebase.js
```

bennygenel:
Thanks for the help @nicolodavis 

I agree since tests are getting crowded too and separating them too might help.

Quick question;
What do you think on adding support for MySQL, Postgres and SQLite with [sequelize](https://github.com/sequelize/sequelize) since all of them support JSON data types?
I know it might not be ideal to store complete JSONs in fields but since we get/set complete state from database, separating fields might not be that important.

nicolodavis:
Adding support for MySQL / Postgres etc. is a good idea. That should give people more options for hosting.

bennygenel:
How do you suggest on doing that then? Same as Firebase and Mongo perhaps?

Another question @nicolodavis;
Is my last commit ok? I added `Firebase `to `packages/server.js` but I'm not sure if I should have.

src/server/db/firebase.js
```diff
@@ -0,0 +1,147 @@
...
+    const { apiKey, authDomain, databaseURL, projectId } = config;
+    this.config = { apiKey, authDomain, databaseURL, projectId };
+    this.dbname = dbname;
+    this.cache = new LRU({ max: cacheSize });
+  }
+  /**
+   * Connect to the instance.
+   */
+  async connect() {
+    if (this.mockFirebase) {
```
nicolodavis: Instead of polluting the code with testing related stuff, how about doing:

```
db.db = mockFirebase;
```

in the test?

nicolodavis: (I think that should also help bring the test coverage up to 100% by executing lines 41 and 42).

src/server/index.js
```diff
@@ -31,6 +31,20 @@ export function Server({ games, db, _clientInfo, _roomInfo }) {
   if (db === undefined) {
```
nicolodavis: How about moving this code to a function somewhere inside server/db?

```
if (db === undefined) { db = DBFromEnv(); }
```

src/server/db/inmemory.js
```diff
@@ -0,0 +1,55 @@
+/*
```
nicolodavis: Can you branch this file from db.js like you did for mongo.js?

nicolodavis:
@bennygenel About SQL, yeah, let's follow a similar pattern to firebase / mongo.

src/server/db/inmemory.js
bennygenel: I'm not sure I understood what you want here. Can you elaborate a little bit more please?

nicolodavis: When I look at the diff for mongo.js, I see that it was renamed from db.js (as opposed to a delete + add). Looks like this is git automatically detecting a rename. Never mind, I don't think there is a git way to do it for another file.

bennygenel: Oh ok.. Now I understand. 
Yes it was a automatic detection. I deleted files after I created all and I suppose git chose mongo file to be set as renamed. Probably because I created that one first or last.

bennygenel:
Should SQL integration be a different PR or should I continue here?

nicolodavis:
@bennygenel Let's do SQL in a separate PR. Thanks so much for volunteering to do this!

nicolodavis:
I'll take another look at this sometime this weekend. Thanks for the PR again.

bennygenel:
@nicoladavis no problem. I started SQL integration too. I'll finish it up this weekend too.

Aerodos12:
I don't know if I'm doing this wrong or not, but I got this problem:

```FirebaseError: [code=invalid-argument]: Function DocumentReference.set() called with invalid data. Nested arrays are not supported```

I found it in the terminal window that I used to run the server for the multiplayer of my game, and I'm trying to integrate my Firebase project with my game so when I deploy it, everyone's stats and data will be synced to Firebase (I'm also thinking about adding in a system for tracking wins too).

PR #226, opened 2018-06-25
MySQL, SQLite and PostgreSQL integration

bennygenel:
Added support for MySQL, SQLite and PostgreSQL.

Currently databases only support init with connection urls via environment variables since there might be some changes to the initializing of databases. (referring to https://github.com/google/boardgame.io/issues/6#issuecomment-398595260)  

#### Checklist


src/server/db/mysql.js
```diff
@@ -0,0 +1,43 @@
...
+
+import { SQL } from './sql';
+
+const GameModel = {
+  id: {
+    type: Sequelize.STRING,
+    unique: true,
+    primaryKey: true,
+  },
+  ctx: {
```
nicolodavis: This should actually be `state`, not `ctx`.

`state = { G, ctx }`

bennygenel:
@nicolodavis fixed the naming but I really don't know why the test fails. It passes fine on my local machine.

nicolodavis:
Can you fix the following warning?
`sequelize deprecated String based operators are now deprecated. Please use Symbol based operators for better security, read more at http://docs.sequelizejs.com/manual/tutorial/querying.html#operators node_modules/sequelize/lib/sequelize.js:242:13`

Also, about the test timeout, one thing that might help diagnose it would be to split it into smaller `describe / test` chunks similar to the Firebase test [here](https://github.com/google/boardgame.io/blob/master/src/server/db/firebase.test.js#L48).

nicolodavis:
@bennygenel Here is another thought about avoiding pulling in all those dependencies (or giving users the impression that the library is pulling in things they don't need).

How about we just move them to `devDependencies` and then document clearly that if you are using firebase, you need to run:
`npm i --save firebase`

in your project?

Maybe we can add some `console.log` warnings as well if the dependencies aren't present?

bennygenel:
@nicolodavis I'm all up for it. I actually really like sequelize package for this reason. It works just like you said.

[sequelize getting started docs](http://docs.sequelizejs.com/manual/installation/getting-started.html)

> ### Installation
> Sequelize is available via NPM and Yarn.
> 
> ```
> // Using NPM
> $ npm install --save sequelize
> 
> # And one of the following:
> $ npm install --save pg pg-hstore
> $ npm install --save mysql2
> $ npm install --save sqlite3
> $ npm install --save tedious // MSSQL
> 
> // Using Yarn
> $ yarn add sequelize
> 
> # And one of the following:
> $ yarn add pg pg-hstore
> $ yarn add mysql2
> $ yarn add sqlite3
> $ yarn add tedious // MSSQL
> ```

What they do is pretty simple to follow actually;

- Checking the dialect option while initializing [sequelize.js#L214](https://github.com/sequelize/sequelize/blob/9b46f91301f40d7a973becb99d062da81e9f6164/lib/sequelize.js#L214)
```
    let Dialect;
    // Requiring the dialect in a switch-case to keep the
    // require calls static. (Browserify fix)
    switch (this.getDialect()) {
      case 'mssql':
        Dialect = require('./dialects/mssql');
        break;
      case 'mysql':
        Dialect = require('./dialects/mysql');
        break;
      case 'postgres':
        Dialect = require('./dialects/postgres');
        break;
      case 'sqlite':
        Dialect = require('./dialects/sqlite');
        break;
      default:
        throw new Error('The dialect ' + this.getDialect() + ' is not supported. Supported dialects: mssql, mysql, postgres, and sqlite.');
    }

    this.dialect = new Dialect(this);
```
- Checking if the selected dialect's package installed [mysql/connection-manager.js#L29](https://github.com/sequelize/sequelize/blob/9b46f91301f40d7a973becb99d062da81e9f6164/lib/dialects/mysql/connection-manager.js#L29)
```
    try {
      if (sequelize.config.dialectModulePath) {
        this.lib = require(sequelize.config.dialectModulePath);
      } else {
        this.lib = require('mysql2');
      }
    } catch (err) {
      if (err.code === 'MODULE_NOT_FOUND') {
        throw new Error('Please install mysql2 package manually');
      }
      throw err;
    }
```

I think we can follow a similar approach. We should have a single class (`const db = new Database(options)` for example) that initializes the db with a dialect or similar option. This way I think we can unify the documentation for database. Also we can give user some sort of a system to add his/her technique to add `get` `set` `has` methods.

**Sample**
```
const db = new Database({
  dialect: 'custom',
  get: async (id) => {
    return await customdboperation.get(id);

  },
  set: async (id, state) => {
    return await customdboperation.set(id, state)
  },
  has: async (id) => {
    return await customdboperation.get(id) !== null;
  },
  // some other options
})
```
Do you think we can implement something similar?

bennygenel:
@nicolodavis I know that but my intention was making things a little bit more clear for the user.

Current documentation for Server (it was not accurate while I am writing this) directs user to an already defined database class to match his/her implementation.

> ```
> const server = Server({
>   games: [game1, game2, ...],
> 
>   // Optional, if you want to hook it up to a
>   // custom storage backend not supported by
>   // the framework. DbImpl must implement the
>   // same interface shown in db.js:
>   // https://github.com/google/boardgame.io/blob/master/src/server/db.js
>   db: new DbImpl(),
> });
> ```

I believe this way is open for potential errors since we are not checking anywhere that if the user's implementation handles the required methods. 

```
nsp.on('connection', socket => {
      socket.on('action', async (action, stateID, gameID, playerID) => {
        // Not checking if set defined or not
        let state = await db.get(gameID);
        // ...
     })
     // ...
})
```

With that in mind having a single class that has a clear documentation on required or optional arguments might be a better way. At least in my opinion.

nicolodavis:
@bennygenel I agree. So you're proposing that we expose a base class that users will inherit from to implement their custom DB adaptors?

BTW, I haven't forgotten about this PR. Just been busy the past week. We still need to remove the direct dependencies and document what to import if using Firebase / Mongo / SQL correct?

nicolodavis:
@bennygenel I moved the dependencies and added some documentation. Let me know what you think.

While writing the documentation, I realized that we could probably strive for a bit more consistency. For example:

1. Do you think it makes sense to flatten out `config` so that users can do:

```
new SQLite({
  host:
  port:
  ...
})
```

instead of 

```
new SQLite({
  config: { ... }
})
```

I realize that you probably wanted to put it inside config to distinguish it from `url` (which provides a shorthand instead), but I feel like having nested options can start to get confusing unless absolutely warranted.

2. Speaking of `url`, it looks like SQLite uses `path`. Shall we settle on one? Also, `database` vs `dbname`.

bennygenel:
@nicolodavis sorry for not being involved for a while, I was a little busy with something else and thank you for the improvements.

> Do you think it makes sense to flatten out config so that users can do

I thought about flattening the config parameter but I didn't feel confident about it since every db class has its own specific parameters -like firebase has `apikey` param and mysql has `password` param- so I ended up using same config param for each. This also doesn't make much sense since Firebase doesn't have `url `parameter. In the end I left them as is because we were talking about a Database class implementation. 

> Speaking of url, it looks like SQLite uses path. Shall we settle on one? Also, database vs dbname

That is completely me messing up. I'll change `database` to `dbname` where possible.

I was planning to talk with you about sqlite path param. That is why I left it like that. Does it make sense to make it possible to use it like `sqlite://{{path}}` and also `{{path}}` and check if `url ` parameter contains `sqlite://` and if not add it. (Hope I was able to explain it)

> So you're proposing that we expose a base class that users will inherit from to implement their custom DB adaptors

My first idea was to have a single exported Database class that can be used by the user. This made much sense in my mind. This way we can do all the checks we need and set the databases while user can configure it much more easily. This would end with a simpler documentation. But I couldn't wrap my head around it. I came up with below snippet but that didn't seem very nice.

```javascript
class Database {
  constructor({ engine, get, /* other params */ }) {
    switch(engine) {
      case 'mysql':
        this.engine = new MySQL({ /* params */ });
        break;
      case 'custom':
        this.engine = 'custom';
        if(!get) throw Error('get must be defined.');
        else this.customGet = get;
        break;
      default:
        // no engine, use memory implemetation
        break;
    }
  }

  async get(id) {
    if(this.engine === 'custom' && this.customGet) return await this.customGet(id);
    return await this.engine.get(id);
  }

  // similar implementation for set has and connect
}
```
I thought about creating a base class that user can inherit from and came up with [this gist](https://gist.github.com/bennygenel/debf89c668f4d259737ca78a38761edd) This doesn't seem that pretty either since there is multiple levels of `extend` usage.

Sorry for the long post and hope you can guide me on a better way.

nicolodavis:
I quite like your Database class gist actually!  It does the checking of required modules and the user just has to extend the class, which is a good pattern for this use-case I think.

About sqlite://, I'm in favor of having the code automatically figure out if it needs to prepend that or do nothing if it is already in the path.

nicolodavis:
We can probably figure out the Database class abstraction later on too.

For this specific PR, do you know why `sql.test.js` is failing?

nicolodavis:
@bennygenel Just checking if you're still working on this. No worries if you are busy.

nicolodavis:
Closing this now that we have a Postgres connector: https://github.com/janKir/bgio-postgres

PR #229, opened 2018-07-01
expose Debug element

amitport:
#### Checklist

Exporting Debug element so it can be re-used

### DIFF
--- a/packages/react.js
+++ b/packages/react.js
@@ -7,5 +7,6 @@
  */
 
 import { Client } from '../src/client/react.js';
+import { Debug } from '../src/client/debug/debug.js';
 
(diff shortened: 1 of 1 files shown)


packages/client.js
```diff
@@ -7,5 +7,6 @@
  */
 
 import { Client } from '../src/client/client.js';
+import { Debug } from '../src/client/debug/debug.js';
 
-export { Client };
+export { Client, Debug };
```
nicolodavis: client.js does not have any React dependencies. This should probably go in packages/react.js.

amitport: updated - moved to react.js

PR #230, opened 2018-07-03
RESTful games/create

amitport:
#### Checklist

### DIFF
--- a/docs/api/Server.md
+++ b/docs/api/Server.md
@@ -47,7 +47,7 @@ server.run(8000);
 
 You can optionally choose to require clients to use credential tokens to prove they have the right to send actions on behalf of a player.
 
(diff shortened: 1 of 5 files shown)


docs/api/Server.md
```diff
@@ -57,7 +57,7 @@ Authentication APIs are available by default on `WebSocket port` + 1.
 
 ### Creating a game
 
-#### `/games/:name/create`
```
nicolodavis: What's the benefit of dropping `create` here?

amitport: That's how [Sir Timothy John Berners-Lee](https://en.wikipedia.org/wiki/Tim_Berners-Lee) wanted it.
URLs repesent resources (games collection in this case) and POST already means 'create'. See also: https://restfulapi.net/http-methods/#post,

(Not very critical in this case though)

PR #231, opened 2018-07-03
API: check secret _before_ handling the request

amitport:
#### Checklist

### DIFF
--- a/src/server/api.js
+++ b/src/server/api.js
@@ -127,15 +127,15 @@ export const createApiServer = ({ db, games }) => {
 
   // If API_SECRET is set, then require that requests set an
   // api-secret header that is set to the same value.
-  app.use((ctx, next) => {
+  app.use(async (ctx, next) => {
     if (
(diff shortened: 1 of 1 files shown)


src/server/api.js
```diff
@@ -127,15 +127,15 @@ export const createApiServer = ({ db, games }) => {
...
-
+  app.use((ctx, next) => {
     if (
       !!process.env.API_SECRET &&
       ctx.request.headers['api-secret'] !== process.env.API_SECRET
     ) {
       ctx.throw(403, 'Invalid API secret');
     }
+    
+    return next();
```
nicolodavis: Let's keep the `async / await` here:

```
app.use(async (ctx, next) => {
      if (
        !!process.env.API_SECRET &&
        ctx.request.headers['api-secret'] !== process.env.API_SECRET
      ) {
        ctx.throw(403, 'Invalid API secret');
      }

      await next();
});
```

PR #234, opened 2018-07-10
expect protocol to be included in 'server' param

amitport:
(otherwise browser blocks for mixed content)

#### Checklist

### DIFF
--- a/src/client/multiplayer/multiplayer.js
+++ b/src/client/multiplayer/multiplayer.js
@@ -92,7 +92,7 @@ export class Multiplayer {
     if (!this.socket) {
       if (this.server) {
         this.socket = io(
-          'http://' + this.server + '/' + this.gameName,
+          '//' + this.server + '/' + this.gameName,
           this.socketOpts
         );
       } else {


nicolodavis:
Good catch!

Looks like some tests are failing. Could you take a look?

amitport:
@nicolodavis tests window.location.protocol is set to `about:` which socket.io can't make sense of.

change the PR to require full origin in the server param. This will also enable setting https server in an http website.

src/client/multiplayer/multiplayer.js
```diff
@@ -91,10 +91,7 @@ export class Multiplayer {
   connect() {
     if (!this.socket) {
       if (this.server) {
-        this.socket = io(
-          'http://' + this.server + '/' + this.gameName,
-          this.socketOpts
-        );
+        this.socket = io(this.server + '/' + this.gameName, this.socketOpts);
```
nicolodavis: You can also connect using '//hostname:port' and socket.io will choose the appropriate transport in order to avoid a mixed connection.

How about we leave the server specification as `<hostname>:<port>` and use this to determine when to switch to https. In order to support the use case of forcing SSL (even on http connections), socket.io has a `{ secure: true }` option that you can pass here that we could probably also expose in the multiplayer spec like this:

```
multiplayer: { server: 'hostname:port', secure: true },
```

amitport: fine, not sure what to do about the tests though... 'socket.io' still seems to default to `about` protocol

amitport:
@nicolodavis any comments? made the change though IMO passing the server url as is is preferable (boardgame.io shouldn't add its own prefix '//' or other)

nicolodavis:
I just want the ability to also just specify hostname:port and have it do something reasonable. How about we:

1. Prepend // when the input is just hostname:port.
2. Do nothing if the input already contains the protocol.

nicolodavis:
Let's close this in favor of #247.

PR #235, opened 2018-07-12
Update tutorial.md (render draw gameover)

EmrysMyrddin:
#### Checklist

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -173,7 +173,7 @@ class TicTacToeBoard extends React.Component {
     let winner = '';
     if (this.props.ctx.gameover) {
       winner =
-        this.props.ctx.gameover.winner !== undefined ? (
+        this.props.ctx.gameover.draw  ? (
           <div id="winner">Winner: {this.props.ctx.gameover.winner}</div>
         ) : (
           <div id="winner">Draw!</div>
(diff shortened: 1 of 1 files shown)


docs/tutorial.md
```diff
@@ -173,7 +173,7 @@ class TicTacToeBoard extends React.Component {
     let winner = '';
     if (this.props.ctx.gameover) {
       winner =
-        this.props.ctx.gameover.winner !== undefined ? (
+        this.props.ctx.gameover.draw  ? (
```
nicolodavis: Isn't this going to say "Draw!" when `this.props.ctx.gameover.draw` is `false`?

EmrysMyrddin: Yes you're totally right. I have fixed it.

EmrysMyrddin:
No it's not really a bug.
It's just that the tutorial is showing how to return an object in the function `endGameIf`.

If the game is draw, it set a flag, but this flag is then not used. I found this a little confusing.

nicolodavis:
In the meantime, I think it's ok to leave the code as is. The two states can either be Winner or Draw, so it's ok to just check:

```
if (winner) { display Winner }
else { display Draw }
```

rather than checking for the `draw` field explicitly.

nicolodavis:
Feel free to continue the discussion here of course if you think we should change something.

PR #240, opened 2018-07-22
Use parse/stringify from flatted lib to support circular structures (fixes #222)

### LINKED ISSUE #222: AI debugging info isn't serialized properly on save
Hi;
I noticed an error while saving on tic-tac-toe single player example. Error is below. It occurs when you want to `save` after you use `step` and it doesn't happen if you manually play.

```
Uncaught TypeError: Converting circular structure to JSON
    at JSON.stringify (<anonymous>)
    at Object.Debug._this6.saveState [as onPress] (bundle.js:63017)
    at KeyboardShortcut._this4.activate (bundle.js:62926)
    at bundle.js:62939
    at _fireCallback (bundle.js:64086)
    at Mousetrap.self._handleKey (bundle.js:64147)
    at Mousetrap.handleKey (bundle.js:64464)
    at HTMLDocument._handleKeyEvent (bundle.js:64207)
```

Stefan-Hanke:
This PR fixes the problem described in #222. It replaces the use of JSON.stringify/JSON.parse with functions from the flatted library.

Instead of just fixing that problem, this PR changes stringify/parse to flatted in all parts that consitute the library itself - i.e. the examples and the storybook are unchanged.

#### Checklist

### DIFF
--- a/src/client/debug/debug.js
+++ b/src/client/debug/debug.js
@@ -16,6 +16,7 @@ import { PlayerInfo } from './playerinfo';
 import { DebugMove } from './debug-move';
 import { GameLog } from '../log/log';
 import { restore } from '../../core/action-creators';
+import { parse, stringify } from 'flatted/cjs';
 import './debug.css';
 
 /**
@@ -102,14 +103,14 @@ export class Debug extends React.Component {
   }
 
   saveState = () => {
-    const json = JSON.stringify(this.props.gamestate);
+    const json = stringify(this.props.gamestate);
     window.localStorage.setItem('gamestate', json);
   };
 
   restoreState = () => {
     const gamestateJSON = window.localStorage.getItem('gamestate');
     if (gamestateJSON !== null) {
-      const gamestate = JSON.parse(gamestateJSON);
+      const gamestate = parse(gamestateJSON);
       this.props.store.dispatch(restore(gamestate));
     }
   };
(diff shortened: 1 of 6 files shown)


src/client/debug/debug.js
```diff
@@ -16,6 +16,7 @@ import { PlayerInfo } from './playerinfo';
 import { DebugMove } from './debug-move';
 import { GameLog } from '../log/log';
 import { restore } from '../../core/action-creators';
+import { parse, stringify } from 'flatted/cjs';
```
bennygenel: I think it should be `flatted/esm`

From https://github.com/WebReflection/flatted#flatted
```javascript
// ESM
import {parse, stringify} from 'flatted/esm';

// CJS
const {parse, stringify} = require('flatted/cjs');

```

Stefan-Hanke: I aggree. I changed to esm locally, and changed the rollup config accordingly. However, now tests do not run anymore because they fail to resolve the "export" keyword inside flatted/esm/index.js.

Somehow jest now needs to be told to transform ES6 modules also. Likely another change in .babelrc is necessary.

Stefan-Hanke: So how was I able to push this branch at all - it should have been caught by the prepush hook? Since I can't push the esm variant because now tests are failing :(

As far as I understand this, it should work since .babelrc contains `transform-es2015-modules-commonjs` inside test env directly.

nicolodavis: Sometimes tests can pass locally but fail on Travis if it depends on something in node_modules that's set up correctly locally manually.

nicolodavis: Oh I thought your question was about how to push now that the tests are failing.

Stefan-Hanke:
Well and now one cannot run the dev server anymore because it complains that "export" keyword is unexpected inside flatted/esm.

nicolodavis:
Will take a look.

nicolodavis:
Fixed by just importing:
`flatted` instead of `flatted/esm` or `flatted/cjs`.

Their [package.json](https://github.com/WebReflection/flatted/blob/master/package.json#L7) includes a `main` and `module` section, which should allow imports to pick up the correct file.

About the coverage check, are you talking about why we use Coveralls in addition to Travis? Travis merely runs the build and reports the coverage to Coveralls, which has better [tooling](https://coveralls.io/builds/18150843) like the ability to view covered / uncovered lines. In theory it is possible to have Travis just spit out a coverage number and fail the build if it doesn't meet a threshold, but Coveralls is useful enough that I decided to keep them separate.

Issue #227, opened 2018-06-29
Game Log Improvements


nicolodavis:
### Secret Log Events
We have secret state, but log events all go to a global log. We need some mechanism to allow sending a customized log per player so that secret state isn't leaked.

**Status**: *not started*

### Custom logging
Allow players to insert custom messages into each log event.

**Status**: *not started*

### Incremental log updates
Sending the entire log on each `sync` is going to eventually become a performance problem for large games. Maybe we should just send the log events since the last event that the client has?

**Status**: *not started*

Stefan-Hanke:
This has been a while. Correct me if I'm wrong. This is just a list of thoughts that I've fished while thinking about the game log.

* The log currently is just an array that is added to at two locations in reducer.js and flow.js. It contains the raw actions that are sent to the reducer. The action itself already contains the `playerID` from the player performing the action.

* I don't know what constitutes `credentials` but looks like this should be stripped for the log in any case (even for the player who'd performed it).

* Actually I don't fully understand what you mean by leaking secret state because of an action. Well, maybe that action's arguments contain some secret for that specific player, which of course should not be visible to other players.

* I'm kind of inclined to using `PlayerView`s for that. You?

* The third point looks like not belonging to a V1.0 release. I don't know whether a file can be linked to an issue; do you have specific reasons to not use GitHubs Milestone feature?

* The fourth point looks like being UI-specific? React? Native? Whatever?

nicolodavis:
@Stefan-Hanke 

- Yes, we should strip credentials from the log.
- About leaking secret state, I meant an action could be something "move card A from my hand to my deck", which shouldn't be visible to other players. Yeah, we could use PlayerView for that too, although we'll have to repurpose it to filter out actions as well as state (as it does currently).
- The third point may not be a release blocker, but we need to do some stress tests before we are sure of that. Not opposed to using GitHub milestones. I will probably start marking issues that specifically target v1.0 as such.
- Yep, the fourth point only affects the bundled in rendering of the log.

Stefan-Hanke:
On Point 4 (Handle endPhase in log). Currently, it is just plain rendered. Did you mean to render something special like it's done for a turn?

nicolodavis:
Yes, something that indicates what the current phase is would be helpful in the logistics think. It could potentially be a color coded wrapper around each log event (withsmall text on the top right every time the phase changes).

Stefan-Hanke:
[...]

PR #243, opened 2018-07-25
Log visualize phases

Stefan-Hanke:
This PR implements a part of  #227 - namely, visualize phases and changes thereof.
I'm no designer, but with this, the current phase and changes thereof are clearly visible.

image

#### Checklist

#### Note

Nicolo will be away for 2-3 weeks starting July 16th 2018, so expect some delays in getting a response.

### DIFF
--- a/src/client/log/log.css
+++ b/src/client/log/log.css
@@ -14,6 +14,8 @@
   background: #fff;
   border: 1px dotted #ccc;
   border-left: 5px solid #ccc;
+  border-top: 2px solid #ccc;
+  border-bottom: 2px solid #ccc;
   margin-bottom: 5px;
   padding: 5px;
   text-align: center;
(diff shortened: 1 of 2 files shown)


Stefan-Hanke:
oh my - I relied too much on the prepush hook to catch errors. First prop type validation, and now uncovered lines. :-1:

nicolodavis:
Hmm, looks like the colors get a bit jarring for longer logs.

I have an alternate proposal in https://github.com/google/boardgame.io/tree/log-phases that looks like the following (turns on the left and phases on the right; it also has the nice property of conveying to the user the fact that turns and phases are orthogonal concepts):
screenshot from 2018-07-26 23-41-07

What do you think?

Stefan-Hanke:
Looks ok to me. I think I'll not spend time with UI stuff anymore...

PR #246, opened 2018-08-01
allow multiplayer client creation to be overridden

amitport:
#### Checklist

(I'm starting to experiment with an alternative multiplayer client, would be great if I could keep connection with source repo instead of forking it)

1- This allows inheritance by exposing Client class directly instead of wrapping a _ClientImpl.
2- 'new Multiplayer' is replace by a call to overridable function 'createMultiplayerClient'

### DIFF
--- a/src/client/client.js
+++ b/src/client/client.js
@@ -91,8 +91,14 @@ class _ClientImpl {
     this.credentials = credentials;
 
     let server = undefined;
-    if (multiplayer instanceof Object && 'server' in multiplayer) {
-      server = multiplayer.server;
+    let MultiplayerClientImpl = Multiplayer;
+    if (multiplayer instanceof Object) {
+      if ('server' in multiplayer) {
+        server = multiplayer.server;
+      }
+      if ('clientImpl' in multiplayer) {
+        MultiplayerClientImpl = multiplayer.clientImpl;
+      }
       multiplayer = true;
     }
 
@@ -134,7 +140,7 @@ class _ClientImpl {
     this.store = null;
 
     if (multiplayer) {
-      this.multiplayerClient = new Multiplayer({
+      this.multiplayerClient = new MultiplayerClientImpl({
         gameID: gameID,
         playerID: playerID,
         gameName: game.name,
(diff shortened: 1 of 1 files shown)


nicolodavis:
Can you describe a bit more what the alternative multiplayer client would do compared to the existing one?

amitport:
1 use libraries other communication libraries then socket.io (which may have more robust connection support- including something like firebase-cloud-messaging)
2 local multiplayer - client side server (for testing / debugging / developing)

as I see it boardgame.io is conceptually 3 packages (at least):
- 1 core logic where the biggest value is providing a (hopefully, eventually) poplar standard
- 2 ui / react client
- 3 multiplayer game session manager 

I'm interested only in the first component (core logic), there is a good separation with the UI, but the multiplayer component is currently intertwined (and a good first step is just to allow for overriding it, maybe later there could be a clear interface for users to select between several multiplayer backends)

nicolodavis:
Sounds good. I've been meaning to do this myself actually, so I'm glad you're taking the first step!

I'd like to minimize any changes to the existing public API as much as possible. Will take a closer look later this week and send you more actionable feedback.

src/client/client.js
```diff
@@ -134,7 +150,7 @@ class _ClientImpl {
     this.store = null;
 
     if (multiplayer) {
```
nicolodavis: Let's just add a new option in the `multiplayer` argument that allows specifying a custom client. Something like:

```
Client({
  game: TicTacToe,
  multiplayer: {
    server: 'hostname:port',
    clientImpl: <object>,
  }
})
```

That way it's a tiny surgical change and the rest of the interface remains unchanged.

amitport: update as requested 

Note that this interface does not reflect that serverUrl and socketOpt are parameters for a specific Multiplayer client type

src/client/client.js
```diff
@@ -91,8 +91,14 @@ class _ClientImpl {
...
     let server = undefined;
-    if (multiplayer instanceof Object && 'server' in multiplayer) {
-      server = multiplayer.server;
+    let MultiplayerClientImpl = Multiplayer;
+    if (multiplayer instanceof Object) {
+      if ('server' in multiplayer) {
+        server = multiplayer.server;
+      }
+      if ('clientImpl' in multiplayer) {
+        MultiplayerClientImpl = multiplayer.clientImpl;
```
nicolodavis: Looks like we're missing a unit test that covers this line.

amitport:
@nicolodavis closing this in favor of external transport refactor (still need GameMaster to be exported)

nicolodavis:
Yep, sounds good. Will work on that.

nicolodavis:
Relevant commits (that will be included in the next release):
https://github.com/google/boardgame.io/commit/d7a1213f1ca7251aed02a280497fe5d3b42870e6
https://github.com/google/boardgame.io/commit/137dd7ca8b2656226a7e4d27bc666099ebcfdb45

PR #247, opened 2018-08-03
Added ability to specify server protocol

iamchristopher:
#### Checklist

### DIFF
--- a/src/client/multiplayer/multiplayer.js
+++ b/src/client/multiplayer/multiplayer.js
@@ -91,10 +91,12 @@ export class Multiplayer {
   connect() {
     if (!this.socket) {
       if (this.server) {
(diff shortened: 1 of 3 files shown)


src/client/multiplayer/multiplayer.js
```diff
@@ -91,10 +91,12 @@ export class Multiplayer {
   connect() {
     if (!this.socket) {
       if (this.server) {
-        this.socket = io(
-          'http://' + this.server + '/' + this.gameName,
-          this.socketOpts
-        );
+        var server = this.server;
```
nicolodavis: Change var to let.

iamchristopher: Good catch! Not sure why I did that..

PR #252, opened 2018-08-09
Compactify transferdata (deltalog, _initial)

Stefan-Hanke:
This PR implements part 3 of #227 (deltalog).

The server's reaction to an `action` command is changed and only added log entries are transferred (the `deltalog`). At client side, the deltalog is added to the current state`s log again.

`_initial` is stripped from `action` commands since it's only necessary to transfer it once during the initial `sync` to the server. It also gets added at client side.

#### Checklist

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -94,6 +94,8 @@ export function Server({ games, db, _clientInfo, _roomInfo }) {
             const newState = Object.assign({}, state, {
               G: game.playerView(state.G, ctx, playerID),
               ctx: ctx,
+              // _initial is sent during "sync" already, no need to send it again
+              _initial: undefined,
             });
 
             if (client === socket.id) {
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -111,7 +111,7 @@ export function Server({ games, db, _clientInfo, _roomInfo }) {
           }
 
           // reconstruct the whole log from deltalogs
-          const log = [...(state.deltalog || []), ...(newState.deltalog || [])];
+          const log = [...(state.deltalog || []), ...newState.deltalog];
           const storeState = { ...newState, log };
 
           await db.set(gameID, storeState);
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -92,13 +92,13 @@ export function Server({ games, db, _clientInfo, _roomInfo }) {
           const roomClients = roomInfo.get(gameID);
           for (const client of roomClients.values()) {
             const { playerID } = clientInfo.get(client);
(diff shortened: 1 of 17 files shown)


src/server/index.js
```diff
@@ -84,26 +84,41 @@ export function Server({ games, db, _clientInfo, _roomInfo }) {
         if (state._stateID == stateID) {
           // Update server's version of the store.
           store.dispatch(action);
-          state = store.getState();
+          const newState = store.getState();
+
+          // get new log entries that were attached during execution of the action.
+          // slice called with negative index extracts last N elements
+          const deltalog = newState.log.slice(
```
nicolodavis: This can probably be done in a later PR (or in this one if you have time and want to make this change):

I think we should have the game reducer generate delta logs. The log object is only appended to anyway (it is never read), so the reducer might as well only generate the log events that were generated after a particular action was applied to the state (rather than returning the entire log). That way a potentially large log object doesn't have to be passed around, and we could in theory even separate the log object out of the state object.

Like I said earlier, feel free to leave this for me / you / someone else to follow up if you'd prefer to not make this PR larger.

nicolodavis: This will become particularly important when we decide to implement part 4 in #227 (attaching the entire state to each log event).

Stefan-Hanke: I'll give it a shot.

Stefan-Hanke:
Oops did not check on coverage - will fix.

When a reducer is used in tests, it gets quite ugly since the log has to be constructed piecewise.

nicolodavis:
Made the changes. Let me know what you think.

The gist of it is:
- `client.js` is responsible for maintaining the log object on the client side. Tests that require the log object use this code instead of the reducer directly (so avoids having the tests assemble the log piecewise).

- The server also has some code to construct the log from a deltalog. I want it to eventually result in a storage level change (store the log in a separate table / area of storage), but for now it is kept with the state object. Once we have the storage level optimization, the server will no longer have to read the log when it processes an update (which I think will result in a performance improvement).

- removed the initialState optimization from this PR (I still think it's a very good thing to do, but let's add it in a future change).

- the socket events are separated clearly by function:
    `UPDATE` - sends back deltalogs.
    `SYNC` - sends back the whole log.

Stefan-Hanke:
Using a reducer middleware is the way to go. While implementing I felt this was not the end of the story, it just was too unwieldy. This PR now contains too much of a change for my taste, but it's OK.

When the framework stores state per log entry, we'd need an initial log entry that has the initial state attached to it.

nicolodavis:
OK, if you're on board with this, I'll merge this in and we can continue work on #227.

PR #261, opened 2018-08-20
Prep for GameLog stuff: Introduce ContextEnhancer

Stefan-Hanke:
This PR is written as prep work for #227, where we want to attach a `log` object to the context, allowing a game to write custom logging information. It introduces a class called `ContextEnhancer` that ties all APIs together that can be attached and detached from a context - currently, these are `random` and `events`. This results in
* less code duplication
* easier attaching of new APIs.

Albeit I failed to also fixup the code creating the initial state, I've not exactly pinpointed why using `ctxApi.update()` does not work. The flow of the various state and context objects gets unwieldy, too. And perhaps there are other places that could use that class also.

Let me know what you think.

#### Checklist

### DIFF
--- a/src/core/random.js
+++ b/src/core/random.js
@@ -32,8 +32,10 @@ export class Random {
    * Updates ctx with the PRNG state.
    * @param {object} ctx - The ctx object to update.
    */
-  update(ctx) {
-    return { ...ctx, _random: this.state };
+  update(state) {
+    var newCtx = { ...state.ctx, _random: this.state };
+    var newState = { ...state, ctx: newCtx };
+    return newState;
   }
 
   /**
--- a/src/core/reducer.js
+++ b/src/core/reducer.js
@@ -67,7 +67,7 @@ export function CreateGameReducer({ game, numPlayers, multiplayer }) {
   initial.G = state.G;
   initial._undo = state._undo;
   initial.ctx = ctxWithEvents;
-  initial.ctx = random.update(initial.ctx);
+  initial.ctx = random.update(initial).ctx;
   initial.ctx = Random.detach(initial.ctx);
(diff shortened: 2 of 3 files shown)


src/core/random.js
```diff
@@ -32,8 +32,10 @@ export class Random {
    * Updates ctx with the PRNG state.
    * @param {object} ctx - The ctx object to update.
    */
-  update(ctx) {
-    return { ...ctx, _random: this.state };
+  update(state) {
+    var newCtx = { ...state.ctx, _random: this.state };
```
nicolodavis: This can be written more concisely as:

```js
const ctx = { ...state.ctx, _random: this.state };
return { ...state, ctx };
```

src/core/reducer.js
```diff
@@ -58,18 +88,17 @@ export function CreateGameReducer({ game, numPlayers, multiplayer }) {
...
+
+  // TODO liked to use apiCtx.update() here also, albeit
+  // this code is more intricate than it looks.
+  const { ctx: ctxWithEvents } = apiCtx.events.update(state);
   initial.ctx = ctxWithEvents;
-  initial.ctx = random.update(initial.ctx);
-  initial.ctx = Random.detach(initial.ctx);
-  initial.ctx = Events.detach(initial.ctx);
+  initial.ctx = apiCtx.random.update(initial).ctx;
+  initial.ctx = apiCtx.detachFromContext(initial.ctx);
```
nicolodavis:
```js
  state = apiCtx.update(state);
  initial.ctx = state.ctx;
  initial.ctx = apiCtx.detachFromContext(initial.ctx);
```

should work here.

A more interesting observation is given that `update` is always followed by a `detachFromContext`, shall we combine them into one call `detach` that accepts `state` and returns `state`?

PR #262, opened 2018-08-21
Add tests documentation

### LINKED ISSUE #260: Game Logic Test Example
It would be great to have an example on how to test the game logic itself without the UI layer and without using low level (redux) methods such as creating a reducer and dispatching events.

yairhaimo:
This PR adds documentation for tests. Covers unit tests, game logic scenario tests, integration tests with React.
Closes #260


yairhaimo:
I signed it!

flamecoals:
+1 for merging. This document is really useful! I was banging my head around how to do the "scenario tests".

Stefan-Hanke:
Agreed, this _is_ useful information. What about having an example where these types of tests are actually used, e.g. for Tic-Tac-Toe?  Doesn't need to be in this PR though, maybe a follow-up.

nicolodavis:
@Stefan-Hanke People will read this section after having gone through the tutorial, which is Tic-Tac-Toe.  I think there is a benefit in keeping the same code example that will be fresh in their memory.

Is there a downside to using Tic-Tac-Toe here?

nicolodavis:
Thanks for the PR @yairhaimo! Merged. This will be helpful to many.

PR #265, opened 2018-08-25
WIP: Custom log messages

Stefan-Hanke:
This PR is intended to implement #227 part 2. Currently, the action inside a log event is wrapped in an object.

Next steps:
* [ ] provide possibility for a move/event to add custom messages
* [ ] show these messages inside the UI

#### Checklist

### DIFF
--- a/src/client/log/log.js
+++ b/src/client/log/log.js
@@ -112,7 +112,7 @@ export class GameLog extends React.Component {
   rewind = logIndex => {
     let state = this.props.initialState;
     for (let i = 0; i <= logIndex; i++) {
-      const action = this.props.log[i].action;
+      const { action } = this.props.log[i];
       if (!action.automatic) {
         state = this.props.reducer(state, action);
       }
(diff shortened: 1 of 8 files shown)


src/client/log/log.js
```diff
@@ -112,7 +112,7 @@ export class GameLog extends React.Component {
   rewind = logIndex => {
     let state = this.props.initialState;
     for (let i = 0; i <= logIndex; i++) {
-      const action = this.props.log[i];
+      const action = this.props.log[i].action;
```
nicolodavis: const { action } = this.props.log[i];

src/client/log/log.js
```diff
@@ -158,14 +158,14 @@ export class GameLog extends React.Component {
 
     let lastAction = 0;
     for (let i = 0; i < this.props.log.length; i++) {
-      const action = this.props.log[i];
+      const action = this.props.log[i].action;
```
nicolodavis: same

src/client/log/log.js
```diff
@@ -158,14 +158,14 @@ export class GameLog extends React.Component {
...
       if (action.type == MAKE_MOVE || !action.automatic) {
         lastAction = i;
       }
     }
 
     for (let i = 0; i < this.props.log.length; i++) {
-      const action = this.props.log[i];
+      const action = this.props.log[i].action;
```
nicolodavis: and here

PR #267, opened 2018-08-27
Allow to set payload onto a log event

Stefan-Hanke:
This PR is part of #227. It allows to actually set a log payload from a move and have it visualized. I've modified the `clickCell` method on the Tic-Tac-Toe example to set a payload.

@nicolodavis Please have a look at the visualization - I played around with the CSS (`overflow` and the fixed height) to no avail; perhaps you have a suggestion on how to visualize a payload.

grafik

#### Checklist

### DIFF
--- a/src/client/log/log.js
+++ b/src/client/log/log.js
@@ -28,7 +28,7 @@ const LogEvent = props => {
 
   const custompayload =
     props.payload !== undefined
-      ? 'payload: ' + JSON.stringify(props.payload, null, 4)
+      ? ', payload: ' + JSON.stringify(props.payload, null, 4)
       : '';
 
   return (
--- a/src/client/log/log.js
+++ b/src/client/log/log.js
@@ -26,6 +26,11 @@ const LogEvent = props => {
     classNames += ' pinned';
   }
 
+  const custompayload =
+    props.payload !== undefined
+      ? 'payload: ' + JSON.stringify(props.payload, null, 4)
+      : '';
+
   return (
     <div
       className={classNames}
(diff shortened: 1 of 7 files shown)


src/core/reducer.js
```diff
@@ -11,38 +11,83 @@ import * as Actions from './action-types';
...
     ctxWithAPI = this.events.attach(ctxWithAPI);
+    ctxWithAPI = this.log.attach(ctxWithAPI);
     return ctxWithAPI;
   }
 
-  detachFromContext(ctx) {
+  static detachAllFromContext(ctx) {
     let ctxWithoutAPI = Random.detach(ctx);
     ctxWithoutAPI = Events.detach(ctxWithoutAPI);
+    ctxWithoutAPI = GameLoggerCtxApi.detach(ctxWithoutAPI);
```
nicolodavis: Call it "API" rather than "Api" to be consistent with the rest of the codebase.

src/client/log/log.js
```diff
@@ -34,12 +39,14 @@ const LogEvent = props => {
       onMouseLeave={() => props.onMouseLeave()}
     >
       {action.payload.type}({args.join(',')})
```
nicolodavis: Wrap this line and the next in a separate <div>:

```
<div>{ type and args }</div>
<div>{custompayload}</div>
```

That should make it better already. I'll take a look later when I'm at my computer and tweak the CSS.

Stefan-Hanke: This does not work, it makes the payload not showing up at all. It gets shifted down below the next log event.

nicolodavis: Ah I see. OK, will take a closer look later and make some changes to your PR.

Stefan-Hanke: ... as always ;) thanks.

nicolodavis:
Changed to the following:

screenshot from 2018-09-02 09-31-23

I would imagine that the user might want to actually want to control how to render the custom payload (rather than just JSON.stringify). What do you think about allowing the GameLog to also accept a React component as a prop that's used to render the payload? (we will use JSON.stringify as a fallback of course).

PR #268, opened 2018-08-29
Redact Log Events

Stefan-Hanke:
_Do not merge, at least for now. This is meant to facilitate discussion._

This PR is part of #227 and serves to illustrate a possible way to strip secrets from log events (part 1).

I've just copied the structure of `playerView` - a function called `logView` can be implemented by games that is allowed to filter log events. Analogous, we need to have some structure attached to the arguments to be able to generically strip secrets as done by `PlayerView.STRIP_SECRETS`.

#### Checklist


nicolodavis:
How about a simpler declarative approach, analogous to `allowedMoves`.  We just ask the developer to provide a list of moves that need to be redacted from the log.  We can even provide some customization like the ability to merely redact the arguments.  We can still provide raw access to the log for even more control, but I'd rather only expose that when we encounter a concrete use-case for it.

```js
redactedMoves: ['clickCell', ... ]
```

```js
redactedMoves: {
  'clickCell': { argsOnly: true },
  ...
}
```

Stefan-Hanke:
OK, yeah, you would need to know the log structure inside the game, and then we would not be allowed to alter it should there be any need. I'm fine with dropping the _logView_ approach and use you proposed one. By default, the framework should assume that the full entry is shown.

Stefan-Hanke:
Note: I've pushed with no verification. Please check whether this matches your intents. When args are redacted from a move, `argsRedacted: true` is added, to allow to identify this situation cleanly in the game log.

nicolodavis:
Yes, this looks good. I'll let you remove the rest of the `logView` code and take a look again when this is ready for review.

Stefan-Hanke:
I've closed and re-opened this PR to trigger another Travis build (see [this SO answer](https://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit)).

nicolodavis:
Travis has been quite flaky recently.  Need to explore other options like https://semaphoreci.com/ when I get the time.

Stefan-Hanke:
I'm trying to repro this error. While working with this repo I'm not doing _npm i_ all the time, perhaps some error has slipped in. Doing a full clean and a fresh install now, then running coverage tests again.

nicolodavis:
I'm not clear about the semantics of `showArgs`.

Looks like setting it to `false` or not having an entry in the redact list are equivalent?

Shall we just have a simple array of strings then:
```
redactedMoves: ['clickCell', ...]
```

without `showArgs` (the semantics being that mere presence in the list gets your move args redacted)?

I thought that this is what you were trying to demonstrate in the example, but the code seems to suggest otherwise.  I think it's ok if we don't support scrubbing out the move name itself.

nicolodavis:
Might have to close and re-open this PR to trigger the Semaphore build.

Stefan-Hanke:
I don't know whether the example is particularly good since it hides the log. Suggestions? We could create another example showing this off.

nicolodavis:
That's a good point. One option would be to enable rendering the log (via the Debug UI). Another option would be to expose the log via props (we should probably do this anyway so people can create their own rendering of the log object).

What do you think?

nicolodavis:
Sounds good to me. I discovered that we already do in fact expose the `log` as a prop.

I exposed `GameLog` in `boardgame.io/ui` in https://github.com/nicolodavis/boardgame.io/commit/e9f039ef7ed5d580a70344ac655f68b36caa336f, so you should be all set.

Stefan-Hanke:
grafik

nicolodavis:
Looks good @Stefan-Hanke. I think I may have also unearthed a bug in the log implementation (unrelated to this PR).

Just want to confirm if you're observing this as well: Clicking on 'Click Cell' results in two (not one) log events being created for the current player (while other players receive just one log event).

I suspect that this is because we create one as part of the locally executed action, and another when the server gets back with a response.

Stefan-Hanke:
Yes I'm seeing this also. Let's track this separately.

src/master/master.js
```diff
@@ -138,10 +179,17 @@ export class Master {
       deltalog: undefined,
     };
 
+    const log = redactLog(
+      this.game.flow.redactedMoves,
+      state.deltalog,
```
nicolodavis: Shouldn't this be `state.log` here?

Stefan-Hanke: Hah. Tests don't care whether to use `state.log` or `state.deltalog` here. Both are `undefined` in all current test cases.

Stefan-Hanke: Yeah agree, `sync` should care abouth the whole log, whereas `update` should care about the delta one.

PR #269, opened 2018-08-29
add allOthers option to setActionPlayers

Stefan-Hanke:
This PR fixes the militia example (see #154, part 5).

There are now two moves (_playMilitia_ and _dropCards_). For dropping cards to work for players that are action players but not the current player, I had to allow calling events for those players, because otherwise they would not be allowed to change action players.

#### Checklist

### DIFF
--- a/src/core/flow.js
+++ b/src/core/flow.js
@@ -101,7 +101,8 @@ export function Flow({
     optimisticUpdate,
 
     canPlayerCallEvent: (G, ctx, playerID) => {
-      return ctx.currentPlayer == playerID;
+      const actionPlayers = ctx.actionPlayers || [];
+      return ctx.currentPlayer == playerID || actionPlayers.includes(playerID);
     },
 
     canPlayerMakeMove: (G, ctx, playerID) => {
--- a/src/core/flow.js
+++ b/src/core/flow.js
@@ -551,9 +551,16 @@ export function FlowWithPhases({
 
     // Update actionPlayers if _actionPlayersOnce is set.
     let actionPlayers = state.ctx.actionPlayers;
-    if (state.ctx._actionPlayersOnce == true) {
+    if (state.ctx._actionPlayersOnce) {
       const playerID = action.playerID;
       actionPlayers = actionPlayers.filter(id => id !== playerID);
+    }
+    if (state.ctx._actionPlayersAllOthers) {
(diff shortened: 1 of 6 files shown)


src/core/flow.js
```diff
@@ -101,7 +101,8 @@ export function Flow({
     optimisticUpdate,
 
     canPlayerCallEvent: (G, ctx, playerID) => {
-      return ctx.currentPlayer == playerID;
```
nicolodavis: Ah, actually, this was a deliberate decision to only allow the current player to be able to call events.

If you need this just so that other players can call `setActionPlayers` I would suggest using the `once` [option](https://github.com/google/boardgame.io/blob/master/src/core/turn-order.js#L37) so that you don't need to call `setActionPlayers` in the first place.

Stefan-Hanke: Unfortunately, using _once_ does not make the current player an action player, leaving the game in a deadlock situation. When all other players made their out-of-order move, the current player needs to be set as an action player.

Why should non-current players not call events?

nicolodavis: We can modify `setActionPlayers` to accept an `everyoneElseAndThenMe` option because I think this is a very common scenario (let's name it something less verbose though).

About only current players being able to call events, I think it's an easy model to reason about and also has some nice properties like being able to call `endTurn` while also triggering `endTurnIf` simultaneously and not having to worry about messing up the next player's turn.

Stefan-Hanke: I've chosen the name _allOthers_ which IMO gets the point across. There is one thing that now, when one of the players hit the _dropCards_ button, the other is allowed to click on _endTurn_. Hopefully this is only an UI issue, the event now gets dropped by the server since he's not the current player.

The basic problem is who is allowed what. In the end I still think it would be useful if a non-current player could call _setActionPlayers_, we cannot create new turnorder stuff for every other game. Let's wait where this is going...

nicolodavis: `allOthers` sounds good to me. Thanks for the changes!

About allowing other players to call events, let's keep the discussion open. My current intuition tells me that if the current turn order mechanism is lacking, we should just make it more powerful / expressive. That way we'll get the benefits of the being able to express the turn orders nicely via options and also have a simple model to reason about (without turn order code being sprinkled in the game logic itself, which I think is prone to bugs).

PR #270, opened 2018-09-01
Add utils for working with hexagonal boards

Korla:
#### Checklist


Korla:
There seems to be an issue in my environment where the wrong user is set as commit author which messes with the CLA. I will try to amend this and create a new PR.

PR #273, opened 2018-09-05
Get reachable hexagonal points within a distance

Korla:
A utility function which returns all reachable points given a start, a movement distance, and a set of blocked points

#### Checklist

### DIFF
--- a/src/ui/hex-utils.js
+++ b/src/ui/hex-utils.js
@@ -6,7 +6,10 @@
  * https://opensource.org/licenses/MIT.
  */
 
+export const isSame = a => b => a.x === b.x && a.y === b.y && a.z === b.z;
+export const createCoordinate = ([x, y, z]) => ({ x, y, z });
 const addPoint = (a, b) => ({ x: a.x + b.x, y: a.y + b.y, z: a.z + b.z });
+const isContained = (a, points) => points.some(isSame(a));
 
 /**
  * Get neighbors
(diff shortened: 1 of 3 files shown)


nicolodavis:
Nice work @Korla. What do you think about https://github.com/vonWolfehaus/von-grid?

Korla:
That project looks great, I will try and use it soon. 

I could take look at using von-grid with boardgame.io similar to how we do the chess things if you like. If I do, should I simply add it as a dependency in the main project, similar to chess.js? It looks like it would be the first part of UI which has an external dependency, but if it's ok with you I can start out like that.

nicolodavis:
I was wondering if we should provide a component toolkit in threejs (hex, card etc.) and we can use von grid as a starting point for the hex grid offering. The author seems open to people porting over his code, but we can ask explicitly to be polite and attribute accordingly.

I'm working on a tic tac toe example in threejs to see if there is anything else we need to build to allow it to play well with boardgame.io.

nicolodavis:
For an API, we have two options:

1. We can keep our API the same (in React):
<Card>
<Hex>
etc.

and have a boolean option that enables rendering to threejs for a 3D version of those components.

This will also solve some of the problems described in #274 because each component has access to its previous state and only updates itself when necessary.

2. We provide a set of threejs components that people would use natively in a threejs client.

We can play around with both as examples (similar to Chess like you suggest) and see what makes sense.

nicolodavis:
FYI, I've created a branch for Idea 1: https://github.com/google/boardgame.io/tree/3d/src/ui/3d

Started implementing Card and Deck.  Will make my way to Hex eventually.  Happy to review PR's as well in case you are interested in contributing.

Korla:
@nicolodavis I am definitely interested in contributing. Did you try anything yet, or should I just get started on a new branch?

nicolodavis:
I'm currently playing around with some ideas. Will get a branch up soon and we can work together on that.

PR #275, opened 2018-09-08
Log errors to console when debugging moves

jeanfredrik:
When using the built in debug ui to make moves, errors are caught and the message is displayed directly in the ui. There’s no way to see the call stack which makes debugging hard. I’ve added a `console.error` call (in development only) so the error also appears in the browser console.

I’m not sure how to add tests for this though. Test coverage for debug-move.js is now only 96%.

#### Checklist


jeanfredrik:
I signed the CLA

jeanfredrik:
Test coverage is now 100% but there’s one line in debug-move.js that’s not covered (yellow).

src/client/debug/debug-move.js
```diff
@@ -37,6 +37,12 @@ export class DebugMove extends React.Component {
       this.props.fn.apply(this, argArray);
     } catch (e) {
       error = '' + e;
+      if (
+        process.env.NODE_ENV === 'development' ||
+        process.env.NODE_ENV === 'test'
+      ) {
+        console.error(e);
```
nicolodavis: Use the logging API we have at https://github.com/google/boardgame.io/blob/master/src/core/logger.js instead.

jeanfredrik: The exported functions in logger.js only print the error message which is not what I wish for. Should I add another function, e.g. `errorWithStack`?

nicolodavis: You can change logger.js to wrap console.info, warn and error respectively.

jeanfredrik:
@nicolodavis I’ve made some new commits in my fork now but I’m not sure how I re-open this pull request. Should I make a new one?

nicolodavis:
I think you can re-open this and it should just work (I'm not 100% sure though). Feel free to send me a new pull request if that doesn't work.

jeanfredrik:
I can’t find any way to re-open this pull request (maybe I don’t have the required permissions), so I’ve opened a new one at #300

PR #276, opened 2018-09-12
server: add new API endpoints 'list' and 'leave'

francoijs:
This PR introduces 2 new API endpoints for the future lobby implementation:
 - _list_ allows returning a list of instances for a given game
 - _leave_ is the symmetric of _join_ ; note that after the last player leaves a game instance, it is automatically destroyed

#### Checklist

### DIFF
--- a/src/server/db/inmemory.js
+++ b/src/server/db/inmemory.js
@@ -45,11 +45,32 @@ export class InMemory {
   }
 
   /**
+   * Remove the game state from the in-memory object.
+   * @param {string} id - The game id.
+   * @returns {object} - The removed game state, or undefined
+   *                     if no game was found with this id.
+   */
+  async remove(id) {
+    if (!await this.games.has(id)) return;
+    const game = await this.games.get(id);
+    this.games.delete(id);
+    return game;
+  }
+
+  /**
    * Check if a particular game id exists.
    * @param {string} id - The game id.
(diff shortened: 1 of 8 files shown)


francoijs:
follow-up to PR #256

src/server/db/inmemory.js
```diff
@@ -44,6 +44,19 @@ export class InMemory {
     return await this.games.get(id);
   }
 
+  /**
+   * Remove the game state from the in-memory object.
+   * @param {string} id - The game id.
+   * @returns {object} - The removed game state, or undefined
+   *                     if no game was found with this id.
+   */
+  async remove(id) {
```
nicolodavis: All the DB implementations need to have the same signature.  You've implemented `remove` and `list` for the in-memory implementation, but someone could be using mongo and that code would fail.

At the same time, I don't want to put too much burden on contributors to have to update multiple DB implementations every time something changes.  I'm happy to bring the Firebase and Mongo implementations in line with this by adding changes to your PR (won't have time till next week, though).  In the meantime, feel free to get to it sooner if you have the time.

francoijs:
Sorry @nicolodavis I lost your comments (add to do a force-push to update the PR)
but the implementation of new API _remove_ and _list_ is now done for Mongo and Firebase as well.

nicolodavis:
@francoijs Solid PR. Thanks!

PR #279, opened 2018-09-24
update npm lockfile

philihp:
Updates package-lock.json file to include `requires` sections. I'm running npm 6.4.1, so I'm guessing some recent version added it...?

#### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,12 +2,20 @@
   "name": "boardgame.io",
   "version": "0.25.5",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "@firebase/app": {
       "version": "0.3.3",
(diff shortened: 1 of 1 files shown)


nicolodavis:
Thanks @philihp! Is this to fix the fact that `npm install` produces uncommitted changes, or is there a deeper problem?

philihp:
@nicolodavis Yeah, it's just that. I didn't make any changes and did an `npm install` on my machine with master. Maybe npm is a lower version on your end? If so, do you wanna make sure an `npm install` on this branch from your end doesn't revert these changes?

nicolodavis:
Ran `npm audit --fix` in b77e4b41af8adb2bfc1918bf40db566a04786a1b and it added the require sections, so I think this is now redundant.

Let me know if you still see diffs on a clean install.

PR #280, opened 2018-09-24
Enforce eslint style rules

philihp:
The most important thing here, because pretty code is pretty:

* **Adds an eslint check to enforce style on commits**

Other cool things:

* Upgrades `eslint` from 4 to 5
* Runs `prettier` from `eslint`
  * with an eslint IDE plugin, this alleviates the need for an additional prettier plugin
  * this makes it easier to ensure eslint and prettier rules don't conflict
* Turns on linter recommended settings
* Removes the import-plugin; it wasn't actually being used as all of its rules are off by default.
* Adds [eslint-unicorn](https://github.com/sindresorhus/eslint-plugin-unicorn) recommended style rules -- the most impactful of which is probably enforcing standard filenames

#### Checklist


nicolodavis:
FYI, I'm fiddling around with Semaphore settings, so the build is likely to fail.

nicolodavis:
@philihp There are a couple of problems in the package-lock.json that make the repo's packages unusable. Specifically, the introduction of @babel/code-frame always results in the final package throwing an error when used (I've run into this in the past).

Can you try syncing with `master` and generating your package-lock.json again?

philihp:
@nicolodavis I did a rebase, and got a good buid. When does the error surface?

nicolodavis:
OK, works for me now!  Earlier I was getting an error while using the package, but I don't see it anymore.  I still got some diffs on package-lock.json when running `npm install`, so added my version of it (I'm using `npm 6.2.0`).

Issue #161, opened 2018-03-20
Disable a move


rzulian:
How do we want to manage the following scenario?
You are the activePlayer. Your available moves are moveA, moveB and moveC. You pick to do moveA, but we realise that you don't have the available resources (e.g. money, VPs)  to do it. Can we rollback the move?

nicolodavis:
moveA can return undefined in this case.

nicolodavis:
Please ask on the Gitter channel if you need further assistance.

PR #287, opened 2018-10-19
Update tutorial: Illegal move hint.

Izhaki:
Related: https://github.com/nicolodavis/boardgame.io/issues/161

#### Checklist

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -62,6 +62,8 @@ a repeatable calculation of state without any side effects.
 !> The move function can receive any number of additional
 arguments that are passed to it from the call-site.
 
+?> To denote an illegal move, the move function should return undefined. Such a move will be ignored - as if the player hasn't acted.
+
 Run the app using:
 
 ```


nicolodavis:
Note that this message is just on the previous page here: https://boardgame.io/#/concepts?id=moves

Do you think it's worth adding it to the tutorial as well?

Izhaki:
Blimey! Now you have pointed this out, I remember actually reading this as I was going through the docs.

So no - I think this piece of information is too specific for something like the tutorial. And it is already document elsewhere, so why repeat?

Maybe I'm sharing too much information, but this is basically what happened:

- Read the getting started section.
- Followed the tutorial.
- Read all other parts of documentation.

Then I started a new project (poker) - at which point I had no recollection where things are in the docs - so I expected to see the documentation on 'moves' under API/Game; but there isn't much there at the moment.

So possibly if anything this should go under API/Game (as well as key concepts) instead?

PR #293, opened 2018-10-21
transform async/await in compile

philihp:
Tried to bump to latest version and encountered this error:

```
11:54:53 AM: > node scripts/build.js
11:54:54 AM: Creating an optimized production build...
11:55:12 AM: Failed to compile.
11:55:12 AM: Failed to minify the code from this file:
11:55:12 AM:  	./node_modules/boardgame.io/dist/react.js:1770
11:55:12 AM: Read more here: http://bit.ly/2tRViJ9
11:55:12 AM: npm ERR! code ELIFECYCLE
11:55:12 AM: npm ERR! errno 1
11:55:12 AM: npm ERR! fields-of-arle@0.2.0 build: `node scripts/build.js`
11:55:12 AM: npm ERR! Exit status 1
```

On line 1770, it looks to hit an async/await. Should these be transformed via babel?

(side note, I also took a stab at at trying to upgrade the project to Babel 7 -- it looks like we're currently blocked until Storyboard upgrades, and by `react-svg-loader` https://github.com/boopathi/react-svg-loader/pull/188)

#### Checklist


nicolodavis:
I see:

```
packages/server.js → dist/server.js...
(!) `this` has been rewritten to `undefined`
https://github.com/rollup/rollup/wiki/Troubleshooting#this-is-undefined
src/server/api.js
1: var _this = this;
               ^
2:
3: /*
```

when running `npm run prepare` now, so it looks like the `async-to-generator` transform causes a weird interaction with Rollup.

nicolodavis:
About whether we should run `async` through Babel, it's not so much about Node or browser support (it's supported well on both sides).

Looks like the problem is that `create-react-app` cannot minify code when a dependency uses `async / await`. We should definitely not break a `create-react-app` build, so we should get this in.

nicolodavis:
@philihp I'm still seeing the following error when running `npm run prepare`:

```
packages/server.js → dist/server.js...
(!) `this` has been rewritten to `undefined`
https://github.com/rollup/rollup/wiki/Troubleshooting#this-is-undefined
src/server/api.js
1: var _this = this;
               ^
2: 
3: /*
```

Any ideas?

philihp:
Hmm, that's weird!! To be candid, I don't have much experience with Rollup or classes in Javascript, so i'm at a loss there.

FWIW, I tried an upgrade to 0.27.0, and did not get the build issue with async/await that I had with 0.26.3, so maybe this isn't necessary anymore?

nicolodavis:
I think this might be a `create-react-app` change.  I noticed that 0.26.3 and earlier also build fine with `create-react-app` v2.  Are you also using v2?

Either way, looks like we can close this now.

philihp:
I'm still using the config from an older version, but I've been upgrading modules (like `react-dev-utils`)

PR #294, opened 2018-11-01
Generic lobby

francoijs:
Submission for a generic lobby implementation (in `lobby`) + an instance with 2 games in `examples`.

#### Checklist


examples/react/turnorder/board.js
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
```
nicolodavis: Just curious why this file appears to be added in your PR. Perhaps you based it off a commit before https://github.com/nicolodavis/boardgame.io/commit/d50015db218fa89edcf898300a7fcbf687f48317 and the rebase / merge lost something?

nicolodavis: Oh I see. Militia isn't really a game. It's a small example from the game Dominion. Let's just have TicTacToe and Chess.

francoijs: ok! fixed.

nicolodavis:
@francoijs I can't seem to get the games inside the lobby to work correctly (the room creation steps seem to work fine).

Once I get inside a game, I can't seem to make any moves that "stick".  For example, if I open two browsers and connect as different players, I would expect a move made on one to be reflected in the other.

francoijs:
> @francoijs I can't seem to get the games inside the lobby to work correctly (the room creation steps seem to work fine).
> 
> Once I get inside a game, I can't seem to make any moves that "stick". For example, if I open two browsers and connect as different players, I would expect a move made on one to be reflected in the other.

Got it! fixed and pushed in cc1b4dc1a30cb0b2fd32fd723e3071d9c9c8d9d3

nicolodavis:
@francoijs I'm still having the problem. Maybe I'm not using this correctly? Here is what I did:

1. Log in as player A.
2. Create a game and join in.
3. Open new browser, log in as player A and join game that was created in [2]. Also click on Play.
4. Refresh the browser on player A in order to reveal Play button.
5. Click on Play.

Now moves made on A don't seem to affect the game state.

francoijs:
Indeed: when refreshing, the credentials of the player are lost - a solution could be to save them in the cookie (along with the playerName).
I will implement that ASAP - thanks for the test!

nicolodavis:
Sorry about the delay in getting to this (been travelling a bit).  Will take a look sometime this week.

nicolodavis:
Finally had some time to spend on this (sorry about the delay).

**High-level comment**: I think we should move more stuff into `lobby/react.js`.

For example, `login-form.js` should probably be part of the package rather than requiring users to implement it.

The ideal end-user experience should be something like:

```js
import { Lobby } from 'boardgame.io/react';

const importedGames = [
  { game: GameTicTacToe, board: BoardTicTacToe },
  { game: GameChess, board: BoardChess },
];

ReactDOM.render(<Lobby games={importedGames} ...otherOptions />);
```

and all the views (create / join / login) are handled under the hood. Right now I think we're having the user do too much (as demonstrated by `examples/react/lobby`).

francoijs:
Ok, let's try this!

nicolodavis:
Solid PR, @francoijs. Thank you!

barraponto:
@francoijs IDK if it's too late, but before we settle on the API for the stable release:
`minPlayers` and `maxPlayers` some games (like Brazilian Truco) have more restricted player numbers: either 2, 4 or 6. For my custom lobby implementation (I was unaware of this PR) I used `allowedPlayers` and it returns an Iterable (an array or whatever else).

barraponto:
BTW, I have no idea whether we could support infinite iterables in `allowedPlayers`, but it'd be interesting to give it a try. Games like setgame.com, for instance, have no reason to limit player numbers.

nicolodavis:
@barraponto Having a function that returns an iterable sounds good to me.

Maybe we can have the semantics be:
1. If the function is not defined, then there is no restriction on player count.
2. If the function returns an iterable, only the values returned are supported.

(that way you don't have to worry about infinite iterables).

Do you have time to make a PR for this? I have way too much on my current TODO list at the moment to get to this soon.

barraponto:
@nicolodavis I'm not sure on the infinite iterables. If I could be assured it is not infinte, `Array.from(iterable)` would be the way to go (useful to display UIs). Maybe we could assume infinite is not supported and call it a day?

francoijs:
@barraponto Good idea. +1 for the solution of `allowedPlayers` returning an iterable generating a (finite) array of allowed player numbers. 
We can always add support for infinite player numbers later on.

PR #300, opened 2018-11-12
Log errors to console when debugging moves

jeanfredrik:
When using the built in debug UI to make moves, errors are caught and the message is displayed directly in the ui. There’s no way to see the call stack which makes debugging hard. With this pull request, the debug UI prints the object to the console with `console.error`.

Continued from #275

#### Checklist


src/core/logger.js
```diff
@@ -9,10 +9,11 @@
 const DEV =
```
nicolodavis: Are you able to see any log messages on the browser? Unless someone uses a Webpack plugin (or equivalent) to set `process.env.NODE_ENV` here I would only expect these log messages to show up on Node.

This is not directly related to this PR, but I thought I'd ask in case you add this new log message and don't see the output.

philihp:
One thing that has been HUGE in helping me debug my move functions is to test them in isolation from the whole React/Boardgame.io environment. Maybe this will help? I had a margarita for lunch so maybe I am just drunk.

https://github.com/philihp/fields-of-arle/blob/master/src/game/actionOptions/__tests__/bakehouse.test.js

nicolodavis:
+1 to what Philihp said.

No harm in getting this in also, though.

PR #316, opened 2018-12-15
Adds drag and drop capability to 2d token

flamecoals:
#### Checklist

### DIFF
--- a/src/ui/2d/token.js
+++ b/src/ui/2d/token.js
@@ -30,6 +30,12 @@ import { Square } from './grid';
  *   onClick - Called when the token is clicked.
  *   onMouseOver - Called when the token is mouse over.
  *   onMouseOut - Called when the token is mouse out.
+ *   draggable - Whether a Token is draggable or not.
+ *   shouldDrag - Whether a draggable token should start drag.
+ *   onDrag - Called when a token was dragged (moved).
+ *            Parameter contain { x, y, originalX, originalY }.
+ *   onDrop - Called when the token was dropped after dragging.
+ *            Parameter contain { x, y, originalX, originalY }.
  *
  * Usage:
  *
@@ -60,6 +66,11 @@ export class Token extends React.Component {
     onMouseOut: PropTypes.func,
     children: PropTypes.element,
     animationDuration: PropTypes.number,
+    draggable: PropTypes.bool,
+    shouldDrag: PropTypes.func,
+    onDrag: PropTypes.func,
+    onDrop: PropTypes.func,
+    svgRef: PropTypes.object,
   };
 
   static defaultProps = {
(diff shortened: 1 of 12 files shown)


flamecoals:
FYI This decrease in coverage does not come from the files I changed:
image

jasonharrison:
@flamecoals is in-flight right now and his internet pass expired.  He asked me to post this GIF demonstrating drag/drop:

GIF of chess drag and drop

nicolodavis:
For some reason this PR seems to be affecting the overall project coverage status (on master) as reported by Coveralls. Weird!

https://coveralls.io/github/nicolodavis/boardgame.io

flamecoals:
> For some reason this PR seems to be affecting the overall project coverage status (on master) as reported by Coveralls. Weird!
> 
> https://coveralls.io/github/nicolodavis/boardgame.io

How so? This is not even merged yet, very weird. 
Maybe the badge is the minimum coverage of all branches? 3d seems to have only 89% coverage, thats probably why?

nicolodavis:
No, for some reason Coveralls thinks that this is a commit made on the `master` branch (it doesn't seem to distinguish between PR's and actual commits on `master`).

I'm not sure if it is Coveralls that's broken or the CI that's sending over incorrect branch data for PR's.

src/ui/2d/token.js
```diff
@@ -60,25 +66,96 @@ export class Token extends React.Component {
...
+    onDrop: PropTypes.func,
+    svgRef: PropTypes.object,
   };
 
   static defaultProps = {
     animationDuration: 750,
     template: Square,
   };
 
+  _startDrag = e => {
```
nicolodavis: Have you taken a look at Card / Deck's drag-n-drop implementation?  Any reason to not do the same here?

I like the fact that this doesn't use an external dependency.  Maybe we should also move over the card / deck implementation to do the same?

flamecoals: I had a look before I implemented this, but it seems to me that the library we are using for the cards (react-dragtastic) does not support mobile (see https://github.com/chrisjpatty/react-dragtastic/issues/36 ...), which is a deal breaker for my application. Also, I wasn't sure if it would support internal SVG elements, because all examples are using divs and other HTML elements. 

The implementation for this PR works both in mobile and SVG.

nicolodavis: OK, I'll try to move over the react-dragtastic implementation to be more in line with this then.

nicolodavis:
Some behavioral comments:

1. Dragging a piece should cause it to appear on top of everything else (right now you can drag a white piece and it appears below the black).

2. Dragging a piece outside the left or bottom sides of the board results in an error.

3. Dragging a piece out and back into the original location should probably result in the same state prior to dragging (right now it still leaves the moves highlighted).

src/ui/2d/token.js
```diff
@@ -60,25 +66,96 @@ export class Token extends React.Component {
...
+    if (this.props.draggable && this.props.shouldDrag(this.getCoords())) {
+      e = e.touches ? e.touches[0] : e;
+      this.setState({
+        ...this.state,
+        dragged: { x: e.pageX, y: e.pageY },
```
nicolodavis: Where are `x` and `y` used?  It seems like you're effectively using `dragged` as a boolean in this file?

flamecoals: Line 96 and 97 :)

Happy new year 🎆 😄

flamecoals:
> Some behavioral comments:
> 
> 1. Dragging a piece should cause it to appear on top of everything else (right now you can drag a white piece and it appears below the black).
> 2. Dragging a piece outside the left or bottom sides of the board results in an error.
> 3. Dragging a piece out and back into the original location should probably result in the same state prior to dragging (right now it still leaves the moves highlighted).

Ok, I will take a look into this later and get back to you.

src/ui/2d/token.js
nicolodavis: Ah missed that. Thanks! :)

flamecoals:
@nicolodavis These issues should be addressed now. I also needed to add some logic to handle mobile better, which is so annoying, because touch screen behaves a little bit different than mouse. I am using this in my new game, and it works well in mobile with the right CSS to avoid scrolling. 

Do you have any idea when you will merge the 3d branch into master? because I am having to maintain a separate fork in order to be able to use this in my project. Do you need help with anything in order to merge? I saw that there is a lot of untested code in the 3d folder.

flamecoals:
The code I sent works in desktop, android but not iOs. I am changing it to work everywhere. It is very ugly but each environment behaves differently...

flamecoals:
OK, the code now works on iOs, Android, Chrome desktop... I spent hours and hours trying to figure out a way to abstract away all this mess, and I think I succeeded :D

nicolodavis:
I'll try to merge the 3d branch in after this PR. There hasn't been much progress on the 3D components, so maybe I'll scrap it for the v1 release and add it as a nice to have for later.

flamecoals:
FYI, I just found out that the click isnt working for Android. I will fix it soon... wait for merging

PR #327, opened 2018-12-26
Add Immer to other events

### LINKED ISSUE #323: Immer not added for flow functions
Immer was only added for moves, but should also work for flow functions like `onMove`.

pgoodrich:
Resolves #323 for:
`onTurnBegin`
`onTurnEnd`
`onMove`
`onPhaseBegin`
`onPhaseEnd`

### DIFF
--- a/src/core/flow.js
+++ b/src/core/flow.js
@@ -311,9 +311,11 @@ export function FlowWithPhases({
     if (conf.onPhaseBegin === undefined) {
       conf.onPhaseBegin = G => G;
     }
+    conf.onPhaseBegin = produce(conf.onPhaseBegin);
     if (conf.onPhaseEnd === undefined) {
       conf.onPhaseEnd = G => G;
     }
+    conf.onPhaseEnd = produce(conf.onPhaseEnd);
     if (conf.movesPerTurn === undefined) {
       conf.movesPerTurn = movesPerTurn;
     }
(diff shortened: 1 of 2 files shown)


pgoodrich:
Signed the CLA

pgoodrich:
Misunderstood the immer library syntax - making updates

src/core/flow.js
```diff
@@ -363,7 +364,10 @@ export function FlowWithPhases({
 
   // Helper to perform start-of-phase initialization.
   const startPhase = function(state, config) {
-    const G = config.onPhaseBegin(state.G, state.ctx);
```
nicolodavis: Couple of comments:

1. Immer supports [currying](https://github.com/mweststrate/immer#currying).
2. Immer returns the input state if nothing is modified so I don't think you need the `if (result)` check.

So, I think this could probably be written more concisely as:

```js
const fn = produce(config.onPhaseBegin);
const G = fn(state.G, state.ctx);
```

You could also just curry everything during initialization [here](https://github.com/nicolodavis/boardgame.io/blob/master/src/core/flow.js#L311).

pgoodrich: @nicolodavis Do you have a preference for if things are curried doing initialization or not? Correct me if I'm wrong but currying during initialization would introduce an else structure to handle defined `conf.onPhaseBegin` 
```
    if (conf.onPhaseBegin === undefined) {
      conf.onPhaseBegin = G => produce(G);
    } else {
      conf.onPhaseBegin = produce(conf.onPhaseBegin);
    }
```

nicolodavis: We already perform the `undefined` check [here](https://github.com/nicolodavis/boardgame.io/blob/master/src/core/flow.js#L310) and ensure that all the callbacks contain either a user provided function or a default `G => G` identity function.

All you need to do is wrap it around `produce` after that I think? Don't think you will introduce a new else structure.

pgoodrich:
@nicolodavis Updated

PR #332, opened 2019-01-06
Upgrade Babel 7

philihp:
Yak shave to end all yak shaves

* Upgrades to Babel 7
* ~~Upgrades Storybook~~
* Upgrades Rollup and config, removing deprecated things
  * Has a warning `(!) postcss plugin: The onwrite hook used by plugin postcss is deprecated. The generateBundle hook should be used instead.`, which I believe may be fixed by this: https://github.com/egoist/rollup-plugin-postcss/pull/113
* Upgrades Firebase
* Upgrades Webpack
* Upgrades React

## Tests run

* [x] `npm run test`
* [x] `npm run prepare`
* [x] `npm run docsify`
* [x] `npm run storybook`
* [ ] `npm run dev`
* [x] `npm run test:integration`
* ...anything else to try out?


nicolodavis:
Awesome!

Can you take a look at the integration test (`npm run test:integration`)?  Looks like it's failing (although the script itself is not reporting itself as failed, which I should fix).

Reference: https://semaphoreci.com/nicolodavis/boardgame-io/branches/pull-request-332/builds/1

package.json
```diff
@@ -67,29 +67,29 @@
...
+    "@babel/preset-env": "^7.2.3",
     "@storybook/addon-actions": "^3.2.16",
     "@storybook/addon-knobs": "^3.3.10",
     "@storybook/addon-options": "^3.2.16",
-    "@storybook/react": "^3.4.11",
     "@storybook/addons": "^3.4.11",
-    "babel-cli": "^6.26.0",
-    "babel-core": "^6.26.3",
+    "@storybook/react": "^4.1.4",
+    "ajv": "^6.6.2",
```
nicolodavis: What is this dependency for?

philihp: `ajv-keywords` wants it. Who uses that? I dunno man... dependency hell. I could take it or leave it, but it resolves this warning:

`npm WARN ajv-keywords@3.2.0 requires a peer of ajv@^6.0.0 but none is installed. You must install peer dependencies yourself.`

philihp:
Holding tight, `npm run dev` needs bumping too. Looks like babel-watch just got a version that supports 7, too.

philihp:
Nevermind... babel-watch still doesn't support Babel 7 yet. I think I can accomplish the same thing though with `nodemon` or `@babel/cli` (which is already included)... but they don't play well when they have to compile things outside of their dir (e.g. module-resolver going to `../../packages`). @nicolodavis any feelings/aversions toward creating an `examples/react/package.json` and putting the examples dependencies in that, similar to how `examples/react-native` has its own package.json?

Pros:
* Removes *A LOT* of devDependencies from the root project (e.g. chess.js)
* Removes need for `babel-watch` entirely.
* `module-resolver` [here](https://github.com/nicolodavis/boardgame.io/blob/master/.babelrc#L18-L22) now goes to `../../dist/` which is packaged with Babel and Rollup and what is ultimately distributed, instead of `../../package/` which is just built on-the-fly with Babel.

Cons:
* When working on an example project, you don't have the convenience of being able to make a change to core/client. Maybe this is a good thing though? You can still make changes to the boardgame.io library and validate them with unit tests using `npm run test:watch`, perhaps not with an example app.

flamecoals:
I am just afraid that not being able to validate with the example app will make our life harder to manually test features. And this is important specially for some browser-dependent features like drag and drop that I am developing now.

philihp:
Could the drag-and-drop components be developed within `/examples/react/`, and then moved into `/src/ui/` when they're more solidified?

flamecoals:
Not really, because I am changing Token.js which is already included in /src/ui/... I could make a full new copy of the file and use that instead for development... That's a way.

I think the drag and drop should be mostly done now, but I just wonder that this might also happen for other features that are not very easily unit-testable.

flamecoals:
But at the same time, I love TDD and I think that forcing people to look more into the tests and less into manually testing is a good thing. So I don't know, im just pointing the trade-offs here :)

philihp:
Yep yep, it is a tradeoff for sure.

Quick question... are you using `npm run storybook` for developing the components?

nicolodavis:
I use the examples a lot when developing to do quick sanity tests. I think it should still be possible to have a separate package.json with all the dev dependencies and also "symlink" the boardgame.io dependency to the code in the repo so that you can make changes and see them live, no?

Also, I don't think anyone uses storybook. I'm in favor of removing it altogether.

nicolodavis:
Any thoughts about the Babel regenerator runtime error that you get when running the integration test?

nicolodavis:
Removed storybook in https://github.com/nicolodavis/boardgame.io/commit/f853da8bf2fdaa830c7a31d6fca77ca43a2eb551.

nicolodavis:
I see.  OK, then let's just keep the examples in the same package.json as the main project until we can figure out a way to split them and still retain the ability to hot reload library changes.

package-lock.json
```diff
@@ -1895,21 +1895,16 @@
...
+      "resolved": "https://registry.npmjs.org/@firebase/polyfill/-/polyfill-0.3.4.tgz",
+      "integrity": "sha512-XvHewHdBdR6frMJszgM1dKo9giH1BoliX2iMssot9W89yItxhbjjwzuv7NY49uT/TKg9dYZV+mwQiwC7YNF76w==",
       "dev": true,
       "requires": {
         "core-js": "2.5.5",
         "promise-polyfill": "7.1.2",
         "whatwg-fetch": "2.0.4"
       },
       "dependencies": {
-        "chokidar": {
```
philihp: this was weird.. it had the same hash as whatwg-fetch, too.

philihp:
@nicolodavis integration test works now, that's a clever test... I like it. `rollup.npm.js` needed an option added.

philihp:
I guess last remaining unsolved question is how to get the examples webapp to hot-reload changes to a package outside of its dir.

nicolodavis:
Thanks @philihp!

Let's punt that to another PR.  This is a good change to get in right now.

Made one change, which is to remove the async-to-generator altogether (which avoids all the regeneratorRuntime hassle and also results in a smaller bundle).

rollup.npm.js
```diff
@@ -149,28 +140,22 @@ export default [
...
+    ],
     plugins: plugins.concat([
       builtins(),
-      commonjs(),
       resolve({ browser: true, preferBuiltins: false }),
+      commonjs(),
       replace({
         'process.env.NODE_ENV': JSON.stringify('production'),
       }),
-      uglify(
```
nicolodavis: Missed this. Why are we no longer minifying the code here?

nicolodavis: I'm guessing you deleted this because it was generating some errors and forgot to add it back. 
Added `rollup-plugin-terser` instead.

nicolodavis:
Also, about `npm run dev`, I think we could use `nodemon / babel-node` as shown here: https://hackernoon.com/using-babel-7-with-node-7e401bc28b04

rollup.npm.js
philihp: Ah yeah, I removed it to simplify things. At first it was having some errors so I upgraded it but then the config for it was broken. `rollup-plugin-terser` looks a better solution anyway, and zero config 💯

PR #334, opened 2019-01-15
Update Husky

philihp:
Nothing of note changed, just bumping packages.


nicolodavis:
It was easier for me to just implement this change in https://github.com/nicolodavis/boardgame.io/commit/ca001657a3dbd86fa679c555e1c17f6421d67072 rather than deal with the `package-lock.json` merge required here.

PR #338, opened 2019-01-18
Fixed empty player name treated as empty slot

lime-cake:
An empty string evaluate to false in Javascript, because whoever created JS says f*ck logic. 

'if (gameMetadata.players[playerID].name) 'will not run when name isn't set, or name is set to anything that evaluate to false, including an empty string. To avoid empty player name causing the engine to treat player as empty slot, changed it check if it's not null or not undefined.

#### Checklist

* [V] Use a separate branch in your local repo (not `master`).
* [V] Test coverage is 100% (or you have a story for why it's ok).


nicolodavis:
Thanks for the fix!

How about we enforce that playerName always be a valid non-empty string instead?

lime-cake:
I think that might be best. I briefly considered the possibility of having players with no name, like blind, anonymous match (maybe for tournament), but then just giving a random name will work for those. So there isn't any chance for empty name to be needed. I'll update it

src/server/api.js
```diff
@@ -159,7 +159,7 @@ export const createApiServer = ({ db, games }) => {
     if (!gameMetadata.players[playerID]) {
       ctx.throw(404, 'Player ' + playerID + ' not found');
     }
-    if (gameMetadata.players[playerID].name) {
```
nicolodavis: Actually, maybe a cleaner way would be to use:
```
gameMetadata.players[playerID].name = playerName;
gameMetadata.players[playerID].occupied = true;
```

down below at line 166

and then you can just check:

```
if (gameMetadata.players[playerID].occupied) {
  ctx.throw(409, 'Player ' + playerID + ' already occupied');
}
```

here.

That would give us flexibility to also have slots without a name.

nicolodavis: Breaking changes are OK for an alpha release.  I'll just make sure to document this carefully in the changelog.

nicolodavis:
@lime-cake What's the status of this PR?

nicolodavis:
Closing this for now.  Feel free to re-open if you get back to this PR.

PR #341, opened 2019-01-20
Allow result of onPhaseBegin to influence turn order

### LINKED ISSUE #339: onPhaseBegin/turnOrder perhaps not working as intended
Here's the code from https://github.com/nicolodavis/boardgame.io/blob/133e23af5f0e1fdf5b73ede0e734195e4a1205fe/src/core/flow.js#L379

```
    let G = config.onPhaseBegin(state.G, state.ctx);
    let ctx = InitTurnOrderState(state.G, state.ctx, config.turnOrder);
```

So the problem here is that I modify the G state during onPhaseBegin. So shouldn't that G argument to initTurnOrderState be G and not state.G?

djMax:
#### Checklist

Fixes #339 by using returned G instead of state.G, added test handling.

### DIFF
--- a/src/core/flow.js
+++ b/src/core/flow.js
@@ -376,7 +376,7 @@ export function FlowWithPhases({
   // Helper to perform start-of-phase initialization.
   const startPhase = function(state, config) {
     let G = config.onPhaseBegin(state.G, state.ctx);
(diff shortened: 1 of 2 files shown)


djMax:
I signed it!

PR #343, opened 2019-01-20
Feature/single koa

### LINKED ISSUE #330: Koa server and transport layer port
The Authentication server (Koa) and transport layer (socket.io) use different port, respectively port and port+1. Where port is passed as a param when creating 'Server'

This is problematic for deploying to most hosting popular service, such as Heroku, which require using a single port passed as system variable.

lehaSVV2009:
#### Checklist

fixes #330


lehaSVV2009:
> I signed it!

lehaSVV2009:
@nicolodavis Not saying it's the best implementation. But current implementation will not require major version change..

From my point of view it would be better to be able to do smth like that:
_For single port_
```
import { Server } from 'boardgame.io/server'

// It runs single server only with http and socket apis
Server.run(8080, () => {})
```

_For multiple ports/deployments_
```
import { ApiServer } from 'boardgame.io/server'

ApiServer.run(8090, () => {})
```
```
import { SocketServer } from 'boardgame.io/server'

SocketServer.run(12345, () => {})
```
But:
1. It will require new major version.
2. Not sure what should happen with DB connections.

So, @nicolodavis any thoughts? Would current PR implementation be enough for now?

docs/api/Server.md
```diff
@@ -20,9 +20,10 @@ The `games` argument takes a list of game implementations
 An object that contains:
 
 1. run (_function_): A function to run the server.
-   Signature: (port, callback) => {}
+   Signature: (port, callback, apiPort) => {}
 2. app (_object_): The Koa app.
 3. db (_object_): The `db` implementation.
+4. singlePort (_boolean_): An option to run api and app using single port or two different ports. `false` by default.
```
nicolodavis: `singlePort` is part of the arguments to `Server()`.  This section documents the properties of the return value.

It's confusing because one of the lines above says "with the options shown below", but it is actually referring to the section below this.

docs/api/Server.md
```diff
@@ -20,9 +20,10 @@ The `games` argument takes a list of game implementations
 An object that contains:
 
 1. run (_function_): A function to run the server.
-   Signature: (port, callback) => {}
+   Signature: (port, callback, apiPort) => {}
```
nicolodavis: I feel like we don't need `port`, `apiPort` and `singlePort`.

How about we make the semantics be:

1. `port` is the only required field.
2. If `apiPort` is not specified, it defaults to the same value that was provided in `port` (i.e. single port mode).
3. If `apiPort` is specified with a different value, then multiple ports are used.

`singlePort` is removed entirely.

lehaSVV2009: Sounds good, I'll try!

docs/api/Server.md
lehaSVV2009: Fixed

src/server/index.js
```diff
@@ -33,23 +33,36 @@ export function Server({ games, db, transport }) {
...
+    run: async (port, callback, apiPort) => {
       await db.connect();
-      let apiServer = await api.listen(port + 1);
-      let appServer = await app.listen(port, callback);
-      logger.info('listening...');
+
+      let apiServer;
+      if (!singlePort) {
+        apiServer = await api.listen(
+          apiPort ? apiPort : port ? port + 1 : null
```
djMax: It might be safer to wrap port/apiPort in Number(), because if it's a string, like an env var, it will end up appending the character 1. Which is not the intent.

lehaSVV2009: I thought about smth like Number.isInteger check. But maybe wrap is also good. One more fix in this PR :-)

docs/api/Server.md
nicolodavis: That's fine. I'll document this clearly in the changelog.

nicolodavis:
Can you also remove `boardgameio.min.js` from this PR?  I usually update it when I run a release.

djMax:
The “argument after callback” bit is very non-idiomatic for node. Maybe it would be better to make the first argument take either a number (meaning port) or an object with port/apiPort and whatever other options we might need in the future?

nicolodavis:
Yeah, that's a good idea.

@lehaSVV2009 What do you think?

lehaSVV2009:
I like it! Will be done soon.

lehaSVV2009:
@nicolodavis @djMax The pr is updated. And I have 2 questions here:
1. Should I move some stuff like routes creating from `server.run()` to `new Server()`? i.e. routes creating is common for multiple servers and for single. But will need time for tests changes :)
2. Now I'm running db before adding it to api-server as a parameter. Will it be ok? I tested in-memory db and it looks good, but not sure if mongo/firebase will work ok.

nicolodavis:
1. What is the route creation that you are referring to?
2. This should be ok.

lehaSVV2009:
> 1. What is the route creation that you are referring to?
> 2. This should be ok.

Sorry, the answer will be a bit complex to read.. Function `addApiToServer({ app, db, games })` consists of `router = new Router();`, router configs like `router.get('/games', ...)`, `app.use(routes)` and `app.use(cors())`. I thought about moving `router = new Router()` and `router.get('/games')` to a separate function that returns configured router. And this function can be called inside Server constructor (`new Server()`), but not inside server.run method. So `server.run` might be a bit faster, cause it will not contain full api creation and config.

nicolodavis:
Oh I see. I think it's OK to leave it as it is for now (I don't think people will care whether we execute code inside `run()` or in the constructor because all of this will go inside one file and run in order (it would make a difference if we were running `run()` repeatedly, but that is not the case).

lehaSVV2009:
@nicolodavis @djMax Any other comments? As I understand, this PR can be merged only after at least one approval.

nicolodavis:
Looks good to me.  Thanks @lehaSVV2009!  Merging now.

src/server/index.js
```diff
@@ -33,23 +53,44 @@ export function Server({ games, db, transport }) {
...
+
+      // API
+      let apiServer;
+      if (!serverRunConfig.apiPort) {
+        addApiToServer({ app, db, games });
+      } else {
+        // Run API in a separate Koa app.
+        const api = createApiServer({ db, games });
+        apiServer = await api.listen(
+          Number(serverRunConfig.apiPort),
```
lehaSVV2009: @nicolodavis Quick note here. There is no need to have `Number()` wrapper here. koa is running on 51234 port if koa.listen("51234") is called. That made sense to wrap with `Number()` when code `port + 1` used. Not now.

nicolodavis: Ah, I see. Ok will remove that then.

Issue #282, opened 2018-10-05
UI framework


nicolodavis:
Just recording the current state and future work planned.

We're going to ship both 2D and 3D (using threejs) components!  The two sets of components will be created using the same API so you can seamlessly switch between the two.

I've started a new branch `3d` to track development of this. I've split the UI framework into 2d and 3d directories:

```
src/ui/
├── 2d
│   ├── card.css
│   ├── card.js
│   ├── card.test.js
│   ├── deck.css
│   ├── deck.js
│   ├── deck.test.js
│   ├── grid.js
│   ├── grid.test.js
│   ├── hex.js
│   ├── hex.test.js
│   ├── logo.js
│   ├── logo.test.js
│   ├── token.js
│   ├── token.test.js
│   ├── ui.js
│   ├── ui.ssr.test.js
│   └── ui.test.js
├── 3d
│   ├── card.css
│   ├── card.js
│   ├── deck.css
│   ├── deck.js
│   ├── grid.js
│   ├── hex.js
│   ├── token.js
│   └── ui.js
├── card.js
├── deck.js
├── grid.js
├── hex.js
├── hex-utils.js
├── hex-utils.test.js
├── token.js
├── ui-context.js
└── ui.js
```

Each component has a file in the root directory (which is exposed as the public API) and one in each of the 2d and 3d directories.  The root component just delegates to the appropriate rendering component based on a prop that you pass to the `<UI>` element.

For example:

```
<UI>
  <Deck><Card/></Deck>
</UI>
```

renders DOM elements as cards and decks, whereas:

```
<UI three={true}>
  <Deck><Card/></Deck>
</UI>
```

renders to a threejs canvas.

Things missing in the 3D component set:
- [ ] hex grids
- [ ] grids
- [ ] tokens

Things missing in the 2D component set:
- [ ] ability to drag / drop tokens (currently only implemented for cards / decks)

nicolodavis:
Let's keep it as a separate branch for now and work on that. I'll merge it in once the 3D components have feature parity with the 2D components.

Korla:
Alright, I'll start working on it on that branch on my fork, and will make PR(s) to your branch as necessary.

flamecoals:
FYI, I am working now on the "ability to drag / drop tokens (currently only implemented for cards / decks)" for the 2D component

harviu:
Is anyone doing the grid, hex map or tokens right now? I am interested in doing them.

[...]

PR #352, opened 2019-01-28
3D Grid and Token

harviu:
#### Checklist

Implemented 3D Grid and Token. #282 
Created a 3D Chess example.

Drag n Drop or animation are not implemented for tokens yet.


nicolodavis:
@harviu This looks great!

I have a couple of concerns about the 3d chess pieces:

1. There is no license on the repo you cloned it from, so we shouldn't make assumptions about whether the author is OK with others copying their work.

2. The .obj files add 17 megabytes of binary data to this repo that is going to be a cost that everybody incurs when they clone the repo for local development.

Shall we make it a checkers game instead?  I can probably whip up a quick checkers piece model in Blender.

RGBKnights:
move heavy weight objects out of this repo into another using [GIT LFS](https://git-lfs.github.com/) then you can have the best of both worlds... I do this often with my AI research and game development.

nicolodavis:
Nice tip, @RGBKnights.  Definitely something to consider if we end up importing multiple large objects into the repo.  Probably unnecessary for the present PR since I think the example would work just as well with Checkers (which uses a single model).

harviu:
> @harviu This looks great!
> 
> I have a couple of concerns about the 3d chess pieces:
> 
> 1. There is no license on the repo you cloned it from, so we shouldn't make assumptions about whether the author is OK with others copying their work.
> 2. The .obj files add 17 megabytes of binary data to this repo that is going to be a cost that everybody incurs when they clone the repo for local development.
> 
> Shall we make it a checkers game instead? I can probably whip up a quick checkers piece model in Blender.

Yep, I think a checkers game example will work.

I also found some low ploy piece models here: https://sketchfab.com/marcelo.medeirossilva, which is less than 200 kilobytes in all and under CC license.

harviu:
> Could you add some unit tests for the code you added in `3d/grid.js` and `3d/token.js`?

I am working on the unit tests, but there seems to be some problem related to using Eyzume with THREE.js.

nicolodavis:
> I also found some low ploy piece models

Cool. I'll leave it to you to decide if you want to use that or switch to checkers. I guess since we already have a 2D Chess example, having a 3D one would be nice.

harviu:
Have changed the model to low poly and fix some dispatch bugs.

Still struggling with unit test.

nicolodavis:
The code in `src/ui` is still experimental so don't worry about testing absolutely everything.  Just get something in so that others can build on top of it.

harviu:
I finally add some unit tests.

src/ui/3d/grid.js
```diff
@@ -7,17 +7,178 @@
...
+  }
+
+  componentWillUnmount() {
+    this.ctx.remove(this.boardGroup);
+  }
+
   render() {
-    return null;
+    return (
+      <UIContext.Consumer>{ctx => this._drawSquare(ctx)}</UIContext.Consumer>
```
nicolodavis: Let's call this `context` to avoid confusing it with boardgame.io's `ctx`? Also take a look at deck.js which splits the context component into a separate wrapper. Let's use the same style here for consistency?

nicolodavis:
Thanks for the PR @harviu.

PR #353, opened 2019-01-29
Add custom renderer parameter to lobby + clean up code

lehaSVV2009:
#### Checklist


src/lobby/react.js
```diff
@@ -286,17 +205,110 @@ class Lobby extends React.Component {
...
+          playerName={playerName}
+          onClickJoin={this._joinRoom}
+          onClickLeave={this._leaveRoom}
+          onClickPlay={this._startGame}
+        />
+      );
     });
+  };
+
+  render() {
```
lehaSVV2009: Most changes are here:
* `render` method was moved to the end of the class
* `renderer` prop was added here
* rooms rendering was simplified.

src/lobby/react.js
```diff
@@ -32,6 +38,12 @@ import LobbyCreateRoomForm from './create-room-form';
  *   A React component that provides a UI to create, list, join, leave, play or spectate game instances.
  */
 
+// TODO add option to long-polling like <Lobby polling={{ period: 5000 }} ... />
```
lehaSVV2009: Sorry for many todos. I just don't think it's a good idea to create github issues before any discussions.

src/lobby/react.js
```diff
@@ -263,7 +182,7 @@ class Lobby extends React.Component {
       return;
     }
 
-    let multiplayer = undefined;
+    let multiplayer = null;
```
nicolodavis: I think there are `multiplayer === undefined` checks that trigger singleplayer mode, so you will break that by changing this to `null` (which was originally fixed in https://github.com/nicolodavis/boardgame.io/commit/d7d6b445cdf9d1bd740bbf4e1e7c139c6ffcc089).

packages/react.js
```diff
@@ -7,6 +7,6 @@
  */
 
 import { Client } from '../src/client/react.js';
-import Lobby from '../src/lobby/react.js';
+import Lobby, { LobbyPhases } from '../src/lobby/react.js';
```
nicolodavis: Is this change really necessary? I would prefer to not expose too many types to the public API (requires documentation and generally adds to the cognitive overhead).

src/lobby/react.js
nicolodavis: I've copied these over to #354.  You can remove these from this PR and we can discuss there.

packages/react.js
lehaSVV2009: I don't think it's really necessary, I'll remove it.

src/lobby/react.js
lehaSVV2009: Thanks!

src/lobby/react.js
lehaSVV2009: oh, thanks, I'll revert it back. But, in general, that's a bit strange..

src/lobby/room-instance.js
```diff
@@ -98,6 +98,7 @@ class LobbyRoomInstance extends React.Component {
           {inst.players.map(this._createSeat).join(', ')}
         </td>
         <td key={'cell-buttons-' + inst.gameID}>
+          {/* TODO show leave button even if game is ready to play */}
```
lehaSVV2009: @nicolodavis Could you please add this TODO to #354 improvement as well?

nicolodavis: Done.

lehaSVV2009: Thanks!

lehaSVV2009:
@nicolodavis it looks like npm run test:integration fails on master. Locally I have the following error: 
```
Failed to minify the bundle. Error: static/js/main.38a01738.chunk.js from Terser
TypeError: Cannot read property 'minify' of undefined
```

nicolodavis:
It's a bug with `terser`. They should roll out a fix soon: https://github.com/terser-js/terser/issues/251

lehaSVV2009:
The approach from the PR is used here - https://lehasvv2009.github.io/territories/#/online (room creating might be a bit long cause of heroku wake up). Probably just waiting for reviews.

nicolodavis:
@lehaSVV2009 Sorry about the delay in getting to this.  Looks good!

lehaSVV2009:
@nicolodavis Thanks!

PR #358, opened 2019-02-04
Missing dependencies for UI package

flamecoals:
These two dependencies were marked as devDependency, which broke [freeboardgame.org](http://freeboardgame.org) tests. Adding them as regular deendency fixed our issues.
See errors:
```
 FAIL  src/Home/Home.test.tsx                                                                                                                                                          
  ● Test suite failed to run                                                                                                                                                           
                                                                                                                                                                                       
    Cannot find module '@tweenjs/tween.js' from 'ui.js'                                                                                                                                
                                                                                                                                                                                       
    However, Jest was able to find:                                                                                                                                                    
        './checkerboard.test.tsx'                                                                                                                                                      
        './checkerboard.tsx'                                                                                                                                                           
                                                                                                                                                                                       
    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['ts', 'tsx', 'js'].                                          
                                                                                                                                                                                       
    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string
```

```
FAIL  src/App/Game/Game.test.tsx                                                                                                                                                      
  ● Test suite failed to run                                                                                                                                                           
                                                                                                                                                                                       
    Cannot find module 'three' from 'ui.js'                                                                                                                                            
                                                                                                                                                                                       
    However, Jest was able to find:                                                                                                                                                    
        './checkerboard.test.tsx'                                                                                                                                                      
        './checkerboard.tsx'                                                                                                                                                           
                                                                                                                                                                                       
    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['ts', 'tsx', 'js'].                                          
                                                                                                                                                                                       
    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string   
```

#### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1257,8 +1257,7 @@
     "@tweenjs/tween.js": {
       "version": "17.2.0",
       "resolved": "https://registry.npmjs.org/@tweenjs/tween.js/-/tween.js-17.2.0.tgz",
-      "integrity": "sha1-IfibcJuvxLMDra56g7TzWg2eR5Y=",
-      "dev": true
+      "integrity": "sha1-IfibcJuvxLMDra56g7TzWg2eR5Y="
     },
     "@types/estree": {
       "version": "0.0.39",
--- a/package-lock.json
+++ b/package-lock.json
@@ -5625,8 +5625,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -5647,14 +5646,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
@@ -5669,20 +5666,17 @@
         "code-point-at": {
           "version": "1.1.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "concat-map": {
           "version": "0.0.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "console-control-strings": {
           "version": "1.1.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "core-util-is": {
           "version": "1.0.2",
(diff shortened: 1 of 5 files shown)


nicolodavis:
Given that the 3D stuff is still experimental, I'm leaning toward just requiring users to install three separately (just like React isn't a dependency).

We can move tween to the deps though.

flamecoals:
> Given that the 3D stuff is still experimental, I'm leaning toward just requiring users to install three separately (just like React isn't a dependency).
> 
> We can move tween to the deps though.

I am not using any 3D stuff in my app though, and it still require three installed. If we do not add three as a dep, the UI import will break.

nicolodavis:
Then we should probably change the `import` to a `require` similar to how we handle Firebase or Mongo (we don't pull in those deps automatically).

flamecoals:
Agreed :). I can try doing that later.

nicolodavis:
BTW, the integration test is broken due to https://github.com/webpack-contrib/terser-webpack-plugin/issues/66.

flamecoals:
OK, I changed to require() and removed it from deps (added it back to dev deps...). This should work now :)

flamecoals:
Gosh, I still get the error using require (only when running webpack, not the tests):
```
ERROR in ./node_modules/flamecoals-boardgame.io/dist/ui.js
Module not found: Error: Can't resolve 'three' in '/home/felizardow/playground/FreeBoardGame.org/node_modules/flamecoals-boardgame.io/dist'
 @ ./node_modules/flamecoals-boardgame.io/dist/ui.js 300:20-36 868:20-36 1098:20-36
 @ ./node_modules/flamecoals-boardgame.io/ui.js
 @ ./src/games/chess/board.tsx
 @ ./src/games/index.tsx
 @ ./src/App/Game/GameInfo.tsx
 @ ./src/App/App.tsx
 @ ./src/app.tsx
```

Line 868 is:
```
      _this.originalY = props.thickness / 2 - 0.0001;
      _this.three = require('three');
      var geometry = new _this.three.BoxGeometry(props.width, props.thickness, props.height);
```

flamecoals:
@nicolodavis any idea of solution for this issue? It seems like webpack is calling the constructor of the 3D UI even though I only use the 2D UI...

nicolodavis:
Hmm, hard to tell without looking at your specific webpack config.  Are you using an option that is causing it to try to resolve `require` statements as well?

flamecoals:
I am using the latest webpack and it does try to resolve requires. Thats probably why indeed. I'm going to close this

nicolodavis:
I see. Do you have a workaround?

flamecoals:
No, I just installed it. The other solution would be to separate the 3d and
the 2d modules, but I think that's just too much work.

PR #363, opened 2019-02-12
Removes undefined values before state is persisted to Firebase

### LINKED ISSUE #362: Error when persisting deltalog to Firebase RTDB
I haven't quite pinpointed where this is happening, but it seems that where ever `deltalog` is being prepared for storage it has `credentials` removed by setting them to `undefined`. Firebase does not accept `undefined` values, preferring `null` instead.

```
Error: Reference.set failed: First argument contains undefined in property 'GAME:ROOM.deltalog.1.action.payload.credentials'
```

Is the `credentials` field meant to be stored with an empty value or removed altogether? If the latter, I'd be happy to submit a PR removing all `undefined` values before saving to Firebase.

jorbascrumps:
#### Checklist

Resolves #362


src/server/db/firebase.js
```diff
@@ -64,14 +64,17 @@ export class Firebase {
       return;
     }
 
-    this.cache.set(id, state);
+    // Remove any undefined values (ie, credentials)
```
nicolodavis: Are you sure that you're seeing `credentials` here?  That shouldn't even be a part of the state.

It's more likely that you're seeing `undefined` values for `deltalog` and `log`.  These are set in `src/master/master.js`, and can probably just be trimmed away there:

```js
function removeUndefinedFields(state) {
  const { log, deltalog, ...rest } = state;
  return rest;
}
```

iamchristopher: You're right about `deltalog`. The original error I was seeing was for `undefined` credentials in `deltalog`.

I can update to your suggested fix. Is there anywhere else this may be an issue? Also, it will still be possible for someone to accidentally include an `undefined` and receive the same error but I guess at that point they can address it in their own codebase.

nicolodavis: I think that one location should do it.

nicolodavis:
@jorbascrumps I forgot if I'm waiting for you or if you're waiting for me on this one.

nicolodavis:
Closing this due to lack of activity.

Issue #368, opened 2019-02-18
PluginPlayer doesn't issue changes for current player


MathieuLoutre:
I have this piece of code:

```js
        onPhaseEnd: (G, ctx) => {
          Object.values(G.players).forEach((player) => {
            G.decks.small = [...G.decks.small, ...player.tempHand]
            player.tempHand = []
          })

          G.decks.small = ctx.random.Shuffle(G.decks.small)

          ctx.events.endTurn()
        },
```

Inside the next phase `onPhaseBegin`, I'm checking is `tempHand` is empty but it turns out that for the current player at the time it was reset, it's not.

```Object.values(G.players).forEach((player) => console.log(player.tempHand.length))```

This logs: `0 0 2` instead of `0 0 0` as expected. Inside `onPhaseEnd`, the current player (available on `G.player`) is still the last player (see https://github.com/nicolodavis/boardgame.io/issues/367#issuecomment-464843807) and accessing it through the `players` object doesn't seem to affect it. 

However, changing the value in `G.player` directly (`G.player.tempHand = []`) works. It would be great if the `players` object could reflect the changes on the current player too.

Edit: This is actually the same everywhere I'm iterating over the `players` object to make changes to every player.

nicolodavis:
The way `PluginPlayer` works right now, it just copies over `G.players[currentPlayer]` to `G.player` before the turn, and copies `G.player` back to `G.players[currentPlayer]` at the end of the turn.  It's a weak attempt at trying to create an alias (which explains the problem that you're running into).

Open to other suggestions on how we can make this better.

MathieuLoutre:
My solution so far is to remove the `player` reference and just use `const player = G.players[ctx.currentPlayer]` at the start of each action. It feels a bit like boilerplate but it’s not that bad. Creating some sort of getter that would preserve the reference might not even help that much? Maybe wrapping the move with a 3rd argument `(G, ctx, player)` when the plugin is active?

nicolodavis:
I think a [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) / setter combination might work.  We can convert `player` and `opponent` to getters / setters so that they retrieve the underlying data on demand.

MathieuLoutre:
[...]

PR #369, opened 2019-02-18
Add test case for modifying all players

MathieuLoutre:
Extend 3 player test case with a new move that changes a field for all players but iterating over the players object provided by the Players plugin.

This is currently a failing test, see #368

### DIFF
--- a/src/plugins/plugin-player.js
+++ b/src/plugins/plugin-player.js
@@ -17,33 +17,36 @@ export default {
   fnWrap: moveFn => {
     return (G, ctx, ...args) => {
       const current = ctx.currentPlayer;
-      const player = G.players[current];
 
-      G = { ...G, player };
+      Object.defineProperty(G, 'player', {
+        get: () => {
+          console.log('getter');
+          return G.players[current];
+        },
 
-      let other = null;
-      let opponent = null;
+        set: function(value) {
+          console.log('setter');
+          this.players[current] = value;
+        },
+      });
+
       if (ctx.numPlayers == 2) {
-        other = current == '0' ? '1' : '0';
-        opponent = G.players[other];
-        G.opponent = opponent;
+        const other = current == '0' ? '1' : '0';
+
+        Object.defineProperty(G, 'opponent', {
+          get: () => G.players[other],
+          set: value => {
+            this.players[other] = value;
+          },
+        });
       }
 
       G = moveFn(G, ctx, ...args);
 
-      const players = {
-        ...G.players,
-        [current]: G.player,
-      };
-
-      if (other !== null) {
-        players[other] = G.opponent;
-      }
-
       {
         /* eslint-disable-next-line no-unused-vars */
         const { player, opponent, ...rest } = G;
-        return { ...rest, players };
+        return { ...rest };
       }
     };
   },
--- a/src/plugins/plugin-player.test.js
+++ b/src/plugins/plugin-player.test.js
@@ -82,7 +82,7 @@ describe('3 player game', () => {
     game = Game({
       moves: {
         A: G => {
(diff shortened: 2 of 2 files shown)


MathieuLoutre:
CLA signed

MathieuLoutre:
I've fiddled around with the plugin to see if I could get the test to pass but because of immutability, the data can only propagate on way. I've also had problems with the cleanup (`players` was empty after `INVALID_MOVE` is returned and sometimes `G.player` would be undefined in some of the hooks) so I made a custom version which doesn't set `player` at all, just has the `setup` phase. It's a little more verbose but at least it's consistent.

nicolodavis:
Playing around with getters / setters in https://github.com/nicolodavis/boardgame.io/pull/369/commits/784d95c894cd4010b4ea3916b56b6b9006e88e92.  Haven't got it working yet, though.

nicolodavis:
@MathieuLoutre Haven't had much time to work on this PR.  Feel free to pick it up if you have the time.

MathieuLoutre:
So I've played with the getter/setter approach a bit and I can't make it work. The getter works great, but setting a property using the getter (`G.player.field = 'A'`) does not work. There are not errors, etc. but the value doesn't update.

There's nothing I could find in the docs that would suggest updating an object through a getter wouldn't be possible. Could it be that `immer` doesn't like the getter getting in the way?
From what I could gather from [this issue](https://github.com/mweststrate/immer/issues/317#issuecomment-463982274), it looks like `immer` might disregard the getters. I'm not too familiar with `immer` so it's hard for me to say.

Any other leads?

nicolodavis:
Hmm, yeah it's possible that there is a bad interaction with immer going on. Immer uses ES6 proxies under the hood.

Taking a step back, do you think it's reasonable to just make the `players` object an internal field that's not meant to be accessed directly if this plugin is used?

MathieuLoutre:
The use case I have to have both `player` and `players` accessible is because the game I'm working on has cards with different areas of effect. Some cards affect the current player and some others affect all players.

In the first case, I'm just changing the `player` state (removing the card from the hand, applying the effect to the player), but in the second one, after removing the card from the current player's hand, it's easier to loop over `players` and apply the effect systematically.

In my implementation of the "Player Plugin", I've omitted `player` because that can be declared easily at the top of the move with `const player = G.players[ctx.currentPlayer]`. It's a bit boilerplate-y but I found that more convenient than not being able to loop over `players` or having to do all the actions on `players` minus `player` and then on `player` alone to cover everyone.

So if `players` was just an internal field, what would be the approach to access the other players?

nicolodavis:
One way could be to make `player`, `opponent`, `all` and `others` objects with getters / setter functions until we figure out how to make them work with ES6 getters / setters:

```
const value = G.player.get();
const value = G.opponent.get();

G.player.set(value);
G.opponent.set(value);
G.all.set(value);
G.others.set(value).
```

All of these manipulate the `G.players` object, which can also be manipulated directly.

MathieuLoutre:
I quite like the idea of the `all` and `others` sugar (both I use often in this game) but I don't know how I feel about the syntax. Do you imagine `set` to be working with updates described as objects or functions when it's for multiple players? (e.g. `{ money: 10 }` and all players get assigned 10 money but keep the rest of their state, or `(p) => p.money += 1` to increase every player's money). Or do you picture getting an array/object of all players, mutating it and then setting it back?

nicolodavis:
I was thinking that set just sets an object that's attached to a particular player (the entire object, as opposed to updates).  We can also try to do updates (a la React's `setState`).

nicolodavis:
This is now obsolete after PluginPlayer has changed to use the new Plugin API.  Closes #368.

Issue #384, opened 2019-03-30
FR: Support firebase-admin storage connector


p00ya:
The current Firebase implementation from https://github.com/nicolodavis/boardgame.io/blob/master/src/server/db/firebase.js uses the plain 'firebase' client:

```
  this.client = require('firebase');
  // ... then later:
  this.client.initializeApp(this.config);
    this.db =
      this.engine === ENGINE_FIRESTORE
        ? this.client.firestore()
        : this.client.database().ref();
```

The documented way to access a Firestore database from the server seems to be via the Firebase Admin SDK.  From [the Firebase docs](https://firebase.google.com/docs/web/setup):

> If you are interested in using Node.js in privileged environments such as servers or serverless backends like Cloud Functions (as opposed to clients for end-user access like a Node.js desktop or IoT device), you should instead follow the instructions for setting up the Admin SDK.

The [Admin SDK docs](https://firebase.google.com/docs/admin/setup) shows that this can be used similarly to the firestore module, i.e.

```
import * as admin from 'firebase-admin';
admin.initializeApp({...config});
const db = admin.firestore();
```

I've tested that it all works with something like:

```
import firebase from 'firebase-admin';
import { Firebase, Server } from 'boardgame.io/server';

const db = new Firebase({ config });

// Replace the boardgame.io client with an admin client, in order to use
// service accounts.
db.client = firebase;

const server = Server({
  db,
  games: [Game],
});
```

So I think we should at least support using `firebase-admin` instead of the plain `firebase` client as an opt-in.  Happy to send a PR for an extra property in the Firebase ctor to enable this, similar to the "engine" property if it SGTY.

However, all this begs the question: what are the current users of the Firebase connector doing with their [Firestore security](https://firebase.google.com/docs/firestore/security/overview)? Given they're not using `firebase-admin`, they're not using a service role account, so either they've got open read/write ACLs on their Firebase database, or they're somehow pushing EUCs through their server.  Or maybe I'm completely missing something? @jorbascrumps who seems to be an active user

nicolodavis:
Happy to approve a PR adding an option to use `firebase-admin`. We'll also need to update the docs as well.

PR #386, opened 2019-04-04
Add adminClient option for Firebase storage

p00ya:
I expect this to be backward-compatible with all existing users.  The
only gotcha might be people blacklisting dynamic "require"s for webpack;
they'll need to also blacklist 'firebase-admin'.

Add documentation for the new and existing properties to the tutorial.
Also update the example to use Firestore and adminClient.

Fixes nicolodavis/boardgame.io#384

#### Checklist


package.json
```diff
@@ -93,6 +93,7 @@
     "eslint-plugin-react": "^7.11.1",
     "eslint-plugin-unicorn": "^6.0.1",
     "firebase": "^5.7.3",
```
nicolodavis: Does `firebase-admin` supersede this, or do we still need plain `firebase`?  Asking because I noticed that you removed it from the documentation at storage.md.

p00ya: Not really.  `firebase` is the Web SDK and `firebase-admin` is the Admin SDK - they're independent (but sufficiently similar APIs).  I've only really retained the support for `firebase` in case there are existing users.  The tests exercise both, and the documentation describes both (although the example uses `firebase-admin` which I'm pretty sure is the better option).

p00ya:
> Can you remove the `package-lock.json` diff from this PR? I usually just generate them myself before merging.

Done.  It seems to make the CI sad though. :'(

docs/storage.md
```diff
@@ -35,35 +35,62 @@ server.run(8000);
 
 ### Firebase
 
-First, install the necessary packages:
+First, install the Firebase Admin Node.js SDK:
 
 ```
-npm install --save firebase
+npm install --save firebase-admin
```
nicolodavis: Just to double-check, users don't have to also install `firebase` (i.e. `npm install --save firebase firebase-admin`)?

p00ya: In practical terms, no: I expect most users will install `firebase-admin`, and set `adminClient` to `true` for their one `Firebase` instantiation, which will not require installing `firebase`.  If they set `adminClient` as `false`, they'll need to install `firebase` (but again not both).  If they create two Firebase objects with both `adminClient` states, they'll probably figure out they need to install both. ;)

nicolodavis: OK, thanks for the clarification!

PR #393, opened 2019-05-07
update _undo to reflect current ctx

mgrau:
add tests for undo and redo of actions that change phase.

#### Checklist

* [x ] Use a separate branch in your local repo (not `master`).
* [x ] Test coverage is 100% (or you have a story for why it's ok).

### DIFF
--- a/src/core/reducer.test.js
+++ b/src/core/reducer.test.js
@@ -320,6 +320,75 @@ test('undo / redo', () => {
   expect(state.G).toEqual({ A: true });
 });
 
+test('undo/redo phase', () => {
+  let game = Game({
+    moves: {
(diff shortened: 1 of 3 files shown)


src/core/reducer.test.js
```diff
@@ -341,6 +341,75 @@ test('undo / redo', () => {
   expect(state.G).toEqual({ A: true });
 });
 
+test('undo/redo phase', () => {
```
nicolodavis: Shall we group this test with the one above inside a `describe()`?

mgrau: Yes. I've grouped them together, and also simplified the test.

PR #395, opened 2019-05-11
Fixes some examples if you are running from other machine not localhost

flamecoals:
#### Checklist


nicolodavis:
Thanks for the fix 🙂

Issue #354, opened 2019-01-29
Generic Lobby


nicolodavis:
Ideas for improvement:

[ ] Poll for changes to state so that refreshes are not required.
[ ] Option to automatically remove room if all users leave the room
[ ] Add optional unique name check
[ ] Add optional limit for rooms (will disable 'create room' button if rooms number reached limit)
[ ] Provide ability to customize ui when waiting for an API call (i.e. loading state)
[ ] Rename `gameInstances` to `rooms` (easier to understand).

flamecoals:
Why don't we use websockets instead of polling?

blunket:
IMO polling is better for this, sockets are overkill when it's just a lobby and doesn't need to be real time

lehaSVV2009:
btw, I also thought that lobby API might be improved a bit as well. To smth like that:
`GET /games
GET /games/my-game/rooms
POST /games/my-game/rooms
POST /games/my-game/rooms/123/join
POST /games/my-game/rooms/123/leave
DELETE /games/my-game/rooms/123`

flamecoals:
One use case I want to bring to everybody's attention... My project will have several games, and I want people to choose the game after creating the lobby. And after a game is done, I want people to be able to play another game with the same people.

Is it ok if we key only by lobby id instead of both game and id? I can work on this latter

blunket:
@flamecoals What do you mean? Multiple lobbies? Couldn't you just filter which games to display however you like? Or do you mean the games will be hosted on many different game servers too, or something like that?

Right now players choose the game to play IN the lobby, they don't create lobbies.

nicolodavis:
Yes, I'm envisioning just one global lobby, and you can create games inside that lobby (each game can be a different kind). Players then join the game you created in order to play.

flamecoals:
@blunket I meant one global lobby like Nicolo said. Basically a "room" persisting across matches too, of different types of games.

Use case:
- We and my friend Joe want to play some games, but we dont know exactly what
- We create a lobby for us, and we agree in playing Chess inside the lobby
- After the chess match is done, we go back to the same lobby and agree in playing tic tac toe

and so on

blunket:
I see, I think you mean sort of like a party system? That'd be cool.

flamecoals:
Hey folks, I started a document for documenting and designing all these ideas we were talking about:
[...]

PR #396, opened 2019-05-12
Creating lobby API config and making the UUID customizable

flamecoals:
Improving lobby, as discussed in #354 ...
We need this change to make it possible to use the Lobby API in [FreeBoardGame.org](https://freeboardgame.org). In FBG users send a link to invite friends to a match, and the current UUID is too long and too verbose for a link.

#### Checklist

### DIFF
--- a/src/server/api.js
+++ b/src/server/api.js
@@ -9,7 +9,7 @@
 const Koa = require('koa');
 const Router = require('koa-router');
 const koaBody = require('koa-body');
-const uuid = require('uuid/v4');
+const shortid = require('shortid').generate;
 const cors = require('@koa/cors');
 
 import { InitializeGame } from '../core/reducer';
@@ -32,6 +32,7 @@ const GameMetadataKey = gameID => `${gameID}:metadata`;
  * @param {object} setupData - User-defined object that's available
  *                             during game setup.
  * @param {object } serverConfig - Configuration options for the server.
+ * @param {object } lobbyConfig - Configuration options for the lobby.
  */
 export const CreateGame = async (
   db,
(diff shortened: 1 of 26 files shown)


nicolodavis:
@flamecoals Do you think it makes sense to just make the default a shorter uuid that is suitable for your purpose?

This PR is fine too, but just curious.

flamecoals:
I think it does make sense, yes, I think making it easy to share a URL could be a built in feature on the framework.

flamecoals:
I can change the default library to be [shortid](https://www.npmjs.com/package/shortid) ... That's the one we currently use. Do you want me to do that, @nicolodavis ?

nicolodavis:
Yeah, let's go with shortid as the default.

src/server/api.js
```diff
@@ -57,12 +63,15 @@ export const CreateGame = async (db, game, numPlayers, setupData) => {
...
+export const createApiServer = ({ db, games, lobbyConfig }) => {
   const app = new Koa();
-  return addApiToServer({ app, db, games });
+  return addApiToServer({ app, db, games, lobbyConfig });
 };
 
-export const addApiToServer = ({ app, db, games }) => {
+export const addApiToServer = ({ app, db, games, lobbyConfig }) => {
+  if (!lobbyConfig || !lobbyConfig.shortid) {
+    lobbyConfig = { ...lobbyConfig, shortid };
```
nicolodavis: If lobbyConfig is `undefined` or `null`, then wouldn't this object-spread cause an error?  I think you want something like:

```
if (!lobbyConfig) {
  lobbyConfig = {};
}

if (!lobbyConfig.shortid) {
  lobbyConfig = { ...lobbyConfig, shortid };
}
```

src/server/api.js
```diff
@@ -44,11 +50,11 @@ export const CreateGame = async (db, game, numPlayers, setupData) => {
   });
 
   for (let playerIndex = 0; playerIndex < numPlayers; playerIndex++) {
-    const credentials = createCredentials();
+    const credentials = lobbyConfig.shortid();
```
nicolodavis: Shall we call this option something more generic like `uuid`?  If someone overrides it to generate a longer ID, then `shortid` won't be a particularly relevant name.

src/server/api.js
```diff
@@ -33,8 +31,16 @@ const GameMetadataKey = gameID => `${gameID}:metadata`;
  * @param {number} numPlayers - The number of players.
  * @param {object} setupData - User-defined object that's available
  *                             during game setup.
+ * @param {object } serverConfig - Configuration options for the server.
```
nicolodavis: Looks like this can be removed.

Issue #345, opened 2019-01-21
Player can undo during other player's turn


blunket:
I have implemented some basic undo/redo buttons. However, Player 1's client is able to call undo during Player 2's turn, which is undoing player 2's moves.

blunket:
Along with this issue, I want to add: it'd be pretty cool if the `ctx` var had a built-in way to check if undo/redo is even available. (right now I can just do that manually though)

nicolodavis:
Thanks for the report! Will take a look.

blunket:
Sounds good to me! I don't know how it would work in a game with multiple people able to move in the same turn -- didn't know that was possible. Wouldn't they just split it into different moves and switch the turn over to different players as necessary?

In any case, I think that'll probably be a fairly rare situation and so that will be a good solution for now. Maybe someone can raise an issue again if needed.

Is there any possibility to add undo and redo to server side code without breaking changes? Or is that too unnecessary for now?

nicolodavis:
What do you mean by undo / redo in server side code?

At the moment an Undo / Redo is triggered on the client (and it is verified by the server before the Undo is broadcast to other players).

PR #400, opened 2019-05-15
Remove some secrets from client in multiplayer game

### LINKED ISSUE #399: `PlayerView.STRIP_SECRETS` does not remove secrets from client `ctx._undo` and `ctx._initial`.
`PlayerView.STRIP_SECRETS` removes private information from the client's copy of `G` - but the information is still passed to the client's `ctx` through `_undo` and `_initial`.

qsona:
Fix #399 .

My current understanding: In the multiplayer game, undo and redo also must be handled by server, but there is no implementation now (in src/master/master.js) .

I explicitly disabled these features to fix #399 for now. I think we can implement multiplayer undo/redo,  but its priority doesn't seems so high... does it?

#### Checklist


qsona:
@nicolodavis Is this change acceptable? If it's okay I'll add the test cases.

src/master/master.js
```diff
@@ -204,6 +204,9 @@ export class Master {
         ctx: { ...state.ctx, _random: undefined },
         log: undefined,
         deltalog: undefined,
+        _undo: undefined,
+        _redo: undefined,
+        _initial: undefined,
```
nicolodavis: Just curious, why does `_initial` also get affected here?

nicolodavis: I think this might break the log, which uses `_initial`.

nicolodavis: We should instead remove `_undo` from inside it.

src/core/reducer.js
```diff
@@ -322,6 +322,10 @@ export function CreateGameReducer({ game, multiplayer }) {
       }
 
       case Actions.UNDO: {
+        if (multiplayer) {
```
nicolodavis: I think we can allow these two actions, as long as we strip `_undo` and `_redo` like you are in `master.js`.  These fields are only needed by the current player (and do not need to be sent over the network).

qsona:
@nicolodavis Thank you for reviewing! I think I understood. I made changes and rebased on master branch.

nicolodavis:
Made some small changes and added an interactive example.

nicolodavis:
Thanks for the fix, @qsona!

I think there are still a couple of issues with the undo / redo mechanism that are not ideal:

1.  Right now any player can call undo at any time (you can even undo another player's move).  We should restrict this to only `ctx.currentPlayer` (https://github.com/nicolodavis/boardgame.io/issues/345).

2.  After this PR, all the undo logic is handled on the server for a multiplayer game.  The client merely sends an undo message to the server and has the wait for the response before its move is undone.  This can be laggy in case the network is slow.  We should instead not override the local copy of _undo during an UPDATE.  That way the client can undo its move immediately and the server later just sends an update that enables other players to see that a move has been undone.

Something like this in `reducer.js`:

```
+      case Actions.UPDATE: {
+        // Override local state, but don't touch _undo or _redo.
+        if (state !== null) {
+          const { _undo, _redo } = state;
+          return { ...action.state, _undo, _redo };
+        }
+        return action.state;
+      }
```

Let me know if you'd be interested in working on either of these.

qsona:
@nicolodavis Thank you for describing the remaining issues. I'm interested in solving them. For 1, I've sent a PR #408 . For 2, I'd like to work on this issue if sometime I have enough time, but actually it's hard to promise (a bit busy for a while).

nicolodavis:
Thanks @qsona!  Will take a look at your PR.

Don't worry about the other one if you don't have time.  I'll probably get down to it eventually.

PR #408, opened 2019-05-26
Fix #345; restrict undo/redo to currentPlayer

### LINKED ISSUE #345: Player can undo during other player's turn

qsona:
I tried to fix #345 (but I'm not very confident that this PR correctly fixes it. Please give me any feedback.)

related to https://github.com/nicolodavis/boardgame.io/pull/400#issuecomment-495854120

#### Checklist

### DIFF
--- a/src/core/action-creators.js
+++ b/src/core/action-creators.js
@@ -84,14 +84,18 @@ export const reset = state => ({
 
 /**
  * Used to undo the last move.
+ * @param {string}  playerID - The ID of the player making this action.
  */
-export const undo = () => ({
+export const undo = (playerID) => ({
   type: Actions.UNDO,
+  payload: { playerID },
 });
 
 /**
  * Used to redo the last undone move.
+ * @param {string}  playerID - The ID of the player making this action.
  */
-export const redo = () => ({
+export const redo = (playerID) => ({
   type: Actions.REDO,
+  payload: { playerID },
 });
--- a/src/core/action-creators.js
+++ b/src/core/action-creators.js
@@ -84,20 +84,14 @@ export const reset = state => ({
 
 /**
  * Used to undo the last move.
- * @param {string}  playerID - The ID of the player making this action.
- * @param {string}  credentials - (optional) The credentials for the player making this action.
  */
-export const undo = (playerID, credentials) => ({
+export const undo = () => ({
   type: Actions.UNDO,
-  payload: { playerID, credentials },
 });
 
 /**
  * Used to redo the last undone move.
- * @param {string}  playerID - The ID of the player making this action.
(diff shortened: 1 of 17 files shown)


src/core/reducer.js
```diff
@@ -336,6 +336,10 @@ export function CreateGameReducer({ game, multiplayer }) {
           return state;
         }
 
+        if (action.payload.playerID && action.payload.playerID !== state.ctx.currentPlayer) {
```
qsona: This `action.payload.playerID && ` guard clause seems unnecessary because playerID always exists. I prefer to delete this clause by regarding `playerID` as required parameter, but it needs fixing some existing tests, so I added it for now and would like to listen your opinion.

nicolodavis: Sure, let's assume that it is always sent and fix the tests instead.

src/core/reducer.test.js
```diff
@@ -387,6 +387,47 @@ test('undo / redo', () => {
   expect(state.G).toEqual({ A: true });
 });
 
+
+test('undo / redo on 2 players game', () => {
```
nicolodavis: Can you group this test and the undo test right above this under one `describe`?

src/core/action-creators.js
```diff
@@ -84,14 +84,18 @@ export const reset = state => ({
 
 /**
  * Used to undo the last move.
+ * @param {string}  playerID - The ID of the player making this action.
  */
-export const undo = () => ({
+export const undo = (playerID) => ({
```
nicolodavis: We should also pass in `credentials` (similar to `makeMove` and `gameEvent`).

qsona:
@nicolodavis Thank you for reviewing this. I fixed them.
Currently undo test in src/client/react-native.test.js is failing. I found that `playerID` in the payload is `null` at reducer, but I haven't understand very much around this code so it's a bit hard to fix it..

nicolodavis:
Ah, that's because in the singleplayer mode the client doesn't send a `playerID` as part of the actions.

nicolodavis:
I'll take a look and see if I can either:

1. Make the singleplayer start sending `playerID`.
OR
2. Fix the tests based on the assumption that it will be `null` sometimes.

qsona:
@nicolodavis Thank you for fixing this! Sorry for my unfinished work 🙇

nicolodavis:
No problem! After looking at it some more, I realized that it's simpler to guard the undo / redo logic in `master.js` instead, so simplified the code a bit.

PR #411, opened 2019-05-28
ref(lobby): Use hooks for lobby [WIP]

billyvg:
This is just a WIP atm, wanted to get feedback before I got this working.

I'm proposing a refactor of lobby to use React hooks (this would require a new minimum version of React, I think 16.8, however).

This would allow better encapsulation of lobby data. You could then expose the hook for consumers downstream as well as provide a Lobby component like you have currently. This would let users that are using 16.8, but have not adopted hooks, to still use it with the render function API.

Thoughts?

#### Checklist


nicolodavis:
@billyvg I like the direction where this is going!  If there are no objections from the the others, I'm in favor of going ahead with this PR.

Do you think it makes sense to also deprecate the props based custom renderer?  I think it's reasonable to ask the end user to use the provided component or use the hook if they want their own custom component.  We could also allow some CSS customizations on the provided component if they just want some cosmetics changes.

billyvg:
@nicolodavis Sounds good!

billyvg:
@nicolodavis 	was playing around with this and having troubles getting the examples to work because of: https://reactjs.org/warnings/invalid-hook-call-warning.html#mismatching-versions-of-react-and-react-dom

Any ideas here? I tried making react a peer dep, but there are dependencies in the core lib that require react (e.g. `react-dragtastic`). I'm not sure of any solutions that don't involve a mono repo.

nicolodavis:
Couple of things:
- `react-dragtastic` is probably going away in the future (I am planning to rework that code).
- React itself isn't an actual direct dependency of the project ([source](https://github.com/nicolodavis/boardgame.io/blob/master/package.json#L130)).  Does moving `react-dragtastic` to peer deps fix the problem?

nicolodavis:
I'm going to close this in favor of our current plan to move all the UI components to Svelte #432

PR #417, opened 2019-06-10
custom client transport should be a constructor (IMO)

amitport:
Hi,

I'm implementing a custom transport. Having a parameter that just override transport is not very useful as the transport will need access to things like the store (same way that local and socketio transport do).

#### Checklist

* [V] Use a separate branch in your local repo (not `master`).
* [V] Test coverage is 100% (or you have a story for why it's ok).

### DIFF
--- a/src/client/client.js
+++ b/src/client/client.js
@@ -273,7 +273,14 @@ class _ClientImpl {
           socketOpts,
         });
       } else if (multiplayer.transport !== undefined) {
-        this.transport = multiplayer.transport;
+        this.transport = new multiplayer.transport({
+          store: this.store,
+          gameID: gameID,
+          playerID: playerID,
+          gameName: game.name,
+          numPlayers,
+          socketOpts,
+        });
       } else {
         error('invalid multiplayer spec');
       }
--- a/src/client/client.test.js
+++ b/src/client/client.test.js
@@ -194,18 +194,22 @@ describe('multiplayer', () => {
   });
 
   describe('custom transport', () => {
-    const transport = { custom: true };
+    class CustomTransport {
+      custom = true;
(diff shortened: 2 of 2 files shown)


nicolodavis:
Just curious, what transport are you implementing?

amitport:
few things:

1. higher level transport a-la https://github.com/primus/primus
2. mirror-like transport for integeration with my own https://github.com/CardForest/master-class (which is like a proxy based state-machine master that is transperantly local or remote :) (also used instead of redux))

nicolodavis:
Sounds good.

Can you update the test in client.test.js that is now failing?

amitport:
@nicolodavis 
changed so transport is now nested under `multiplayer.server`. Makes sense to me since transport is a property of the server config. The entire `multiplayer.server` is passed to the transport and may contain more custom configuration.

BTW I think there is room for bigger refactor here (I wrote to you in in gitter: 

may want to lift entire multiplayer/transport selection logic outside of `_ClientImpl`, enable cleaner config and smaller bundles for those not using socketIO

)

src/client/client.js
```diff
@@ -272,8 +272,6 @@ class _ClientImpl {
           server: multiplayer.server,
           socketOpts,
         });
-      } else if (multiplayer.transport !== undefined) {
-        this.transport = multiplayer.transport;
```
nicolodavis: Shall we just leave this option as `multiplayer.transport` and treat it as a constructor here?  I'd prefer to use `server` for the socketio stuff.

Basically I'm proposing changing this line to `this.transport = new (multiplayer.transport)({ ... })` and pass in everything that's passed in from the if branch right above.

PR #423, opened 2019-06-21
WIP: Improve handling of `ctx._actionPlayersOnce` across phases

### LINKED ISSUE #422: `ctx._actionPlayersOnce` is not reset at the beginning of a new phase
Thanks for this library. I’ve been having fun getting to know it.

When starting a new phase, if a previous phase set `ctx._actionPlayersOnce` to `true`, e.g. by using `TurnOrder.ANY_ONCE`, this is not reset to `false` by `InitTurnOrderState`:

https://github.com/nicolodavis/boardgame.io/blob/243ffef6953cc39a9414fc08e10a6dd894502501/src/core/turn-order.js#L92-L108

Setting `ctx._actionPlayersOnce` back to `false`, requires passing in an `actionPlayers` field, e.g.:

```js
turnOrder: {
  actionPlayers: {
    once: false
  }
}
```

But doing that overrides the default action players value, setting it to `[]`.

Is this intentional? The behaviours I would expect are for new phases to reset to `false` unless otherwise specified and for an `actionPlayers` field containing none of `value`, `all`, or `others` to maintain the default value of `[ currentPlayer ]`.

Happy to try a PR if this seems like a bug.

delucis:
Fixes #422 

So far, this PR just makes sure that when initialising a phase without a `turnOrder.actionPlayers` field, `ctx._actionPlayersOnce` is reset to `false`.

The second behaviour that might be improved (see #422) is that a turn order with `actionPlayers: { once: false }` results in an empty `ctx.actionPlayers`, where one might expect it to be `[ currentPlayer ]`. To achieve the behaviour I expect (or to add a `current` option to `setActionPlayers` similar to the `all` or `others` options) would require `setActionPlayers` being aware of the new `currentPlayer` currently returned by `getCurrentPlayer` inside `InitTurnOrderState`.

Could we pass `currentPlayer` to `setActionPlayers` or does `setActionPlayers` get called from other places and need to deduce `currentPlayer` internally?

#### Checklist

#### Edit

After digging around trying to understand phases/turns better, I realise you’re currently working on a complete API overhaul in the `phases-overhaul` branch, so perhaps these issues may no longer be relevant once that is merged.


delucis:
This PR should probably also take care of `UpdateTurnOrderState`. `actionPlayers` are currently only updated if a user didn’t pass an `actionPlayers` field in their turn order or `endTurn` was called with a next player argument.

https://github.com/nicolodavis/boardgame.io/blob/243ffef6953cc39a9414fc08e10a6dd894502501/src/core/turn-order.js#L141-L143

Perhaps the `actionPlayers` field should offer an `update` option so users can provide an update function similar to the `value` function?

nicolodavis:
Thanks for the PR!  I'm traveling until next week, so will take a look once I'm back.

nicolodavis:
You're also right that this code will likely be going away with the API overhaul. However, if it's a small fix I'm happy to get it in before that's out as well.

delucis:
The tiny change in the PR so far was enough for me to build workarounds for the other issues. The other things likely need a bit more work and attention so may not be worth it if they're just overwritten after a short time.

Excited to try the new API when it's ready.

Safe travels!

PR #424, opened 2019-06-22
server api: fix test of argument 'playerID' 

francoijs:
In the lobby example, playerID=0 was being considered 'undefined' in request 'leave', leading to error 403

#### Checklist


nicolodavis:
Thanks @francoijs. Looks like I got another PR with the same fix, so merged that instead.

PR #425, opened 2019-06-24
fix: Fix join/leave a room when playerID is 0

### LINKED ISSUE #420: Player.id=0 cannot leave game
When the first player (player.id=0) tries to use the /leave endpoint, they get a 403 error because of the following check in the /server/api.js:

`if (!playerID) {
      ctx.throw(403, 'playerID is required');
    }`

The check should instead be

`if (typeof playerID === 'undefined') {`

This should also be changed for the /join endpoint.

billyvg:
Fixes joining/leaving a room for playerID = 0.

Closes #420

#### Checklist

### DIFF
--- a/src/server/api.js
+++ b/src/server/api.js
@@ -149,7 +149,7 @@ export const addApiToServer = ({ app, db, games, lobbyConfig }) => {
   router.post('/games/:name/:id/join', koaBody(), async ctx => {
     const playerID = ctx.request.body.playerID;
     const playerName = ctx.request.body.playerName;
-    if (!playerID) {
+    if (typeof playerID === 'undefined' || playerID === null) {
       ctx.throw(403, 'playerID is required');
     }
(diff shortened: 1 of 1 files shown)


flamecoals:
+1 for merge, this thing broke FreeBoardGame.org when we upgraded to HEAD

flamecoals:
@nicolodavis Can we please merge this quickly? This is breaking the lobby for everybody. 

#425 another duplicate PR fixing the same issue

flamecoals:
It seems like @jasonharrison forgot to push a commit with that before merging the PR.
We should probably add some regression test to avoid this happening agian

nicolodavis:
Sorry about the delay.  I'm still out on vacation.  Will be back next week.

nicolodavis:
I've merged it in.  Will make a release once I'm back.

PR #427, opened 2019-06-25
Add search plugin to docs

### LINKED ISSUE #415: Text Search for docs page
Looks like docsify now has a plugin for text search, so we should consider using it.

Link: https://docsify.js.org/#/plugins?id=full-text-search

chrisheninger:
Adds text search plugin to docsify– closes #415.

boardgame-io-docs-search

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -42,6 +42,7 @@
     loadSidebar: true,
     subMaxLevel: 2,
     auto2top: true,
+    search: 'auto', // default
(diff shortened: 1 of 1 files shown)


nicolodavis:
Chris, this does seem to slow down the initial page load of the docs though.

Any ideas on how to speed it up? In the worst case, maybe I'll create a separate landing page that's fast and link to the docs from there.

PR #430, opened 2019-07-09
No payload is not an authentic player

### LINKED ISSUE #410: isActionFromAuthenticPlayer allows moves without credentials if payload is missing
If the `action` has no `payload` (https://github.com/nicolodavis/boardgame.io/blob/master/src/master/master.js#L71), we currently just allow the action, but should disallow it instead.

blunket:
This fixes #410. Unless that issue is way more complex than I thought.

Sorry I been gone for a while. Still want to help a lot with this repo. :smile: 

#### Checklist

### DIFF
--- a/src/master/master.js
+++ b/src/master/master.js
@@ -69,7 +69,7 @@ export const isActionFromAuthenticPlayer = ({
   }
 
   if (!action.payload) {
-    return true;
+    return false;
   }
 
   const hasCredentials = Object.keys(gameMetadata.players).some(key => {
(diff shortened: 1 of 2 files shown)


nicolodavis:
Welcome back @blunket!

This PR does indeed fix the case where a player tries to force their way through in an authenticated game by merely removing the payload.

However, I think we might also be inadvertently disallowing moves when the game doesn't require authentication in the first place.

blunket:
Thanks for the review @nicolodavis, didn't think about that. Hopefully this update fixes that issue. I simply moved the code down a few lines in both files. haha

I didn't squash, I figured you'd squash if you wanted to! Let me know if it still needs fixing.

nicolodavis:
Perfect.  Thanks a lot Andrew!

PR #436, opened 2019-08-10
Surface game metadata and player nicknames in client / react props

### LINKED ISSUE #413: surface player nicknames in client
This is available in the player metadata section [here](https://github.com/nicolodavis/boardgame.io/blob/master/src/server/api.ts#L172) and should be surfaced to the game client.

jasonharrison:
Will close #413 
#### Checklist

### DIFF
--- a/src/client/transport/local.js
+++ b/src/client/transport/local.js
@@ -121,7 +121,7 @@ export class Local {
    */
   subscribe() {}
 
-  subscribeGameMetadata() {}
+  subscribeGameMetadata(_metadata) {} // eslint-disable-line no-unused-vars
 
   /**
    * Updates the game id.
--- a/src/client/transport/local.js
+++ b/src/client/transport/local.js
@@ -121,6 +121,8 @@ export class Local {
    */
   subscribe() {}
 
+  subscribeGameMetadata() {}
+
   /**
    * Updates the game id.
    * @param {string} id - The new game id.
--- a/src/master/master.js
+++ b/src/master/master.js
@@ -259,14 +259,15 @@ export class Master {
   async onSync(gameID, playerID, numPlayers) {
     const key = gameID;
 
-    let state;
+    let state, gameMetadata;
 
     if (this.executeSynchronously) {
       state = this.storageAPI.get(key);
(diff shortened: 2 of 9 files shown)


src/master/master.js
```diff
@@ -301,7 +302,7 @@ export class Master {
     this.transportAPI.send({
       playerID,
       type: 'sync',
-      args: [gameID, filteredState, log],
+      args: [gameID, filteredState, log, gameMetadata],
```
nicolodavis: We should scrub the credentials from `gameMetadata` before it is sent over.

jasonharrison: I overlooked this, you are right.

src/master/master.js
```diff
@@ -259,14 +259,20 @@ export class Master {
   async onSync(gameID, playerID, numPlayers) {
     const key = gameID;
 
-    let state;
+    let state, gameMetadata, filteredGameMetadata;
 
     if (this.executeSynchronously) {
       state = this.storageAPI.get(key);
+      gameMetadata = await this.storageAPI.get(GameMetadataKey(gameID));
```
nicolodavis: I think you want to `await` in the `else` block, not here.

src/client/transport/local.js
```diff
@@ -121,6 +121,8 @@ export class Local {
    */
   subscribe() {}
 
+  subscribeGameMetadata() {}
```
nicolodavis: Change line 89 in this file to also reflect the new `gameMetadata` argument that's sent back (even if it doesn't really do anything in the local mode).

nicolodavis:
Thanks @jasonharrison.  Can you also double-check that this doesn't break anything when the client is used in a singleplayer mode (i.e. no lobby or multiplayer)?

haoyangnz:
@nicolodavis Are we able to release a patch that includes this? Or do we have some fixed release cycle or testing to do?
Sorry for all the questions 🙏

nicolodavis:
The release will include everything until the tip in `master`.  I want to hold off on that until I get the docs ready for all the new features that have changed the API regarding phases.

One option is you can use the library from the repo directly by using `npm link`.  That way you don't have to wait for a release.

haoyangnz:
Alright 👍

PR #440, opened 2019-08-20
Add tests to phases overhaul

delucis:
This adds tests to the `phases-overhaul` branch. They either show a bug, or test for an API implementation that I think would be more logical than the current implementation (could be a bug, or could have been a conscious decision).

---

### `turn.stages` not applied on new turn

When initialising a phase/game, the `turn.stages` object is used to set up `ctx.stage` for all players. However, given that `stages` is a property of `turn`, one might expect `stages` to also be used to update `ctx.stage` when a turn starts, but this is not the current behaviour.

To test the behaviour I would expect, I have added a test to the “militia” suite, asserting that after `endTurn`, we would expect the `'A'` stage to pass to the next player (c41a84c).

### ~~`ctx.currentPlayerMoves` exposed but not set~~

~~This branch adds a `currentPlayerMoves` field to `ctx`, but this is never incremented.~~

~~https://github.com/nicolodavis/boardgame.io/blob/6644667ee708a38db03051bdd3e269a5b027dfec/src/core/flow.js#L770~~

~~I have added a failing test for the expected behaviour (ef3fa81) and also an equivalent (passing) test for `ctx.numMoves` (fbb76ea).~~

### `ctx.numMoves` counts moves from all players

As per discussion below, for `moveLimit` to work as expected, `ctx.numMoves` should only count the current player’s moves. I have added a failing test that expects that behaviour (448c351).

---

There are a few things I haven’t manage to isolate yet, but I’ll add tests as I figure them out.

### DIFF
--- a/src/core/flow.test.js
+++ b/src/core/flow.test.js
@@ -155,6 +155,20 @@ describe('phases', () => {
   });
 });
 
+describe('move counts', () => {
+  test('numMoves', () => {
+    const flow = Flow({});
+    let state = flow.init({ ctx: flow.ctx(2) });
+    expect(state.ctx.numMoves).toBe(0);
+    state = flow.processMove(state, makeMove().payload);
+    expect(state.ctx.numMoves).toBe(1);
+    state = flow.processMove(state, makeMove().payload);
+    expect(state.ctx.numMoves).toBe(2);
+    state = flow.processGameEvent(state, gameEvent('endTurn'));
+    expect(state.ctx.numMoves).toBe(0);
+  });
+});
+
 test('moveLimit', () => {
   {
     const flow = Flow({
--- a/src/core/flow.test.js
+++ b/src/core/flow.test.js
@@ -167,6 +167,30 @@ describe('move counts', () => {
     state = flow.processGameEvent(state, gameEvent('endTurn'));
     expect(state.ctx.numMoves).toBe(0);
   });
+
+  test('currentPlayerMoves', () => {
+    const flow = Flow({});
+    let state = flow.init({ ctx: flow.ctx(2) });
+    console.log({ flow, state });
+    expect(state.ctx.currentPlayerMoves).toBe(0);
+    state = flow.processMove(
+      state,
+      makeMove(undefined, undefined, '0').payload
+    );
+    expect(state.ctx.currentPlayerMoves).toBe(1);
+    state = flow.processMove(
+      state,
+      makeMove(undefined, undefined, '0').payload
+    );
+    state = flow.processMove(
+      state,
+      makeMove(undefined, undefined, '1').payload
+    );
+    expect(state.ctx.currentPlayerMoves).toBe(2);
+    expect(state.ctx.numMoves).toBe(3);
+    state = flow.processGameEvent(state, gameEvent('endTurn'));
+    expect(state.ctx.currentPlayerMoves).toBe(0);
+  });
 });
 
 test('moveLimit', () => {
--- a/src/core/flow.test.js
+++ b/src/core/flow.test.js
@@ -155,42 +155,17 @@ describe('phases', () => {
   });
 });
 
-describe('move counts', () => {
-  test('numMoves', () => {
-    const flow = Flow({});
-    let state = flow.init({ ctx: flow.ctx(2) });
-    expect(state.ctx.numMoves).toBe(0);
-    state = flow.processMove(state, makeMove().payload);
-    expect(state.ctx.numMoves).toBe(1);
-    state = flow.processMove(state, makeMove().payload);
-    expect(state.ctx.numMoves).toBe(2);
-    state = flow.processGameEvent(state, gameEvent('endTurn'));
-    expect(state.ctx.numMoves).toBe(0);
-  });
-
-  test('currentPlayerMoves', () => {
-    const flow = Flow({});
-    let state = flow.init({ ctx: flow.ctx(2) });
-    console.log({ flow, state });
-    expect(state.ctx.currentPlayerMoves).toBe(0);
-    state = flow.processMove(
-      state,
-      makeMove(undefined, undefined, '0').payload
-    );
-    expect(state.ctx.currentPlayerMoves).toBe(1);
-    state = flow.processMove(
-      state,
-      makeMove(undefined, undefined, '0').payload
-    );
-    state = flow.processMove(
-      state,
-      makeMove(undefined, undefined, '1').payload
-    );
-    expect(state.ctx.currentPlayerMoves).toBe(2);
-    expect(state.ctx.numMoves).toBe(3);
-    state = flow.processGameEvent(state, gameEvent('endTurn'));
-    expect(state.ctx.currentPlayerMoves).toBe(0);
-  });
+test('numMoves', () => {
+  const flow = Flow({});
+  let state = flow.init({ ctx: flow.ctx(2) });
+  expect(state.ctx.numMoves).toBe(0);
+  state = flow.processMove(state, makeMove(undefined, undefined, '0').payload);
+  expect(state.ctx.numMoves).toBe(1);
+  state = flow.processMove(state, makeMove(undefined, undefined, '0').payload);
+  state = flow.processMove(state, makeMove(undefined, undefined, '1').payload);
+  expect(state.ctx.numMoves).toBe(2);
+  state = flow.processGameEvent(state, gameEvent('endTurn'));
+  expect(state.ctx.numMoves).toBe(0);
 });
 
 test('moveLimit', () => {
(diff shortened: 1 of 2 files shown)


nicolodavis:
@delucis Thanks for doing this.  This is really helpful!

### currentPlayerMoves
This was an unintentional addition caused by a global rename.  I've removed it.

### stages
I agree that it's a bit counter-intuitive.  There are a couple of problems here:

1. `turn.order` contains information about how the turn is passed between players and also how a single turn can be subdivided into stages.

Maybe we should just move `stages` out of `order` and put it inside `turn` directly?

```
turn: {
  order: { TurnOrder.DEFAULT },
  stages: { ... },
}
```

2. How do we handle the case where the stages should be set only at the beginning of the phase?

Maybe this is something that we shouldn't really worry about because it's rare (maybe?).  Either way, you can also do:

```
phases: {
  A: {
    onBegin: (G, ctx) => ctx.events.setStage({ ... })
  }
}
```

delucis:
Oh right, I forgot it’s `turn.order.stages`. I guess the problem with moving it is that there’s already `turn.stages` for the stage definitions. Maybe that’s a sign that having two `stages` fields that do different things is not ideal. Perhaps they could be renamed:

```js
phase: {
  turn: {
    order: TurnOrder.DEFAULT,
    stageInit: { currentPlayer: 'A', others: 'B' },
    stages: { A: {}, B: {} }
  }
}
```

Then perhaps you could have a `stageInit` object per `phase` _and_ per `turn`? (Kind of like `onBegin` and other events.)

```js
{
  phase: {
    stageInit: { currentPlayer: 'A' },
    turn: {
      order: TurnOrder.DEFAULT,
      stageInit: { currentPlayer: 'A', others: 'B' },
      stages: { A: {}, B: {} }
    }
  }
}
```

With that API you could choose to leave out `turn.stageInit` and only set stages when the phase starts.

In cases where both `phase.stageInit` and `phase.turn.stageInit` are present, `phase.stageInit` would take precedent.

---

For your last example, I think this is a bug, but haven’t figured out how to write a test for it yet. I found that running `setStage` in a `turn.onBegin` failed as follows:

```js
phase: {
  turn: {
    onBegin: (G, ctx) => ctx.events.setStage({ currentPlayer: 'A', others: 'B' })
  }
}
```

resulted in:

```js
ctx: {
  stage: {
    0: 'B',
    1: 'B',
    undefined: 'A'
  }
}
```

My initial test suggested that `ctx.currentPlayer` was defined as expected inside the `onBegin` call, so I’m not sure exactly why `setStage` isn’t seeing that. Will have to dig deeper to isolate it and make sure it’s not a bug in my specific game implementation.

delucis:
Oh, and couldn’t `currentPlayerMoves` or similar be useful? I noticed most of the move statistics were removed here and the new `moveLimit` only checks `numMoves`. How would you end a turn after `n` moves by a specific player, while also allowing moves by other players?

Perhaps `ctx.numMoves` could look like:

```js
{
 0: 1,
 1: 2,
 total: 3
}
```

nicolodavis:
Ah, yes, we'd have to use two different names.  We should probably use the name of the event itself:

```js
turn: {
  order: { ... },
  stages: { ... },
  setStage: { currentPlayer: 'A', others: 'B' }
}
```

About having this option in both the phase config as well as the turn config, I'm not so sure.  The current semantics of how the config works is as follows:

```js
// No phases.
const game = {
  moves: { ... },
  turn: { ... },
}

// With phases.
const game = {
  moves: { ... },
  turn: { ... },

  phases: {
    A: {
      moves: { ... },  // in phase A we use this instead of the global moves.
      turn: { ... },      // in phase A we use this instead of the global turn config.
    }
  }
}
```

nicolodavis:
> Oh, and couldn’t `currentPlayerMoves` or similar be useful? I noticed most of the move statistics were removed here and the new `moveLimit` only checks `numMoves`. How would you end a turn after `n` moves by a specific player, while also allowing moves by other players?
> 
> Perhaps `ctx.numMoves` could look like:
> 
> ```js
> {
>  0: 1,
>  1: 2,
>  total: 3
> }
> ```

The move statistics were removed because they were showing up prominently on a performance benchmark I ran.  So, I want to eventually add them in as an optional plugin so that users that don't want them don't take a performance hit.

You have a good point about `numMoves` not being the right value to look at to compare against `moveLimit` when multiple players can play on the same turn.  The fix is easy, though.  We just increment `numMoves` for the current player (and `moveLimit` is only applied to the current player).

nicolodavis:
The following test passes for me:

```js
  const turn = {
    stages: { A: {}, B: {} },
    onBegin: (G, ctx) => ctx.events.setStage({
      currentPlayer: 'A',
      others: 'B',
    })
  };

  const game = {
    phases: {
      A: { turn, start: true },
    },
  };

  const client = Client({ game, numPlayers: 3 });

  expect(client.getState().ctx.phase).toBe('A');
  expect(client.getState().ctx.stage).toEqual({
    '0': 'A',
    '1': 'B',
    '2': 'B',
  });
```

Perhaps your phase config was incorrect?  It should be:

```
phases: {
  phaseName: { ... },
},
```

If you wrote:

```
phases: {
  turn: { ... }
}
```

it will actually create a phase called "turn".

delucis:
`stages` can only be a property of `turn` right? In that case, you’re probably right: `turn.setStage` can set stages every turn and for the less common case of setting stages, `phase.onBegin` can be used.

Interesting that the statistics were so heavy. If `numMoves` reflects the `currentPlayer` and `moveLimit` looks at that, then it sounds like a move statistics plug-in would work (definitely not needed for most games I’d imagine).

nicolodavis:
Yes, `stages` can only be a property of `turn`.

delucis:
> The following test passes for me:
> 
> ```js
>   const turn = {
>     stages: { A: {}, B: {} },
>     onBegin: (G, ctx) => ctx.events.setStage({
>       currentPlayer: 'A',
>       others: 'B',
>     })
>   };
> 
>   const game = {
>     phases: {
>       A: { turn, start: true },
>     },
>   };
> 
>   const client = Client({ game, numPlayers: 3 });
> 
>   expect(client.getState().ctx.phase).toBe('A');
>   expect(client.getState().ctx.stage).toEqual({
>     '0': 'A',
>     '1': 'B',
>     '2': 'B',
>   });
> ```

I think the problem would be the second turn. First turn works for me too (maybe because of `stages` setting `ctx.stage` correctly). For the second turn, `stages` is ignored and `onBegin` applies and errors.

nicolodavis:
```js
  const turn = {
    stages: { A: {}, B: {} },
    onBegin: (G, ctx) => ctx.events.setStage({
      currentPlayer: 'A',
      others: 'B',
    })
  };

  const game = {
    phases: {
      A: { turn, start: true },
    },
  };

  const client = Client({ game, numPlayers: 3 });

  expect(client.getState().ctx.currentPlayer).toBe('0');
  expect(client.getState().ctx.stage).toEqual({
    '0': 'A',
    '1': 'B',
    '2': 'B',
  });

  client.events.endTurn();

  expect(client.getState().ctx.currentPlayer).toBe('1');
  expect(client.getState().ctx.stage).toEqual({
    '0': 'B',
    '1': 'A',
    '2': 'B',
  });
```

This passes as well.  Is this the behavior that you would expect?

Also, to clarify, `stages` just contains a description of the stages (it does not set anything in `ctx`).  In the example above `onBegin` is doing that for both the first and second turns.

delucis:
Huh! Yeah, that looks good to me. I guess there’s something wrong in my game implementation.

nicolodavis:
Also, I just realized that calling `setStage` only works when called from a `turn.onBegin` (i.e. the beginning of a player's turn).

You can't really call it from a phases's `onBegin` because there is no turn that is active yet.  So, if someone wanted to only call `setStage` at the beginning of the phase (i.e. on the turn of the first player to play in that phase), they would need to call it inside `turn.onBegin` like this:

```js
phases: {
  A: {
    turn: {
      onBegin: (G, ctx) {
        if (!G.stageSet) ctx.events.setStage(...);
        G.stageSet = true;
      } 
    }
  }
}
```

Not the most elegant, but I think it's worth the savings in API surface area for a situation that we think is rare.

delucis:
Ok, not sure what exactly I’d done wrong before, but I now have the `turn.onBegin`  call to `setStage` working too.

I’ve updated this branch to remove the `currentPlayerMoves` test and update the `numMoves` test to expect it only to be incremented by moves by the current player.

nicolodavis:
OK, it looks like the two tests in this PR now pass.  Are all the points raised in this PR addressed now?

delucis:
@nicolodavis Yes, for now I think that’s everything!

I’ll keep working with the branch and let you know if I find anything else. Thanks for the taking the time to turn this around!

nicolodavis:
Thank *you* for helping me test this!

I'll close this PR for now (feel free to either re-open the same or open a new one as you find more issues).

PR #441, opened 2019-08-20
feat: Expose the `Stage` defaults object via core package

delucis:
Because `setStage` has been moved out of `turn.order`, the `TurnOrder` defaults have been split in `src/core/turn-order.js`. This PR exposes the new `Stage` object via the core package to use like the `TurnOrder` object:

```js
import { Stage } from 'boardgame.io/core'
```

```js
turn: {
  setStage: Stage.ANY_ONCE
}
```

### DIFF
--- a/packages/core.js
+++ b/packages/core.js
@@ -7,7 +7,7 @@
  */
 
 import { INVALID_MOVE } from '../src/core/reducer.js';
-import { Pass, TurnOrder } from '../src/core/turn-order.js';
+import { Pass, Stage, TurnOrder } from '../src/core/turn-order.js';
 import { PlayerView } from '../src/core/player-view.js';
 
-export { TurnOrder, Pass, PlayerView, INVALID_MOVE };
+export { Stage, TurnOrder, Pass, PlayerView, INVALID_MOVE };


nicolodavis:
Thanks @delucis!

Do you have any opinions on naming here?  I'm still not entirely satisfied with `setStage` and `Stage`.

For example:

```js
turn: {
  setStage: Stage.ANY,
}
```

reads like it is setting the stage of the current player to some sort of "any" value (i.e. the player isn't in any particular stage), but it is actually configuring the turn so that *any* player can make a move.

What we want is something that conveys the following:

```js
turn: {
  setActivePlayersAndTheStagesThatTheyAreIn: ActivePlayers.ANY,
}
```

but not as verbose of course.

delucis:
Definitely agree that the naming is not ideal. How about `StagePlayers` just for the defaults object?

```js
turn: {
  setStage: StagePlayers.ANY,
}
```

I think `setStage` is clear, because it sets `ctx.stage`. As long as it’s clearly documented, I don’t think having a defaults object that doesn’t match exactly is a problem, because it’s providing a specific subset of stage functionality.

The only change I might make would be to make it plural: `setStages` and `ctx.stages`? That might clarify that we’re not in a single stage, but stages apply to different players.

nicolodavis:
After having thought about this for a while, I actually like `activePlayers`.  Here is my thinking:

The call:

```js
turn: {
  activePlayers: ActivePlayers.ANY,
}
```

is really configuring which players can play in this turn.  It can also set what stages they are in, but this is actually a completely optional step.  In fact, most of the bundled stage configurations like ANY, OTHERS etc. actually set the stage to the empty string (I should probably use a different value than an empty string to designate a special value here, but that's a separate discussion).  They're just adding players to the set of "active players" without imposing a stage on them.

The event can then be called `setActivePlayers` if called from a move:

```js
move: (G, ctx) => {
  ctx.events.setActivePlayers(...);
}
```

The argument to this could be one of:
`ActivePlayers.ALL` (I think ALL is a better choice than ANY)
`ActivePlayers.OTHERS`
...
or a custom object specifying the options manually.

nicolodavis:
About how this will be stored in `ctx`, we can probably use `ctx.activePlayers`.

delucis:
@nicolodavis That sounds good to me!

If it’s helpful, I reckon I can do the renaming work in #442 or — perhaps better — do it in a new branch.

For the remaining API in `ctx`:

- `_activePlayersOnce` instead of `_stageOnce`?
- `activePlayersDone` instead of the `stageCompleted` I proposed?

nicolodavis:
Sure, if you have the time that'll be great!  Feel free to reuse #442 or a new one if you prefer.

The two names you suggested for the properties in `ctx` sound good to me.

PR #443, opened 2019-08-23
Update examples for phases overhaul

delucis:
As discussed in #442. Mainly this updates the turn order simulator, but I also fixed the Phases Diagram example.

Some of the turn order implementations are quite different after the overhaul, so you might want to check you’re happy with my interpretation of the originals.

I’ve tweaked the simulator by adding a display for the current stage of each player and made the phase show `null` when the phase is `null` (previously it would have been `'default'` in those cases).

### DIFF
--- a/examples/react-web/src/phases/diagram.js
+++ b/examples/react-web/src/phases/diagram.js
@@ -12,10 +12,8 @@ import { Client } from 'boardgame.io/react';
 import './diagram.css';
 
 const game = {
-  moves: {},
-  startingPhase: 'A',
   phases: {
-    A: { next: 'B' },
+    A: { start: true, next: 'B' },
     B: { next: 'C' },
     C: { next: 'A' },
   },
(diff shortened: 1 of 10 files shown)


nicolodavis:
Looks great!

Couple of minor comments:
1. Let's not display 'stage undefined' (but just remove the stage message altogether for this case).
2. Can we get rid of the quotes around the stage and phase names?  It might be helpful when the stage is the empty string, but looks unnecessary otherwise.

delucis:
@nicolodavis Great!

1. Fixed this.

2. Getting rid of the quotes might actually make it less clear (for those who don’t already understand the mechanics). 3 of the 4 examples that use stages only use the empty string stage, so I think it’s helpful to show there is something there somehow. I guess there’d be an empty grey rectangle, but that isn’t obviously an empty string. Happy to change it if you’re sure, but I do think this could be clearer for someone trying to figure out the mechanics.

nicolodavis:
Oh I meant we can show the quotes if it is any empty string, but remove it when the string is longer.

delucis:
Ohhhh, right! Sorry for the misunderstanding. Should be fixed now.

PR #444, opened 2019-08-26
Maintain currentPlayer/playOrderPos when phase ends

delucis:
Currently on phase end, `playOrderPos` is hard reset to `0`:

https://github.com/nicolodavis/boardgame.io/blob/a2961c81ce1393e3d6e789e33347ead180b33983/src/core/flow.js#L562

This could make sense, but it makes it difficult to create a phase which maintains or increments the current position when it starts. If `turn.order.first` tries to use `ctx.playOrderPos` or `ctx.currentPlayer`, they have already been reset.

This PR adds a test asserting `playOrderPos` shouldn’t change and removes the hard reset. It also changes one test that was now failing, but I think shouldn’t have been given the proposed changes.

I understand the simplicity of a new phase starting new play sequence, so maybe that could be maintained as a default, but `first: (G, ctx) => ctx.playOrderPos` should probably also return the prior position to be useful.

What if the reset to `0` only took place when `turn.order.first` is undefined? For now, this would be equivalent to modifying `TurnOrder.DEFAULT`, because its `first` isn’t currently doing anything (if I’ve understood this correctly):

```diff
DEFAULT: {
-  first: (G, ctx) => ctx.playOrderPos,
+  first: () => 0
  next: (G, ctx) => (ctx.playOrderPos + 1) % ctx.playOrder.length,
}
```

### DIFF
--- a/src/core/flow.js
+++ b/src/core/flow.js
@@ -559,7 +559,7 @@ export function Flow({ moves, phases, endIf, turn, events, plugins }) {
     G = conf.onEnd(G, ctx);
 
     // Reset the phase.
-    ctx = { ...ctx, phase: null, playOrderPos: 0 };
+    ctx = { ...ctx, phase: null };
 
     // Add log entry.
     const action = gameEvent('endPhase', arg);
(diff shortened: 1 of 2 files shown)


nicolodavis:
Looks good.

> What if the reset to 0 only took place when turn.order.first is undefined?

This is probably unnecessary because the turn order can just return `0` in `first`.

Issue #445, opened 2019-08-27
reset behavior of activePlayers


nicolodavis:
Spawning off a thread from a previous discussion at #442, which is talking about (among other things) what to do to `activePlayers` after it becomes empty (this is a new concept in the `phases-overhaul` branch).

### Background

`activePlayers` is a map that keeps track of all the players that can make a move on the current turn.  This is an evolution of the current concept `actionPlayers`.  It also keeps track of what "stage" each active player is in.  A "stage" is analogous to a "phase", and basically restricts what moves the player in that stage can make.  It can be manipulated using something like this:

```js
setActivePlayers({ others: 'discard', once: true });
```

which does the following things:
- put every player (other than the current player) in `activePlayers`.
- sets the stages of those players to `discard`.
- allows them to make one move before removing them from `activePlayers`.

This allows implementing cards that allow other players to take an action before reverting back to the current player (the Militia card in Dominion, for example).

### Problem

Players are removed from `activePlayers` once they make a move.  After `activePlayers` becomes empty, it is set to `null` (which means that only the current player can play, but they are not in any stage).

However, if the `currentPlayer` was in a stage to begin with (i.e. `activePlayers` was something like  `{ currentPlayer: 'play' }` before `setActivePlayers` was called), then this state is lost.

### Proposals

1. Reset `activePlayers` to the previous state before `setActivePlayers` was called once it becomes empty.

2. Pass in the state to reset to in the `setActionPlayers` call itself.

### Notes

- Both proposals need to clearly define when to trigger the reset behavior.  Is it sufficient to just assume that a reset is required when `activePlayers` becomes empty?

nicolodavis:
Proposal 1 is implemented in https://github.com/nicolodavis/boardgame.io/commit/4a51096bee73f1caca429fa37322ed7fb76ea195.

I'll keep this RFC open for a while in case people come up with better ideas that would warrant reverting the change and implementing something else.

delucis:
I'll try out this implementation when I'm back at my computer, but I am still concerned about this approach as I wrote in [#442](https://github.com/nicolodavis/boardgame.io/pull/442#issuecomment-524446402):

> I wonder if it implements too specific a behaviour: setting `once` to true would trigger this reset behaviour, but no other setting would and there’s no `endActivePlayers` event to do the equivalent manually.

There would need to be an opportunity to idiomatically intervene on this reset behaviour to at least declare that you want to disable it or pass a function that can return `activePlayers` from state.

[...]

PR #452, opened 2019-09-10
feat: Support move limits in `setActivePlayers`

delucis:
Continuing work towards #445 

- [X] Support `moveLimit` option in `setActivePlayers`, storing result in `ctx._activePlayersMoveLimit`

- [x] Track moves by currently active players in `ctx._activePlayersNumMoves`

- [x] Reset player counts whenever `setActivePlayers` is called and when the turn ends

- [x] End stages when player has reached move limit

- [x] Remove `once` option in favour of `moveLimit`

- [x] Restore previous move limits and counts when using `revert: true`

### DIFF
--- a/examples/react-web/src/turnorder/example-others-once.js
+++ b/examples/react-web/src/turnorder/example-others-once.js
@@ -11,7 +11,7 @@ import React from 'react';
 const code = `{
   moves: {
     play(G, ctx) {
-      ctx.events.setActivePlayers({ others: 'discard', once: true });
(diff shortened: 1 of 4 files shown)


delucis:
@nicolodavis

delucis:
@nicolodavis Let me know if you think anything needs changing here.

I think it's ready to merge, but you might have better ideas for variable naming or more efficient implementations.

nicolodavis:
Will take a look soon. A bit busy this weekend, so probably sometime on Monday. Thanks for the great work!

PR #458, opened 2019-09-15
feat: Add `endStage` and `setStage` events

### LINKED ISSUE #445: reset behavior of activePlayers

### LINKED ISSUE #446: removing players from activePlayers
Spawning off a thread from a previous discussion at #442, which is talking about (among other things) the behavior of how players are removed from `activePlayers`.

### Background

`activePlayers` is a map that keeps track of all the players that can make a move on the current turn.  This is an evolution of the current concept `actionPlayers`.  It also keeps track of what "stage" each active player is in.  A "stage" is analogous to a "phase", and basically restricts what moves the player in that stage can make.  It can be manipulated using something like this:

```js
setActivePlayers({ others: 'discard', once: true });
```

which does the following things:
- put every player (other than the current player) in `activePlayers`.
- sets the stages of those players to `discard`.
- allows them to make one move before removing them from `activePlayers`.

This allows implementing cards that allow other players to take an action before reverting back to the current player (the Militia card in Dominion, for example).

### Problem

`once: true` is too specific.

- What if the player must make 2 moves before they are removed?
- What if some other condition has to be satisfied before the player is removed?

### Proposals

1. Add `moveLimit` to the `stages` config.  Now each stage can declare how many moves ought to be made before a player is removed from `activePlayers`.  This has the benefit that each stage can implement a different `moveLimit`.  The downside is that it cannot be customized at the call-site.  It also has the downside that we have to track the number of moves made by each player, which is additional state that can slow things down.

2. Create a `removeFromActivePlayers` event that can be called from a move.  This can be used to remove players after a complex condition is met (rather than merely checking how many moves they have made).

### LINKED ISSUE #447: transitioning between stages
In the `phases-overhaul` branch, the only way to manipulate a player's stage is the `setActivePlayers` event.

For example, to set the current player's stage to 'A', you can do:
```js
ctx.events.setActivePlayers({ currentPlayer: 'A' });
```

However, this has two problems:

1. It clobbers the state in `activePlayers` (if some other player was in there, they would be removed).

2. It cannot set the stage of the player that made the move (it only works on the "current player").

We could change the semantics of `setActivePlayers` to overcome these, or split it into multiple events that do different things.

delucis:
Closes #445
Closes #446
Closes #447
🎉 

I have tried adding the `endStage` and `setStage` events as discussed in #445, #446 and #447.

I’m most of the way there, but ran into a roadblock with adding the `moveLimit` option for `setStage`. If I understood correctly (and I’ll admit I felt a little lost reading through all the event-related code), when calling an event from the client, the single passed argument is used as the payload for the action sent to the reducer. So that makes my imagined syntax of `setStage('B', { moveLimit: 2 })` not as trivial to implement as I imagined.

Presumably there would need to be a wrapper somewhere that formats arguments to client events so that supporting a second optional argument is possible, but I’m not sure exactly where or how that should be implemented.

In the meantime, I have added a failing test for `moveLimit` and `UpdateStage` in flow.js does contain the correct logic to hypothetically update `ctx` were the `moveLimit` argument to reach it. If you have any pointers, I can try to implement something, or if you prefer to get this merged without the `moveLimit` support, I can strip those parts out to be figured out later.


nicolodavis:
The easiest solution might be to use the short form / long form pattern that we follow elsewhere.

```js
setStage('B')
setStage({ stage: 'B, moveLimit: 2 })
```

So now you only have one argument.

delucis:
Oh, right, good idea. I’ll go with that for now :-)

delucis:
@nicolodavis How should I test for the `automatic` option to `endStage` and should it even exist? I’m guessing for `endPhase` and `endTurn` this is passed when it is called from `endIf`, which isn’t currently available for stages.

nicolodavis:
The only "automatic" end of stage that we have right now is triggered via `moveLimit`.  Maybe have `ProcessMove` call `EndStage` and pass `automatic: true`?

delucis:
OK!

One thing: right now `endStage` and `setStage` operate on `currentPlayer` only, but `moveLimit` can remove any active player. I’ll try adding a `playerID` optional argument that is used internally by `ProcessMove`.

If I understood correctly, events are only callable by `currentPlayer` right? What if an active player that is not the current player wants to `endStage` from a move? Would that work?

nicolodavis:
> If I understood correctly, events are only callable by currentPlayer right? What if an active player that is not the current player wants to endStage from a move? Would that work?

We might have to relax the condition.  Maybe let's have any player that is "active" (i.e. one of the `activePlayers` if it is set OR `currentPlayer` if it is not set) be able to call events?

delucis:
> let's have any player that is "active" (i.e. one of the `activePlayers` if it is set OR `currentPlayer` if it is not set) be able to call events?

That makes sense to me. I guess at some point some config for what events are enabled etc. might be needed, but for now stages are the loosest of all the game elements. (Unlike phases there are no checks that a stage is actually declared when setting it and at the moment they only declare moves.)

I’m not sure exactly how to go about those changes, so I’ll proceed like I suggested with `ProcessMove` and `EndStage` as that doesn’t need the changes to events in general.

nicolodavis:
Thanks again for getting the `phases-overhaul` effort over the finish line!

delucis:
It’s been fun slowly figuring out how all the internals work!

`ProcessMove` now uses `EndStage` as you suggested — avoids some code duplication too, which is nice.

delucis:
OK, I think this is good to go. Test coverage is just missing one branch where `events.setStage` passed to `Flow` is not undefined. Haven’t fixed that yet because I noticed that at the end of `Flow`, `events.endPhase` is used to enable both the `endPhase` and `startPhase` events. I copied that pattern for `setStage` and `endStage`, but if I’ve understood everything correctly this means that passing

```js
Flow({
  events: {
    setStage: false,
    setPhase: false,
  },
});
```

will not actually have any effect. So I guess, maybe the logic that sets `events.setStage` and `events.setPhase` to `true` is redundant and should be removed? (All tests pass without them.)

delucis:
Also, just realised this will resolve everything in #445 #446 & #447 😄 🎉

src/core/flow.js
```diff
@@ -498,6 +505,44 @@ export function Flow({ moves, phases, endIf, turn, events, plugins }) {
...
+      _activePlayersMoveLimit,
+      _activePlayersNumMoves,
+    } = ctx;
+
+    if (arg.stage) {
+      if (activePlayers === null) {
+        activePlayers = {};
+      }
+      activePlayers[playerID] = arg.stage;
+      _activePlayersNumMoves[playerID] = 0;
```
nicolodavis: Do we need to set this if a `moveLimit` is not defined?

src/core/flow.js
```diff
@@ -498,6 +505,44 @@ export function Flow({ moves, phases, endIf, turn, events, plugins }) {
...
+      activePlayers,
+      _activePlayersMoveLimit,
+      _activePlayersNumMoves,
+    } = ctx;
+
+    if (arg.stage) {
+      if (activePlayers === null) {
+        activePlayers = {};
+      }
+      activePlayers[playerID] = arg.stage;
```
nicolodavis: Should we check if this is a valid stage before letting the user enter it (similar to what we do for phases)?

src/core/flow.js
```diff
@@ -631,6 +676,63 @@ export function Flow({ moves, phases, endIf, turn, events, plugins }) {
...
   }
 
+  function EndStage(state, { arg, next, automatic, playerID }) {
+    playerID = playerID || state.ctx.currentPlayer;
+
+    let { ctx } = state;
+    let { activePlayers, _activePlayersMoveLimit } = ctx;
+
+    if (next) {
+      next.push({ fn: UpdateStage, arg, playerID });
```
nicolodavis: Don't add UpdateStage to `next` if `arg` is undefined (as opposed to adding it and then having `UpdateStage` being a no-op).

src/core/turn-order.js
```diff
@@ -131,6 +131,51 @@ export function SetActivePlayers(ctx, arg) {
   };
 }
 
+/**
+ * Update activePlayers, setting it to previous, next or null values
+ * when it becomes empty.
+ * @param {Object} ctx
+ */
+export function UpdateActivePlayers(ctx) {
```
nicolodavis: Let's rename this to a more descriptive `UpdateActivePlayersOnceEmpty`.

src/core/flow.js
delucis: When implementing `moveLimit`, I decided to always increment `_activePlayersNumMoves` in `ProcessMove` even when `moveLimit` is not enabled, which means it would need resetting for new stages.

https://github.com/nicolodavis/boardgame.io/blob/297eb87433ce165d5ff231ef7485a05e1192d797/src/core/flow.js#L695

It isn’t used by the framework if `moveLimit` is not defined, but I thought it might be helpful for people who want to keep track of moves without using `moveLimit`.

nicolodavis: OK.

src/core/flow.js
delucis: That’s not currently the case in `setActivePlayers`. I would suggest merging it as is and then tightening up things like that in a separate PR. (I can’t see any reason to _not_ ensure stages are valid.)

nicolodavis: OK sure.

PR #465, opened 2019-09-17
 Upgrade firebase to fix build on node 12

### LINKED ISSUE #463: Fails to build with node v12
Fails to build grpc@1.20.0 (see https://github.com/grpc/grpc-node/issues/834)

```
$ npm i

node-pre-gyp WARN Tried to download(404): https://node-precompiled-binaries.grpc.io/grpc/v1.20.0/node-v72-darwin-x64-unknown.tar.gz 
node-pre-gyp WARN Pre-built binaries not found for grpc@1.20.0 and node@12.10.0 (node-v72 ABI, unknown) (falling back to source compile with node-gyp) 

...

../ext/channel.cc:292:56: error: too few arguments to function call, expected 2, have 1
  int try_to_connect = (int)info[0]->Equals(Nan::True());
```

```
$ npm-why grpc
  boardgame.io > firebase > @firebase/firestore > grpc@1.20.0
  boardgame.io > firebase-admin > @google-cloud/firestore > google-gax > grpc@1.20.0
  boardgame.io > firebase-admin > @google-cloud/firestore > google-gax > grpc-gcp > grpc@1.20.0
```
Can we upgrade firebase?

haoyangnz:
#### Checklist

closes #463


package.json
```diff
@@ -130,9 +131,11 @@
     "superagent": "^3.8.3",
     "supertest": "^3.1.0",
     "ts-jest": "^24.0.2",
+    "ts-node": "^8.4.1",
     "ts-transformer-imports": "^0.4.3",
     "ttypescript": "^1.5.7",
-    "typescript": "^3.5.3"
+    "typescript": "^3.5.3",
+    "webpack": "^4.40.2"
```
nicolodavis: Why is webpack being added?

package.json
```diff
@@ -130,9 +131,11 @@
     "superagent": "^3.8.3",
     "supertest": "^3.1.0",
     "ts-jest": "^24.0.2",
+    "ts-node": "^8.4.1",
```
nicolodavis: Also curious about ts-node.

package.json
haoyangnz: There were peer dependency warnings.

nicolodavis: But isn't that something that the user of the library needs to deal with? We only add peer deps here to get the tests working. The actual users of the library install firebase themselves.

haoyangnz: They have nothing to do with firebase. There were other libraries probably babel loader that had a peer dep on it, so I added to dev deps. Would you like me to remove them?

nicolodavis: I think let's keep the deps to a minimum here.  We're only upgrading firebase because the tests don't run on Node v12 correct?

haoyangnz: Upgrading firebase because `npm i` fails.

nicolodavis: Got it.  In that case let's just upgrade firebase without adding any additional deps that are not needed here.

haoyangnz: Removed.

package.json
haoyangnz: Removed.

package.json
```diff
@@ -75,6 +75,7 @@
     "@babel/preset-env": "^7.5.4",
     "@babel/preset-react": "^7.0.0",
     "@babel/preset-typescript": "^7.3.3",
+    "@firebase/app": "^0.4.16",
```
haoyangnz: I think I can remove this too. Hold on.

haoyangnz: This is needed for tests to pass.

haoyangnz:
@nicolodavis All done.

package.json
nicolodavis: Still a bit confused about this.  Shouldn't this be pulled in as one of the deps of the firebase package below?

nicolodavis: Notice that we already have it in `package-lock.json`, which means that something else is already pulling it in automatically.

haoyangnz: Installing `firebase@latest` and `firebase-admin@latest` removes `@firebase/app` from `package-lock.json`. Manually maintaining `package-lock.json` is probably not a good idea?

nicolodavis: We don't manually maintain it. Stuff gets added there when you run `npm install`.

So, it sounds like firebase no longer pulls in @firebase/app and they expect users to directly install that package?

haoyangnz: Correct.

nicolodavis:
Thanks for fixing this @haoyangnz!

PR #474, opened 2019-09-26
Disallow game names with spaces

### LINKED ISSUE #459: Multi word game name with Title Case causes indefinite connecting... in Lobby
Take the tic tac toe game, change the game name from `tic-tac-toe` to `Tic Tac Toe` and the lobby shows `connecting...` indefinitely when trying to start the game.

jasonharrison:
Fix #459 by throwing an error when encountering a game name with spaces.

#### Checklist

### DIFF
--- a/src/core/game.js
+++ b/src/core/game.js
@@ -107,6 +107,7 @@ export function Game(game) {
 
   return {
     ...game,
+    name: game.name.replace(' ', '-'),
 
     moveNames: game.flow.moveNames,
 
--- a/src/core/game.js
+++ b/src/core/game.js
@@ -25,38 +25,6 @@ import { Flow } from './flow';
  * action.args contain any additional arguments as an
  * Array.
  *
- * ({
- *   name: 'tic-tac-toe',
- *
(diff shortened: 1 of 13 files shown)


src/core/game.js
```diff
@@ -107,6 +107,7 @@ export function Game(game) {
 
   return {
     ...game,
+    name: game.name.replace(' ', '-'),
```
nicolodavis: I would think that this is the only change that is necessary.  Why do we also need to change socketio.js?

src/server/api.ts
```diff
@@ -82,7 +82,7 @@ export const addApiToServer = ({ app, db, games, lobbyConfig }) => {
 
   router.post('/games/:name/create', koaBody(), async ctx => {
     // The name of the game (for example: tic-tac-toe).
-    const gameName = ctx.params.name;
+    const gameName = ctx.params.name.replace(' ', '-');
```
nicolodavis: Hmm, I think this might open a whole other can of worms (if we start to allow malformed names in the API).

Issue #472, opened 2019-09-23
Support dynamic/changing number of iterations in MCTS bot 


jchadwick:
When using the out-of-box MCTS in a game where the number of available moves (and therefore number of possible game states) declines as the game progresses, the time to calculate the best move reduces over time.

In other words: the _time_ to calculate _N_ `iterations` reduces as the `turn` count increases.

Currently, the `iterations` value is hard-coded when the bot is instantiated.  This should be updated to accept a function (much like the `objectives` function) that can return a dynamic number of iterations to support scenarios such as increasing the iteration count as the game progresses while the cost (in time) stays relatively flat.

The [AlphaBeta library](https://github.com/panchishin/alphabeta) offers [an alternative solution](https://github.com/panchishin/alphabeta#alphabetastepformilliseconds-milliseconds--callback-) which achieves the same effect of taking the same amount of time - running the algorithm for as many iterations as it can until the wall time expires - which also seems like a great option.

nicolodavis:
Definitely open to the idea of having `iterations` accept a function in addition to an integer. Sounds like a great idea!

I'm probably not going to be able to get to this for a while, though.  Would you have time to make a PR for this?

jchadwick:
Yes, I'd love to!

nicolodavis:
Thanks a lot!  Let me know if you need pointers to the code or clarifications if anything is not clear.

jchadwick:
Oh that sounds great, too!  I’ll look into it.

jchadwick:
FYI, I only implemented the dynamic values and not the Debug Panel integration in this PR.  I've added a new issue (#476) to track the Debug Panel which I would like to do, too.

jchadwick:
Oh, BTW, I included `playoutDepth`, too, because why the hell not. :)

delucis:
Closing as this was added in [v0.35.0](https://github.com/nicolodavis/boardgame.io/blob/master/docs/documentation/CHANGELOG.md#v0350).

PR #475, opened 2019-09-26
Allowing support for both numbers and functions for MCTS bot iterations and playoutDepth

jchadwick:
Context: https://github.com/nicolodavis/boardgame.io/issues/472

#### Checklist


jchadwick:
@googlebot I signed it!

jchadwick:
Ready for review, @nicolodavis

src/ai/bot.js
```diff
@@ -211,11 +221,13 @@ export class MCTSBot extends Bot {
...
 
-    for (
-      let i = 0;
-      i < this.playoutDepth && state.ctx.gameover === undefined;
-      i++
-    ) {
+    const playoutDepth = this.playoutDepth(
+      state.G,
+      state.ctx,
+      state.ctx.currentPlayer
```
nicolodavis: Why are we passing `currentPlayer` here? I would think that the playoutDepth shouldn't really depend on whose turn it is.  Also, you're passing `ctx` in, which contains this field anyway.

src/ai/bot.js
```diff
@@ -276,7 +288,9 @@ export class MCTSBot extends Bot {
   play(state, playerID) {
     const root = this.createNode({ state, playerID });
 
-    for (let i = 0; i < this.iterations; i++) {
+    const iterations = this.iterations(state.G, state.ctx, playerID);
```
nicolodavis: Similar.  Do we need to pass `playerID` in?

src/ai/bot.js
```diff
@@ -113,10 +113,20 @@ export class MCTSBot extends Bot {
       objectives = () => ({});
     }
 
+    if (typeof iterations === 'number') {
```
nicolodavis: If you move these lines below 129 (and also adjust `this.iterations` / `iterations`), then you don't have to change the code there that's setting the default values.

jchadwick: Honestly I think that is just my TypeScript kicking in. Doing it this way means that the constructor arg might be either number or func but the property will only ever be a func. Simpler that way. I can change it if you like, though.

src/ai/bot.js
jchadwick: I agree. I was only trying to keep the API consistent since it seems most other methods throughout the framework (including/especially the sibling ‘objectives’ method) seemed to have this signature.  I was curious about why player ID was added but I went along with it anyway.

src/ai/bot.js
nicolodavis: Oh I see.  OK, let's leave it as is then.

src/ai/bot.js
jchadwick: Did you want me to remove the `playerID` parameter, @nicolodavis?  I'm happy to do that.  But, if so, I'd also encourage we look at the other places where this is passed in unnecessarily.

nicolodavis: Yes, let's remove it for the code related to the present PR.

About the usage in other places, we only really need it to enumerate moves (which is why you see it in `createNode` and `playout`).  `objectives` are also player specific, which is why the `playerID` is passed there.  Let me know if there is any specific spot where you think it is redundant and we can revisit it.

nicolodavis:
Thanks @jchadwick!

PR #480, opened 2019-09-27
docs: Minor tweaks and updates

delucis:
I read through the new docs and here are a handful of tiny proposed changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,7 +44,7 @@ $ npm run test:coverage
 Use the following command to preview them:
 
 ```
-$ npm run docsify
+$ npm run docs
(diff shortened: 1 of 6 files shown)


delucis:
Couple of suggestions I haven’t implemented:

1. What do you think about not including `propTypes` in [the React events example](
https://gracious-swanson-64c0ae.netlify.com/documentation/#/events?id=triggering-an-event-from-a-react-client), just to keep it to the bare minimum?

2. Should [the Game API docs](https://gracious-swanson-64c0ae.netlify.com/documentation/#/api/Game?id=game) mention that the `name` must not include spaces (or other prohibited characters?) as discussed in #459?

PR #481, opened 2019-09-28
refactor: Change moveLimit syntax in setActivePlayers

### LINKED ISSUE #479: moveLimit syntax inside setActivePlayers
I'm not the biggest fan of having a parallel structure inside `moveLimit`:

```js
setActivePlayers({
  player: 'A',
  others: 'B',
  moveLimit: {
    player: 1,
    others: 2,
  }
})
```

How about we support the following instead:

```js
setActivePlayers({
  player: { stage: 'A', moveLimit: 1 },
  others: { stage: 'B', moveLimit: 2 },
})
```

This is also consistent with how we do it in `setStage` (you can either pass a single string which is treated as a stage name, or an object containing the stage and `moveLimit`).

delucis:
Closes #479 

#### Design choices

- Player-specific move limits take precedence. For example, the following sets the player’s move limit to 2, not 1:

    ```js
    setActivePlayers({
      player: { stage: 'A', moveLimit: 2 },
      moveLimit: 1,
    })
    ```

- I wrote a helper `ApplyActivePlayerArgument()`, to simplify setting a player’s stage and move limit given an argument that can be a stage (`string`, `boolean`, `number`, `null`) or an object (shape: `{ stage, moveLimit }`). It sacrifices a little bit of efficiency, because parsing the argument happens for every player (for example for `all` and `others`) instead of once and then applied for all. I think it’s still cleaner to separate this logic out into a single function.

- Setting a move limit with the long-form syntax *requires* an accompanying stage field, so the following would not apply a limit:

    ```js
    setActivePlayers({
      all: { stage: 'A' },
      player: { moveLimit: 2 },
    });
    ```

### DIFF
--- a/src/core/turn-order.js
+++ b/src/core/turn-order.js
@@ -195,7 +195,7 @@ function ApplyActivePlayerArgument(
   playerID,
   arg
 ) {
-  if (typeof arg !== 'object' || arg === null) {
+  if (typeof arg !== 'object' || arg === Stage.NULL) {
     arg = { stage: arg };
   }
 
--- a/src/core/turn-order.js
+++ b/src/core/turn-order.js
@@ -51,74 +51,60 @@ export function SetActivePlayers(ctx, playerID, arg) {
     _prevActivePlayers = [];
   }
 
-  let activePlayers = {};
-  let _activePlayersMoveLimit = {};
+  let localCtx = {
+    activePlayers: {},
+    _activePlayersMoveLimit: {},
+  };
 
   if (Array.isArray(arg)) {
     let value = {};
     arg.forEach(v => (value[v] = Stage.NULL));
-    activePlayers = value;
+    localCtx.activePlayers = value;
   }
 
   if (arg.value) {
     for (const id in arg.value) {
-      const { stage, moveLimit } = getActivePlayerParameters(arg.value[id]);
-      if (stage !== undefined) {
-        activePlayers[id] = stage;
-        if (moveLimit) _activePlayersMoveLimit[id] = moveLimit;
-      }
+      const value = arg.value[id];
+      localCtx = ApplyActivePlayerArgument(localCtx, id, value);
     }
   }
 
   if (arg.player !== undefined) {
-    const { stage, moveLimit } = getActivePlayerParameters(arg.player);
-    if (stage !== undefined) {
-      activePlayers[playerID] = stage;
-      if (moveLimit) _activePlayersMoveLimit[playerID] = moveLimit;
-    }
+    localCtx = ApplyActivePlayerArgument(localCtx, playerID, arg.player);
   }
 
   if (arg.others !== undefined) {
-    const { stage, moveLimit } = getActivePlayerParameters(arg.others);
-    if (stage !== undefined) {
-      for (let i = 0; i < ctx.playOrder.length; i++) {
-        const p = ctx.playOrder[i];
-        if (p !== playerID) {
-          activePlayers[p] = stage;
-          if (moveLimit) _activePlayersMoveLimit[p] = moveLimit;
-        }
+    for (let i = 0; i < ctx.playOrder.length; i++) {
+      const id = ctx.playOrder[i];
+      if (id !== playerID) {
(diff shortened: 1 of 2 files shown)


src/core/turn-order.js
```diff
@@ -175,6 +164,27 @@ export function UpdateActivePlayersOnceEmpty(ctx) {
...
 }
 
+/**
+ * Apply an active player argument to the given player ID
+ * @param {Object} ctx A context object
+ * @param {String} playerID The player to apply the parameter to
+ * @param {(String|Object)} arg An active player argument
+ */
+function ApplyActivePlayerArgument(ctx, playerID, arg) {
+  if (typeof arg !== 'object' || arg === null) {
```
nicolodavis: We should use `Stage.NULL` instead of `null` directly here.

src/core/turn-order.js
```diff
@@ -51,79 +51,68 @@ export function SetActivePlayers(ctx, playerID, arg) {
     _prevActivePlayers = [];
   }
 
-  let activePlayers = {};
+  let localCtx = {
```
nicolodavis: I find it a bit confusing to use `ctx` to represent `{ activePlayers, _activePlayersMoveLimit }`.  I can see the desire to club them together for the sake of `ApplyActivePlayerArgument`.

Shall we just pass them individually to `ApplyActivePlayerArgument` like this:

```js
ApplyActivePlayerArgument(activePlayers, _activePlayersMoveLimit, playerID, value);
```

and have the function mutate them directly?  We're passing in new objects anyway, so we won't be violating any immutability principle here.

delucis: Sure. I tried sticking to the other related signatures — `(ctx, playerID, arg)` — but I was a little unhappy with how this felt too.

That said, given we’re returning `{ activePlayers, _activePlayersMoveLimit }`, I think it’s nice to pass that in in that form already. Also because otherwise in `setActivePlayers` we have to do this each time:

```js
({ activePlayers, _activePlayersMoveLimit } = ApplyActivePlayerArgument(
  activePlayers, _activePlayersMoveLimit, playerID, arg
));
```

nicolodavis: Do we need to have `ApplyActivePlayerArgument` return anything at all, though?  Can it not just mutate `activePlayers` and `_activePlayersMoveLimit` directly?

src/core/turn-order.js
delucis: Are you sure? This is checking we have a valid object (`typeof null === 'object'`, but it’s a `TypeError` to try to read its properties) rather than whether the argument is the null stage. If in the future `Stage.NULL` got switched to `'null-stage'` or something, that would produce errors here.

src/core/turn-order.js
delucis: Oh, I see, you’re suggesting mutating directly, so we’d use it simply like this:

```js
ApplyActivePlayerArgument(activePlayers, _activePlayersMoveLimit, playerID, arg);
```

OK 😀

delucis: Done in  07ca81c

src/core/turn-order.js
nicolodavis: I'm probably missing something here.  Is this code not trying to (effectively) say:

```js
if (argument is a stage name or the null stage) {
  arg = { stage: arg };
}

// from here arg is guaranteed to be an object (either passed in directly by the user or set by us above).
```

Is there any other possible reason someone might pass `null` to `arg` if not to indicate the null stage?

delucis: I was more going for `if (argument is really an object)`. You’re right that for now, `null` would usually indicate `Stage.NULL`.

A user might pass in `null` directly or have a buggy function that returns `null` unintentionally. I guess that if that happens, that’s a user error in any case, so I can change it to `Stage.NULL` if you prefer.

nicolodavis: No, it's fine as is.  Just wanted to clarify.

delucis: Done in 5c357a3.

(Sorry to be picky — just that until there’s some more checking around whether a stage argument is valid, it feels a bit weird to test against `Stage.NULL`)

Issue #383, opened 2019-03-29
Add bots to multiplayer games


jakehsiao:
I am currently writing a game using this engine. Now this feature is required.

It's a 10-person game. In single player's mode, player will play with 9 AIs. When it's AI's turn, AI needs to operate automatically(not by player clicking "step" or "simulation").

Mixed players and AIs in online game is not required now(but maybe required in the future).

If this feature is WIP, please tell me where I can help. I'm willing to implement this and contribute.

nicolodavis:
Hey! If all you need is a simple thing that activates the AI player in a non networked game, you can just call step() from your client automatically. I believe this is what @flamecoals does.

If you need bots for a multiplayer game, that's a bit more involved. For example, we probably need to support a few different configurations where the bot sits on a separate server / process and doesn't choke the game server when it is "thinking". The design for this is still WIP and feel free to get involved in this as well if you're interested.

However, for your use case, I don't think you actually need any changes to the framework (you can do everything on your client).

jakehsiao:
Adding "step" to board really let AIs moves on their own, thanks!

However, when AI let player to discard a card(set action player to "0" which is player's ID but currentPlayer is still "1"), player cannot do that unless click "updatePlayerID" on debug interface.

Here's my solution in "client/react.js":

`
 if (board) {
        _board = React.createElement(board, {
          ...state,
          ...rest,
          isMultiplayer: multiplayer !== undefined,
          moves: this.client.moves,
          events: this.client.events,
          gameID: this.gameID,
          playerID: this.playerID,
          step: this.client.step,
          reset: this.client.reset,
          undo: this.client.undo,
          redo: this.client.redo,
          updatePlayerID: this.updatePlayerID, //Added
        });
      }
`
By adding "updatePlayerID" function to board's props, it supports AI let player make a choice in AI's turn.

lukegothic:
> Hey! If all you need is a simple thing that activates the AI player in a non networked game, you can just call step() from your client automatically. I believe this is what @flamecoals does.
> 
> [...]
> 
> However, for your use case, I don't think you actually need any changes to the framework (you can do everything on your client).

Two scenarios:

1. Two-player game where one of the players is an AI, and a single move per turn (e.g. TicTacToe)
2. Another game with one or many AI players and multiple moves per turn (e.g. Dominion)

Where should the .step() be called to tell the AI to begin playing? What checks should be performed to determine if the AI can begin to play in any of the scenarios?

[...]

PR #485, opened 2019-10-02
docs: Improve documentation around bots and `step()`

delucis:
This tries to improve the documentation around AI/bots following @julienroulle’s questions on Gitter. In the process I’ve also tidied up some small things I came across.

I’m not sure if some of this was deliberately undocumented while waiting for #383 — i.e. not explicitly advising how to call `step` from the client — but it  seemed like it might be worth adding for now.

I think [this is the CodeSandbox referenced](https://codesandbox.io/s/boardgameio-5lbe5) on Gitter, but I suspect it was not intended to have auto-stepping AI included in any case.

### DIFF
--- a/docs/documentation/tutorial.md
+++ b/docs/documentation/tutorial.md
@@ -292,16 +292,19 @@ const App = Client({
 export default App;
 ```
 
-That's it! You will notice that you now have two more options in
-the **Controls** section (`step` and `simulate`). You can use the
-keyboard shortcuts `4` and `5` to trigger them.
+That's it! Now that you have configured AI, there are two more options in
+the **Controls** section of the Debug Panel:
 
-Press `5` and just watch your game play by itself!
+- `step` causes the AI to calculate and make a single move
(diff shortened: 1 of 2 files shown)


docs/documentation/tutorial.md
```diff
@@ -312,6 +315,29 @@ hood to explore the game tree and find good moves. The default uses 1000 iterati
...
+props to your `<Board>` component. Adapting our previous example, we could add
+a call to `step` each time the human player clicks on a cell:
+
+```js
+class TicTacToeBoard extends React.Component {
+  onClick(id) {
+    if (this.isActive(id)) {
+      this.props.moves.clickCell(id);
+      this.props.events.endTurn();
+      setTimeout(() => this.props.step(), 1000);
```
nicolodavis: Let's just document this in `Client.md` for now.  I would prefer to not document it prominently in the tutorial as I expect this to change pretty soon.

delucis: 👍 Done in 88f6e00b

Issue #451, opened 2019-09-10
npm start broken on Windows


nicolodavis:
https://github.com/nicolodavis/boardgame.io/blob/master/scripts/dev.sh
doesn't run on Windows and needs to be made platform agnostic.

On Linux / OSX you can do the following to run both client and server:

```js
npm install
npm start
```

Since dev.sh is broken on Windows you can't do this at the moment, although you can use the following manual workaround:

#### Run Client

```js
npm install
cd examples/react-web
npm install
npm start
```

#### Run Server

```js
npm install
npm run dev:server
```

Stefan-Hanke:
@tsioap Refering to the PR - this is not only about the directory separator. I've just clone this repo, and `npm start` didn't work - because the `bash` command was not found (it is installed though). I'm not sure what scripting language to use for cross-platform purposes. Since NodeJS must be installed, maybe `dev.sh` could be converted to a JS script? Or maybe inline it completely into `package.json`?

Stefan-Hanke:
As a sidenote, maybe a CI job running on windows could take care that some scenarios do not break to begin with.

nicolodavis:
Yes, we'd have to make it a cross-platform script, similar to https://github.com/nicolodavis/boardgame.io/blob/master/scripts/clean.js

Welcome back @Stefan-Hanke!  Long time no see.

tisoap:
Ah, I understand now. I did work for me because I use [Git Bash for Windows](https://gitforwindows.org/). I agree a Node script would be the best choice to make it cross platform regardless of terminal used, I'll look into converting it.

nicolodavis:
Thanks @tisoap.  Feel free to update the existing PR or open a new one if you prefer.

Stefan-Hanke:
@nicolodavis thanks - hope that I can contribute on a somewhat regular basis. Let's see how things are going.

nicolodavis:
Your contributions are much appreciated.  As always, the current areas of focus are documented in the [roadmap](https://github.com/nicolodavis/boardgame.io/blob/master/roadmap.md).  The large ones are the ability to add bots to games, the game lobby and the Svelte migration.

PR #490, opened 2019-10-06
Fix npm start for Windows

tisoap:
Command `npm start` didn't work on Windows because it could not find `examples/react-web` path. This was fixed by using two cd commands and avoiding using slashes. 

Fixes issue #451

#### Checklist


tisoap:
@googlebot I signed it!

nicolodavis:
Thanks for the fix!

There is some additional discussion on https://github.com/nicolodavis/boardgame.io/issues/451 that suggests a different implementation.  Can you take a look?

tisoap:
Following the discussion at #451 , I've updated my approach to replace `dev.sh` with a Node script, so it should now work on Windows.

scripts/install-web.js
```diff
@@ -0,0 +1,33 @@
...
+  console.log('Installing web dependencies...');
+  const npmInstall = spawn('npm', ['install'], { cwd: webExamplePath });
+
+  npmInstall.stdout.on('data', function(data) {
+    console.log(data.toString());
+  });
+
+  npmInstall.stderr.on('data', function(data) {
+    console.error(data.toString());
+  });
```
nicolodavis: The `npm install` subprocess seems to be silent for me (i.e. I can't see any output from it so it just appears that the terminal is hanging while it is doing the install).  Perhaps we're not subscribing to the data stream correctly?

scripts/install-web.js
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable unicorn/no-process-exit */
```
nicolodavis: Can you add the copyright notice here like we have for the other JS files in the repo?

scripts/install-web.js
delucis: I’m not 100% sure of best practices, but if the aim is just to let `npm install` run and display it’s output, I think the simplest is using the `stdio` option:

```js
const npmInstall = spawn(
  'npm',
  ['install'],
  {
    cwd: webExamplePath,
    stdio: 'inherit'
  }
); 
```

This outputs everything straight to the parent `stdout`.

delucis: Node docs: https://nodejs.org/dist/latest-v12.x/docs/api/child_process.html#child_process_options_stdio

tisoap: Thanks @delucis , using `inherit` solved the issue!

tisoap:
Got a hold on a Windows machine again and updated the install script. On Windows the install command is slightly different, so I had to add a check for that.

Issue #486, opened 2019-10-02
moves not changing G if in stage


MichalPP:
I have a very simple game: roll dice (1-6) then move by 1-6 positions.

I created two moves (rollDice and move) and two stages (stageDice and stageMove). works fine.

however, if I restrict stages to use only some moves, the moves don't work. so

`turn: { stages: stageMove{  } }`
works fine (G is changing), but
`turn: { stages: stageMove{ moves: { move: G => G } } }`
does not (G remains unchanged).

any thoughts?

nicolodavis:
It would be much easier to debug if you provide a https://codesandbox.io/ link.

delucis:
@MichalPP Do you have some more detailed code? The following should work:

```js
turn: {
  stages: {
    stageDice: {
      moves: {
        roll: (G, ctx) => {
          G.rollResult = ctx.random.D6()
        }
      }
    },
    stageMove: {
      moves: {
        move: G => {
          G.position = G.position + G.rollResult // or whatever
        }
      },
    },
  }
}
```

You do need to make sure the player is actually in one of the stages, otherwise there are no moves for them to make. The following would set the player to `stageDice` at the start of each turn:

```js
turn {
    activePlayers: { player: 'stageDice' }
}
```

MichalPP:
I had to keep the moves in moves section (I reuse them in different stages). I suspect the form of restricting the moves in stages is causing problems (@delucis  format works): 

```
moves: {
    moveRollDice: (G, ctx, id) => {
        G.dice=ctx.random.D6();
    },
turn: {
    stages: {
        rollDice: { moves: { moveRollDice: G => G }
   }
}
```

If you don't see any error in the above syntax, I will strip out a minimalistic demo of my game (currently it's bit longer).

delucis:
If you need the moves for different stages, the best approach is to declare the function once and re-use it:

```js
const moveRollDice = (G, ctx) => {
  G.dice = ctx.random.D6();
}

const game = {
  moves: { moveRollDice },
  turn: {
    stages: {
      rollDice: {
        moves: { moveRollDice },
      }
    }
  }
}
```

If you’re still having trouble, a minimal demo would be great.

MichalPP:
@delucis thanks, it's working like charm.

I will create some PR for documentation.

nicolodavis:
What specifically is missing in the documentation that you want to fix?

MichalPP:
[...]

PR #491, opened 2019-10-07
clarification of cards moves

MichalPP:


nicolodavis:
Based on the discussion in #486, it sounds like this is not the change that we want.  Closing this.

Issue #477, opened 2019-09-27
Implement a Pass event


nicolodavis:
This will be simple event that merely ends the turn and also removes the player from `ctx.playOrder`.

blunket:
Could this also be used as a "surrender" event?

Or can we make the turnOrder removal optional and/or add a callback for after this?

nicolodavis:
I think a surrender event might be better implemented as a plugin because it needs to store state (while the pass event is stateless).

nicolodavis:
What do you mean by making the removal from playOrder optional?  I would think that it is necessary to do that to implement a pass move, no?

danielimmke:
I'd like to work on this. I've been onboarding onto the codebase and think I could have a PR within the week for it.

nicolodavis:
Thanks @danielimmke.  Looking forward to it.

danielimmke:
Hey @nicolodavis thanks for replying. I have a very basic version of this working but wanted to make sure I 100% understood the requirements before submitting a PR. A "Pass" to me is just the same as ending a turn without doing anything. So I created a new event but copied the EndTurn method and removed the check that stops EndTurn from happening (around the comment `// Prevent ending the turn if moveLimit haven't been made.`)

This seems to work in my testing. It increments the turn and switches the player. I wanted to make sure I wasn't missing something. The other question I had was, it seems like a move calls `endTurn` automatically at the end. Could a user implement this functionality right now by just creating a move called `pass` that doesn't do anything? Are we just formalizing that as an event?

nicolodavis:
@danielimmke A pass is similar to a no-op move that merely calls `endTurn` (and you can in fact create a move to do exactly this).  However, it also accomplishes another important thing which is to remove the player from `playOrder`.  This will cause the turn to skip over the player the next time it comes around.  It's also possible to simulate this using a move and a custom turn order, but we're merely formalizing this into a convenient event that makes it easier for users.

Now that I think about it, it probably makes sense to support two behaviors (one that removes the player from the turn order and one that does not):

#### 1. `pass()`

Identical to `endTurn()`.

#### 2. `pass({ remove: true })`

Also removes the player from `playOrder`.  Once `playOrder` is empty, we should automatically end the phase.

As far as the implementation is concerned, I would recommend just calling `endTurn` directly from the "pass" event rather than duplicating any logic.

PR #492, opened 2019-10-13
Add pass event

danielimmke:
#### Checklist

For issue #477 - I did my best based on my understanding of the codebase. It appears to work well. I understand there's probably going to be some changes needed before it can be moved in but I am happy to make some adjustments if needed.


src/core/flow.js
```diff
@@ -529,6 +529,19 @@ export function Flow({ moves, phases, endIf, turn, events, plugins }) {
     // Reset activePlayers.
     ctx = { ...ctx, activePlayers: null };
 
+    // Remove player from playerOrder
+    if (arg && true === arg.remove) {
+      const playOrder = [...ctx.playOrder];
+
+      playOrder.splice(playOrder.indexOf(ctx.currentPlayer), 1);
```
nicolodavis: This can be more efficiently written as:
```
const playOrder = playOrder.filter(i => i != ctx.currentPlayer);
ctx = { ...ctx, playOrder };
```

src/core/flow.js
```diff
@@ -529,6 +529,19 @@ export function Flow({ moves, phases, endIf, turn, events, plugins }) {
...
+
+      ctx = { ...ctx, playOrder };
+
+      if (0 === playOrder.length) {
+        EndPhaseEvent(state);
```
nicolodavis: `state = EndPhaseEvent(state)`

src/core/flow.js
nicolodavis: However, you should actually be using `playerID` (the player that called the event) rather than `currentPlayer`.  This value is passed around to all events and is available if you just extract it (similar to `EndStage` on line 563).

danielimmke:
Hey Nicolo, thank you for the very quick review time! I have a new commit with the changes you requested please let me know what you think.

nicolodavis:
Looks good!  Can you add some unit tests in `flow.test.js` demonstrating the usage of the Pass event?

danielimmke:
Sure thing, I will try to have that this week.

danielimmke:
Hey Nicolo,

So I basically copied over the tests for endTurn and removed some items to prevent redundancy (since `pass` calls the endTurn method under the hood and added a specific test for the `remove` parameter. Seems to work!

src/core/flow.js
```diff
@@ -529,6 +529,18 @@ export function Flow({ moves, phases, endIf, turn, events, plugins }) {
...
 
+    // Remove player from playerOrder
+    playerID = playerID || state.ctx.currentPlayer;
+
+    if (arg && true === arg.remove) {
+      const playOrder = [...ctx.playOrder].filter(i => i != playerID);
+      ctx = { ...ctx, playOrder };
+
+      if (0 === playOrder.length) {
+        state = EndPhaseEvent(state);
```
nicolodavis: Could you add a test that also captures ending the phase?

danielimmke:
Ok - I got that fixed. I think I broke my personal record for "most amount of time spent debugging for something that turned out to be a one line code change" - lol.

I also added an additional test to check that scenario.

src/core/flow.js
```diff
@@ -529,6 +529,18 @@ export function Flow({ moves, phases, endIf, turn, events, plugins }) {
...
 
+    // Remove player from playerOrder
+    playerID = playerID || state.ctx.currentPlayer;
+
+    if (arg && true === arg.remove) {
+      const playOrder = [...ctx.playOrder].filter(i => i != playerID);
+      ctx = { ...ctx, playOrder };
+
+      if (0 === playOrder.length) {
+        return EndPhaseEvent(state);
```
nicolodavis: Instead of calling `EndPhaseEvent` directly here, you should queue it for processing by adding it to `next`.  See line 377 for an example of how to do this.

src/core/turn-order.js
```diff
@@ -259,12 +259,23 @@ export function UpdateTurnOrderState(G, ctx, turn, endTurnArg) {
...
-      currentPlayer = endTurnArg.next;
-    } else {
+    if ('object' !== typeof endTurnArg) {
       logging.error(`invalid argument to endTurn: ${endTurnArg}`);
     }
+
+    Object.keys(endTurnArg).forEach(arg => {
+      switch (arg) {
+        case 'remove':
+          currentPlayer = getCurrentPlayer(ctx.playOrder, playOrderPos) || '0';
```
nicolodavis: Would this ever fallback on the `0` at the end of the line?

src/core/flow.js
```diff
@@ -529,6 +529,18 @@ export function Flow({ moves, phases, endIf, turn, events, plugins }) {
     // Reset activePlayers.
     ctx = { ...ctx, activePlayers: null };
 
+    // Remove player from playerOrder
+    playerID = playerID || state.ctx.currentPlayer;
+
+    if (arg && true === arg.remove) {
+      const playOrder = [...ctx.playOrder].filter(i => i != playerID);
```
nicolodavis: If you don't also adjust `playOrderPos` here you might leave it out of bounds (when removing the player at the end of `playOrder`).

The following test demonstrates the problem:

```js
  test('test', () => {
    let t = state;
    t = flow.processEvent(t, gameEvent('pass'));
    t = flow.processEvent(t, gameEvent('pass'));
    t = flow.processEvent(t, gameEvent('pass', { remove: true }));
    expect(t.ctx.currentPlayer).toBe('0');   // this fails because currentPlayer is undefined here.
  });
```

src/core/turn-order.js
danielimmke: I was basically attempting to do what adjusting playOrderPos in the endTurn method would do (catch out of bounds.) Adjusting that makes this unnecessary I think.

src/core/flow.js
danielimmke: Good idea. Added my idea about it and added that test as well for posterity. Thanks for the review.

nicolodavis:
Nice job @danielimmke!

Mohre:
Hi. Where can i find the documentation of this feature?

PR #495, opened 2019-10-16
Enable a function to be used for objective weights in MCTS bot

danmkent:
Objectives that can return their weight based on the game state are much more useful than those with fixed weights (a simple example is an objective for maximising the player's score in games that have a score)

Example usage:

```
const objectives = () => ({
    'score': {
        checker: () => true,
        weight: (G, ctx) => (G.scores[parseInt(ctx.currentPlayer)])
    }
});
```

#### Checklist

### DIFF
--- a/src/ai/bot.test.js
+++ b/src/ai/bot.test.js
@@ -282,6 +282,31 @@ describe('MCTSBot', () => {
     }
   });
 
+  test('objectivesWithWeightFunction', () => {
+    const objectives = () => ({
+      'play-on-square-0': {
+        checker: G => G.cells[0] !== null,
+        weight: () => 10.0,
+      },
+    });
+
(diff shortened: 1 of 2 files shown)


src/ai/bot.test.js
```diff
@@ -282,6 +282,31 @@ describe('MCTSBot', () => {
     }
   });
 
+  test('objectivesWithWeightFunction', () => {
```
nicolodavis: Let's club this and the test above (objectives where the weight is specified as a number) under one `describe`?

danmkent: Makes sense - I will make that change

nicolodavis:
@danmkent Closing this since there hasn't been activity for a while.  Feel free to re-open if you do get back to working on this!

Issue #418, opened 2019-06-11
[refactor] lift transport selection from _ClientImpl


amitport:
following a discussion on gitter:

We should extract transport selection logic from raw _ClientImpl. instead it should just receive `transportFactory` (or a more abstract `masterFactory`) as a parameter.

Ideally _ClientImpl would be very slim and multiplayer/local/ai selection logic would be extracted. This separation of concerns will simplify the client interface and will allow for better bundling (e.g., avoid bundling SockeIO when it is not used).

open for contribution- if somebody starts working on it please write here to avoid collisions

(should probably wait until https://github.com/nicolodavis/boardgame.io/tree/phases-overhaul is merged)

amitport:
working on this

PR #502, opened 2019-10-29
refactor transport out of clientImpl

amitport:
clientImpl now expects {multiplayer: falsy or transportFactory}
react client's usage is not affected

closes https://github.com/nicolodavis/boardgame.io/issues/418
#### Checklist

### DIFF
--- a/examples/react-web/src/chess/multiplayer.js
+++ b/examples/react-web/src/chess/multiplayer.js
@@ -8,6 +8,7 @@
 
 import React from 'react';
 import { Client } from 'boardgame.io/react';
+import { SocketIO } from 'boardgame.io/multiplayer';
 import ChessGame from './game';
 import ChessBoard from './board';
 
@@ -15,7 +16,7 @@ const hostname = window.location.hostname;
 const App = Client({
   game: ChessGame,
   board: ChessBoard,
-  multiplayer: { server: `${hostname}:8000` },
+  multiplayer: SocketIO({ server: `${hostname}:8000` }),
   debug: false,
 });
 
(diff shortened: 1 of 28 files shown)


packages/local-transport.js
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright 2017 The boardgame.io Authors
+ *
+ * Use of this source code is governed by a MIT-style
+ * license that can be found in the LICENSE file or at
+ * https://opensource.org/licenses/MIT.
+ */
+
+import { Local } from '../src/client/transport/local.js';
```
nicolodavis: Let's put all the transport implementations under one package `boardgame.io/multiplayer`.  Modern bundlers will tree shake the stuff you don't import.

src/ai/ai.test.js
```diff
@@ -84,7 +84,7 @@ const enumerate = (G, ctx, playerID) => {
 
 describe('Step', () => {
   test('advances game state', async () => {
-    const client = Client({
+    const client = new Client({
```
nicolodavis: I would prefer to keep the external interface a function (rather than a constructor).

src/client/client-legacy.js
```diff
@@ -0,0 +1,34 @@
...
+ * Use of this source code is governed by a MIT-style
+ * license that can be found in the LICENSE file or at
+ * https://opensource.org/licenses/MIT.
+ */
+
+import { error } from '../core/logger';
+import { SocketIO } from './transport/socketio';
+import { Local } from './transport/local';
+
+export function legacyCoerceMultiplayerOption(opts) {
```
nicolodavis: I'd say let's get rid of this (i.e. force all users to explicitly name the transport).

PR #507, opened 2019-10-31
remove unused 'debug' prop from WrappedBoard

amitport:
#### Checklist

### DIFF
--- a/src/client/react.js
+++ b/src/client/react.js
@@ -65,15 +65,12 @@ export function Client(opts) {
       // This client's authentication credentials.
       // Only relevant in multiplayer.
       credentials: PropTypes.string,
-      // Enable / disable the Debug UI.
-      debug: PropTypes.any,
     };
 
     static defaultProps = {
       gameID: 'default',
       playerID: null,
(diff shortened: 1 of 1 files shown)


src/client/react.js
```diff
@@ -65,15 +65,12 @@ export function Client(opts) {
       // This client's authentication credentials.
       // Only relevant in multiplayer.
       credentials: PropTypes.string,
-      // Enable / disable the Debug UI.
-      debug: PropTypes.any,
```
nicolodavis: If the user has multiple players on the same screen, they might want a way to enable the debug panel for a specific player.  How about we pass this prop to RawClient below and remove the `debug` option on top instead?

We could also have both:
1. if `debug` is false, then there is no Debug Panel regardless of the state of the prop.
2. Otherwise, the prop can turn on / off the panel for individual players.

That way there is a convenient way to turn off the panel for all players with a single option, and also a more granular prop level option.

amitport:
submitted https://github.com/nicolodavis/boardgame.io/pull/509

PR #508, opened 2019-10-31
show debug only if it has a truthy value

amitport:
(I'm assuming this is the intention)
(otherwise debug 0 or undefined or null also causes the debug panel to display)

#### Checklist


nicolodavis:
Actually the intention is to:
1. Disable it in production builds (this is why it is guarded by process.env).
2. Enable it by default in non-production while requiring you to explicitly disable it.

I think if we also set the default value of `this.debug` to `true`, then we can have both.

PR #510, opened 2019-11-01
lobby auto refresh + leave game ready to play

francoijs:
A few improvements to lobby as listed in #354:
 - auto refresh of rooms state
 - add button 'leave' to rooms that are full and ready to play

#### Checklist

### DIFF
--- a/src/lobby/react.js
+++ b/src/lobby/react.js
@@ -33,6 +33,7 @@ const LobbyPhases = {
  * @param {string} gameServer - Address of the game server (for example 'localhost:8001').
  *                              If not set, defaults to the server that served the page.
  * @param {function} clientFactory - Function that is used to create the game clients.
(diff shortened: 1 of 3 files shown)


nicolodavis:
Thanks @francoijs!

Haven't had a chance to look at this PR up close yet, but I'm wondering if we should be directing our effort more toward migrating the Lobby to Svelte rather than add features to the existing React codebase that is going away?  More features here will also make the migration harder because there is more code to port over.

francoijs:
Yes that makes perfect sense!
The PR does not add much code though, it's mainly 2 features that were blatantly missing and refered to in several issues. There is a bit of refactoring that make things cleaner and should not impede the migration...
I did not take the time yet to look into Svelte. I'd be glad to help though!

nicolodavis:
OK, will merge this in and we can migrate it to Svelte later.

Issue #487, opened 2019-10-03
setActivePlayers function in onBegin hooks only work once


Pong420:
```js
{
  turn: {
    onBegin: (G, { numPlayers, currentPlayer, events }) => {
      const value = {};
      for (let i = 0; i < numPlayers; i++) {
        value[i] = currentPlayer === String(i) ? 'main' : 'others';
      }
      events.setActivePlayers({ value });
      return G;
    }
  }
};
```

in the next turn `ctx.activePlayers` will be null

delucis:
This might be related to #482. I haven’t tested if `setActivePlayers` works in `onBegin`.

That said, at least for your example, you could achieve this more simply like this:

```js
{
  turn: {
    activePlayers: { player: 'main', others: 'others' }
  }
}
```

Pong420:
it works, thanks so much

nicolodavis:
I don't think #482 is involved here (that bug only kicks in if an event does something like ending the turn (or phase) which further triggers an `onEnd` that contains another event).

@delucis has already pointed out the more idiomatic way of doing this, but I'm still curious why the original example does not work.

Pong420:
For your reference,
I have try `setActivePlayers` in below example.
```js
{
  onBegin: (G, ctx) => {
    ctx.setActivePlayers({ player: 'main', others: 'others' });
  };
}
```
The player who last takes the move in the previous phase is in `main` stage but not the `ctx.currentPlayer`

delucis:
Just a quick note to say I ran into this myself today. I had a situation where I wanted to configure turn active players conditionally, based on game state. In a simplified form:

```js
turn: {
  onBegin: (G, ctx) => {
    if (G.flag) ctx.events.setActivePlayers(/*...*/)
  }
}
```

But this always results in `ctx.activePlayers === null`

I guess this would also come up if you wanted to pass different values to `setActivePlayers` based on some game state.

I’ll try to dig into why this is happening, but if you have any ideas @nicolodavis, let me know!

delucis:
So, the events flow is a nightmare to debug, but I can confirm what @Pong420 reported, that the `playerID` used by events dispatched in `turn.onBegin` seems to be stale. If `currentPlayer` has been incremented, `playerID` here will still be `currentPlayer - 1`.

I _think_ the bug is that `playerID` here is provided by the `Events` class when it creates the event action:

[...]

Issue #482, opened 2019-09-29
events.js does not handle any events that are further generated in the code that it runs


nicolodavis:
For example:

```js
const game = {
  moves: {
    endTurnMove: (G, ctx) => {
      ctx.events.endTurn();
    },
  },
  turn: {
    onEnd: (G, ctx) => {
      ctx.events.endPhase();
    },
  },
  phases: {
    A: { start: true, next: 'B' },
    B: {},
  },
};

test('this test passes', () => {
  const client = Client({ game });
  expect(client.getState().ctx.phase).toBe('A');

  client.events.endTurn();

  expect(client.getState().ctx.currentPlayer).toBe('1');
  expect(client.getState().ctx.phase).toBe('B');
});

test('this test fails', () => {
  const client = Client({ game });
  expect(client.getState().ctx.phase).toBe('A');

  client.moves.endTurnMove();

  expect(client.getState().ctx.currentPlayer).toBe('1');
  expect(client.getState().ctx.phase).toBe('B');    // phase is still A
});
```

This is due to the fact that events are processed only [once](https://github.com/nicolodavis/boardgame.io/blob/master/src/core/reducer.js#L191).

PR #514, opened 2019-11-10
Fix playerID passed to events called from hooks

delucis:
~This is not complete yet and I’m not sure to what extent this makes sense, but~ this is in response to #487.

I added tests for using `events.setActivePlayers` in `onBegin` hooks. Phase `onBegin` hooks appeared to be working as expected, but turn `onBegin` hooks weren’t. I have added a change that fixes the turn hooks: `StartTurn` calls into the `Events` instance to update its internal `playerID` to the new current player.

**N.B.** Events called in hooks still suffer from the bug described in #482 if the turn or phase was ended from a move.

#### Checklist

### DIFF
--- a/src/core/events.js
+++ b/src/core/events.js
@@ -32,6 +32,8 @@ export class Events {
       };
     }
 
+    events._setPlayerID = playerID => (this.playerID = playerID);
+
     return { ...ctx, events };
   }
 
--- a/src/core/flow.js
+++ b/src/core/flow.js
@@ -308,7 +308,6 @@ export function Flow({ moves, phases, endIf, turn, events, plugins }) {
 
     // Initialize the turn order state.
     if (currentPlayer) {
-      if (ctx.events) ctx.events._setPlayerID(currentPlayer);
       ctx = { ...ctx, currentPlayer };
       if (conf.turn.activePlayers) {
         ctx = SetActivePlayers(ctx, ctx.currentPlayer, conf.turn.activePlayers);
@@ -318,6 +317,8 @@ export function Flow({ moves, phases, endIf, turn, events, plugins }) {
       // when there is no currentPlayer yet.
       ctx = InitTurnOrderState(G, ctx, conf.turn);
     }
+
+    if (ctx.events) ctx.events._setPlayerID(ctx.currentPlayer);
 
     G = conf.turn.onBegin(G, ctx);
 
--- a/src/core/flow.js
+++ b/src/core/flow.js
@@ -523,6 +523,8 @@ export function Flow({ moves, phases, endIf, turn, events, plugins }) {
       return state;
     }
 
+    if (ctx.events) ctx.events._setPlayerID(ctx.currentPlayer);
+
     // Run turn-end triggers.
     G = conf.turn.onEnd(G, ctx);
 
--- a/src/core/flow.js
+++ b/src/core/flow.js
@@ -308,6 +308,7 @@ export function Flow({ moves, phases, endIf, turn, events, plugins }) {
 
     // Initialize the turn order state.
     if (currentPlayer) {
+      if (ctx.events) ctx.events._setPlayerID(currentPlayer);
       ctx = { ...ctx, currentPlayer };
       if (conf.turn.activePlayers) {
         ctx = SetActivePlayers(ctx, ctx.currentPlayer, conf.turn.activePlayers);
--- a/src/core/flow.test.js
+++ b/src/core/flow.test.js
@@ -902,4 +902,81 @@ describe('activePlayers', () => {
       '2': 'B',
     });
   });
+
+  test('setActivePlayers in turn.onBegin works', () => {
+    const game = {
+      turn: {
+        onBegin: (G, ctx) => {
+          console.log('turn.onBegin', ctx.currentPlayer);
+          ctx.events.setActivePlayers({ player: 'A', others: 'B' });
+        },
+        stages: { A: {}, B: {} },
+      },
+    };
+
+    const client = Client({ game, numPlayers: 3 });
+
+    expect(client.getState().ctx.currentPlayer).toBe('0');
+    expect(client.getState().ctx.activePlayers).toEqual({
+      '0': 'A',
+      '1': 'B',
+      '2': 'B',
+    });
+
+    console.log('end turn');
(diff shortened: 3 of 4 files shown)


delucis:
Here are some clarifications on this PR after I thought about it a bunch more. I instinctively wasn’t happy with this solution, but I _think_ it (or something pretty similar) is probably the appropriate solution.

### The bug

- The events that depend on `playerID` for their effect are `endStage`, `setStage`, and `setActivePlayers`.

- When called from hooks, I think we would expect these events to use the following player IDs (at least when the hook starts):

    |                | `playerID` | `currentPlayer`
    ---------:|:---------:|:-------------:
    `onBegin` |                   |          ✓
    `onMove` |        ✓         |
    `onEnd`   |                    |          ✓

    > (Stage-related events are currently the only events that depend on `playerID` and these are basically pointless in `onEnd` hooks as stages will be wiped clean in the new turn, so this is kind of hypothetical, but for consistency this should probably be true.)

- Currently, `Events` receives and stores a `playerID` when it is instantiated, passing this to any event functions it manages. Events called from `onEnd` hooks will receive the `playerID` of the player whose action ended the turn, which may or may not be `currentPlayer`. Events in `onBegin` hooks will similarly see the stale `playerID` of the player who ended the previous turn.


### This PR

So far, this PR does the following:

1. Adds a `_setPlayerID` method to the `Events` class

2. Calls `events._setPlayerID` from `StartTurn` and `EndTurn` with `currentPlayer`

This makes sure that events in `onBegin` and `onEnd` hooks use the expected player ID. All tests are passing, so I’m assuming this doesn’t have any adverse side effects, but maybe there’s something about the event queue, I’ve missed.

My main concern is that this adds a method to the `Events` class that in theory could be called by external code. @nicolodavis Do you think there’s a good way to protect the `_setPlayerID` method? Or another model for achieving this behaviour?

nicolodavis:
Sorry that I've taken a while to get to this PR.  Will hopefully get some time this week to take a look.

nicolodavis:
Thanks for your patience and spot-on insights while debugging this!  I agree that the `_setPlayerID` method isn't ideal.

#### A different idea

Instead of trying to provide a method to override the internal state of the `Events` class (and having it behave differently depending on whether it is called from `onBegin` or `onMove`, why don't we make the `setActivePlayers` API explicitly declare which "player" it is referring to.

For example,
```
{ player: 'A' }
```
sets the stage of the player that called the event.

```
{ currentPlayer: 'A' }
```
sets the stage of the current player.

#### What about `others`?

`others` takes the set difference of *all players - players that have had their stages set in this call* and applies the argument to them.  For example:

```
{ player: 'A', others: 'B' }
```
and
```
{ currentPlayer: 'A', others: 'B' }
```

`others` in the examples above mean `all - player` and `all - currentPlayer` respectively.

This will also allow doing things like this (that are not possible currently):
```
{
  value: { '0': 'A', '1': 'B' },
  others: 'C',
}
```

delucis:
So in practice: `player` in an `onBegin` hook, for example, would be the player who ended the previous turn and `currentPlayer` would be the new current player? I guess that is an option and documenting it would just need to clarify that `player` can leak over into a new turn.

I’ll try to build out what you propose and see how it feels.

nicolodavis:
Yes, we'll have to document that `player` always refers to the player that initiated the action that resulting in the game moving to this `onBegin` hook whereas `currentPlayer` is the player that owns the current turn.

delucis:
Just came up against a few arguments you didn’t account for: 

1. `{ others: 'A' }`

2. `{ player: 'A', currentPlayer: 'B' }` where `player === currentPlayer`

### `{ others: 'A' }`

Previously, this would set everyone except `player`. Based on your suggestion, I was implementing this with `if(!(id in activePlayers))`, but for an argument that _only_ includes `others`, that will set all players, breaking the previous behaviour.

If `player` and `currentPlayer` aren’t set, how do we decide who the others are? Or, more importantly, how do we exclude players from `others` so that they are not in a stage? 

Should there be a special constant that can be used to exclude players?

```js
setActivePlayers({ currentPlayer: Stage.NONE, others: 'A' })
```

Or another syntax? (This seems a bad idea.)

```js
setActivePlayers({ others: 'A', exclude: [ 'player', 'currentPlayer' ]})
```

In both of the above cases, `all` and `others` start to get rather close to one another.

Or should `others` always exclude both `player` _and_ `currentPlayer`?

```js
setActivePlayers({ others: 'A' }) // currentPlayer and player not in a stage
```

This seems like the cleanest, but could be a source of confusion given how `player` and `currentPlayer` can overlap.

Or should `others` only have this special behaviour of excluding `player` and `currentPlayer` _only_ when no other argument is active? Would allow combinations with `value`, but again, could be confusing to have `others` change behaviour in this way.

### `{ player: 'A', currentPlayer: 'B' }`

If `player` and `currentPlayer` refer to the same player, how should differing stage arguments be resolved?

There’s an argument that the main responsibility will be on the user to not call `setActivePlayers` with conflicting arguments like this (i.e. to do so only when it’s not the current player triggering it), but we’ll still need to decide which takes precedence in these cases and if we should log a warning or something.

nicolodavis:
Ah yes, I think my idea makes things more confusing.

What do you think of this then:
Let's just support `currentPlayer` and drop `player` completely.  The user can still use `ctx.playerID` and `value` to achieve what `player` attempted to provide and the semantics become much clearer now (including avoiding the `others` problem).

`currentPlayer` is almost always what you want in the `activePlayers` section when used declaratively, so I think it's not too much of a burden to have the user use `value` when calling `setActivePlayers`.

nicolodavis:
```js
const value = {};
players.forEach(p => {
  if (p === ctx.playerID) value[p] = 'A';
  else value[p] = 'B';
});
ctx.events.setActivePlayers({ value });
```

This is an example snippet of what they might do to simulate `{ player: 'A', others: 'B' }`.

delucis:
Actually, because `value` is applied last, it’s even simpler:

```js
ctx.events.setActivePlayers({ others: 'B', value: { [ctx.playerID]: 'A' } })
```

nicolodavis:
Ah perfect!

nicolodavis:
Closing this now that we have #523

PR #523, opened 2019-11-27
Replace `player` with `currentPlayer` option in `setActivePlayers`

delucis:
Supersedes #514, related to #487.

In order to clarify who is being impacted by `setActivePlayers` calls — especially in hooks — this changes the syntax to require a `currentPlayer` option instead of a `player` option:

### Breaking Changes

Anyone currently using the `setActivePlayers` event or the `turn.activePlayers` option in their game config, who is using a `player` option may see breaking changes.

Migrating will take one of two forms:

1. #### Switching to using `currentPlayer`

    For `turn.activePlayers` or `setActivePlayers` calls where `player` was always the current player in any case, users will only have to rename the option:

    ```diff
    activePlayers: {
    - player: 'A',
    + currentPlayer: 'A'
    }
    ```

2. #### Maintaining `player` behaviour

    For `setActivePlayers` calls where the behaviour impacted the player who triggered the move/event specifically, users will need to switch to using the `value` option, for example:

    ```diff
    ctx.events.setActivePlayers:({
    - player: 'A',
    + value: { [ctx.playerID]: 'A' }
    })
    ```

### Further considerations

`setStage` and `endStage` still rely on `playerID`, which is unavoidable. These should probably be disabled in `onBegin` and `onEnd` hooks for clarity’s sake.

### DIFF
--- a/src/core/turn-order.js
+++ b/src/core/turn-order.js
@@ -32,11 +32,11 @@ export const Pass = (G, ctx) => {
 /**
  * Event to change the active players (and their stages) in the current turn.
  */
-export function SetActivePlayersEvent(state, playerID, arg) {
-  return { ...state, ctx: SetActivePlayers(state.ctx, playerID, arg) };
+export function SetActivePlayersEvent(state, _playerID, arg) {
+  return { ...state, ctx: SetActivePlayers(state.ctx, arg) };
 }
 
-export function SetActivePlayers(ctx, playerID, arg) {
+export function SetActivePlayers(ctx, arg) {
   let { _prevActivePlayers } = ctx;
 
   const _nextActivePlayers = arg.next || null;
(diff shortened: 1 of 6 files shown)


src/core/turn-order.js
```diff
@@ -32,11 +32,11 @@ export const Pass = (G, ctx) => {
 /**
  * Event to change the active players (and their stages) in the current turn.
  */
-export function SetActivePlayersEvent(state, playerID, arg) {
-  return { ...state, ctx: SetActivePlayers(state.ctx, playerID, arg) };
+export function SetActivePlayersEvent(state, _playerID, arg) {
```
nicolodavis: Do we need the `_playerID` argument here?

nicolodavis: Oh yeah we do because all events receive it.

PR #531, opened 2019-12-14
fix(server): Use namespaced ID to delete persisted game data

delucis:
The intended behaviour of the server is to delete persisted game data when all players leave the game. When deleting the game data, the database implementation is currently passed the wrong ID, so the game data is never deleted, only the metadata. This PR fixes this bug.


nicolodavis:
Looks like there is a long error.

delucis:
`api.test.js` was testing the implementation and expected the remove method to have been called with a specific ID. I’ve updated the test to match the metadata test. (And now there’s a lint error, which serves me right for making changes directly on GitHub instead of in a text editor 😂.)

Issue #429, opened 2019-07-01
Lobby vulnerabilities


flamecoals:
I was looking at the code while developing a PR, and I think the Lobby endpoints have several vulnerabilities:

## Leaking credentials

The credentials are initialized by playerID when the room is created, and does not change if hte user leaves or joins the room. So if a user leaves the room, they could possibly keep the credential, and when another user joins that room in their place, the previous user could play on behalf of the new user. 

## Reflective XSS for instance here:
```
router.get('/games/:name/:id', async ctx => {
 const gameID = ctx.params.id;
 const room = await db.get(`${gameName}:${GameMetadataKey(gameID)}`);
  if (!room) {
    ctx.throw(404, 'Room ' + gameID + ' not found');
```

if gameID is a javascript, then this javascript will be able to execute in the user browsers.

## XSRF
It seems like it has some protection with API_SECRET, but this is very weak as it is a global for the whole server. An attacker can just check what API_SECRET is being returned by the frontend once and a server restart would be needed to change the credential. Also, this option is buried deep inside the code and not documented anywhere.

## Race conditions
The "db" API has no interface for transactions or atomic checks, which leads to possible race conditions. Theoretically we could have two HTTP requests changing the state at the same time, and both executes their "checks" at the same time without seeing the change of each other states. This will definitely be a problem with several concurrent users, and specially if we need to scale to multiple servers accessing the database in parallel.

flamecoals:
I will spend some time improving the lobby in general. I want to refactor the code too, and add more features.

nicolodavis:
Thanks for looking into this @flamecoals!

nicolodavis:
Should we look into [JSON Web Tokens](https://jwt.io/) for auth stuff in general?

delucis:
JWT could be interesting. For now, might the following make sense to prevent leaking credentials?

1. Set seat credentials when a player _joins_ a game, instead of initialising them all when the game is created.

2. By default generate credentials using `shortid` as currently, but allow the credentials function to be customised, calling it with relevant request information. You can already customise the ID generator using `lobbyConfig.uuid`, but it might be helpful to separate ID generation from credential generation.

nicolodavis:
@delucis Sounds good!

delucis:
Cool. It'll probably take me a little while to get to all of this, but I've found another bug with the database caching that's an easy fix, so I'll take care of that in the next couple of days.

nicolodavis:
[...]

PR #532, opened 2019-12-20
feat(server): Lobby server improvements

delucis:
Addresses some of the issues raised by #429 

1. Creates seat credentials when a player joins a game, deleting them when they leave, rather than instantiating them all once when a game is created. This stops credentials leaking as described by @flamecoals.

2. Enables some basic customisation of authentication by adding `generateCredentials` and `authenticateCredentials` options to the Lobby server.

    If provided, `generateCredentials` can be called in place of `shortid` and receives the Koa `ctx` object, so request headers etc. can be used for credential generation. This can be an async function if needed.

    `authenticateCredentials` receives a simplified signature of `(credentials, playerMetadata)` when called from the server master (compared to `{ action, gameMetadata, gameID, playerID }` used internally). In non-synchronous masters, `authenticateCredentials` can also be an async function to support third-party API calls.

There should be no breaking changes here. I changed the signature of the `SocketIO` server transport, but as far as I can tell, that wasn’t exposed externally in any way.


docs/documentation/api/Server.md
```diff
@@ -24,6 +24,10 @@ A config object with the following options:
 
 3. `transport` (_object_): the transport implementation.
    If not provided, socket.io is used.
+   
+4. `generateCredentials` (_function_): an optional function that returns player credentials to store in the game metadata and validate against. If not specified, the Lobby’s `uuid` implementation will be used.
+
+5. `authenticateCredentials` (_function_): an optional custom function that tests if a player’s move is made with the correct credentials when using the default socket.io transport implementation.
```
nicolodavis: Let's call this "an optional function" (remove the word "custom") to be consistent with the line above.

docs/documentation/api/Server.md
```diff
@@ -24,6 +24,10 @@ A config object with the following options:
 
 3. `transport` (_object_): the transport implementation.
    If not provided, socket.io is used.
+   
+4. `generateCredentials` (_function_): an optional function that returns player credentials to store in the game metadata and validate against. If not specified, the Lobby’s `uuid` implementation will be used.
```
nicolodavis: I feel like there is opportunity to merge `generateCredentials` and `uuid` into one thing.  Is there value in them being separate?

delucis: `uuid` is currently used to generate both player credentials and game IDs, which seem like quite separate things to me. That was fine for a context-less random generator, but I’m passing Koa’s `ctx` to `generateCredentials`, so that changes things a bit.

I also chose to have `generateCredentials` fall back first to `lobbyConfig.uuid` and then to `shortid.generate` as that preserves current behaviour, but falling back to `shortid` directly would probably make more sense. (The `lobbyConfig.uuid` option was misdocumented as `shortid` before, so I’m guessing there probably wouldn’t be too many people impacted if that changed.)

docs/documentation/api/Server.md
delucis: Resolved in d07c8b2

docs/documentation/api/Server.md
nicolodavis: Sorry if I missed some context in the original issue discussing this, but what's the use-case for having a customized function to generate credentials that isn't just invoking something like `uuid`?

flamecoals:
I think previously the default unique ID was very long, so it made bad
links. I asked to change to a shorter one, which at the time we decided to
open up for customization. My only requirement is for it to be short and
preferably URL friendly

docs/documentation/api/Server.md
delucis: The example I gave in the docs is of calling into an authentication service.

For example, Firebase Auth allows you to get an ID token for a logged in user on the client. You could then send that in the join game request, have `generateCredentials` decode the token, and return the decoded user ID (which is public) to be stored in the metadata. Then when making a move, the client could send along a similar (potentially refreshed) token and the `authenticateCredentials` method could again decode that token, and check if it matches the user ID stored in the metadata.

That could provide a way for users to authenticate that doesn’t rely on session storage.

delucis:
I hadn’t thought about URLs. I have seen some people providing links to share for each seat when a game is created. I guess these changes would remove the credential leaking, but might prevent people from doing that kind of thing?

flamecoals:
For freeboardgame.org, if I remember correctly, we share only the room id
and the server assigns the seat on the fly. But I need to review it to be
sure

docs/documentation/api/Server.md
nicolodavis: Makes sense.  Thanks for clarifying!

nicolodavis:
> I hadn’t thought about URLs. I have seen some people providing links to share for each seat when a game is created. I guess these changes would remove the credential leaking, but might prevent people from doing that kind of thing?

Why would it prevent people from creating links to join specific seats?  If `generateCredentials` returns something that is URL friendly, it should still be possible, no?

delucis:
> If `generateCredentials` returns something that is URL friendly, it should still be possible, no?

Absolutely, but the credentials would have to be created one-by-one for each seat via the `join` endpoint (because they’re created on join, rather than on create game). That said, at the moment, you can only get a seat’s credentials by joining it, so presumably people are already “joining” several times to get the credentials and then displaying them all to a user.

nicolodavis:
@delucis Yeah, we'll have to figure out what's the best way to support joining specific seats via a URL (I think that's a nice feature to have).  That's beyond the scope of this PR of course, so let's get this in first.

src/server/api.ts
```diff
@@ -170,7 +180,8 @@ export const addApiToServer = ({ app, db, games, lobbyConfig }) => {
     }
 
     gameMetadata.players[playerID].name = playerName;
-    const playerCredentials = gameMetadata.players[playerID].credentials;
+    const playerCredentials = await lobbyConfig.generateCredentials(ctx);
```
nicolodavis: Should we also pass `playerID` to `generateCredentials` (in case it wants to do something player specific)?

delucis: Potentially. It is in theory already available as `ctx.request.body.playerID`, but there’s probably an argument for some handy aliases. Anything else you can think of that might be useful?

nicolodavis: That's the only one that comes to mind at the moment.

nicolodavis: I'm also happy leaving it out at the moment until we see a use case for it.  I'll let you make the call.

nicolodavis:
@delucis What's the next step here?  Just double-checking that this is not waiting on me.

delucis:
Sorry, dropped the ball on this one. It’s basically ready.

The one thing I’ve been indecisive on is the signature of the new `generateCredentials` and `authenticateCredentials` methods. The first just gets the raw Koa `ctx`, while the latter gets `credentials` from the player’s action payload and `playerMetadata`, which is the stored metadata for the player ID making the move. In other words, the first is very general (to allow use of custom headers in the `join` request), while the second is locked to some specific aliases (to hide the implementation complexity).

I kind of don’t like that these aren’t balanced and worry if they’re too specific to my experience. I forked a version of the server and integrated with Firebase Auth, but am not sure how helpful this `generate`/`authenticate` pattern would be for other services, because I don’t have experience using them.

In other words, I think it’s good to go and the fix for credential leaking is a definite improvement, I’m just fretting about details 😃

nicolodavis:
That's fine.  We can change the signatures if we see a future use-case that would benefit from something a bit different.

src/server/index.ts
```diff
@@ -33,14 +33,32 @@ export const createServerRunConfig = (portOrConfig?: any, callback?: any) => {
   return config;
 };
 
+/**
+ * Wrap a user-provided auth function to simplify external API
+ * @param  {function} fn The authentication function to wrap
+ * @return {function} Wrapped function for use by master
+ */
+const wrapAuthFn = fn => ({ action, gameMetadata, playerID }) =>
+  fn(action.payload.credentials, gameMetadata[playerID]);
```
nicolodavis: Could you add a test that covers this line?

delucis: Sure! Might take a moment for me to get to it, but I will.

nicolodavis: No hurry. Thanks!

PR #549, opened 2020-02-20
fix local multiplayer test example

sarumpaet:
.connect() has been renamed to .start() in 2f86d92


sarumpaet:
@googlebot I signed it!

PR #550, opened 2020-02-27
fixes build in rushjs pnpm monorepo context

ksjogo:
Not necessarily final version, but I did run into some build issues when using boardgame.io together with rushjs.io/

Namely some undeclared dependencies which were okay to having trans deps in node_modules before and rollup putting logs on stderr which prevents rush from caching.


src/client/debug/log/LogEvent.svelte
```diff
@@ -1,8 +1,8 @@
 <script>
   export let logIndex;
-  export let onLogClick;
-  export let onMouseEnter;
-  export let onMouseLeave;
+  //export let onLogClick;
+  //export let onMouseEnter;
+  //export let onMouseLeave;
```
nicolodavis: Why is this changing?

src/client/debug/main/PlayerInfo.svelte
```diff
@@ -1,7 +1,7 @@
 <script>
   export let ctx;
   export let playerID;
-  export let onClick = () => {};
+  //export let onClick = () => {};
```
nicolodavis: Why this change?

src/client/transport/socketio.js
```diff
@@ -7,9 +7,10 @@
  */
 
 import * as ActionCreators from '../../core/action-creators';
-import io from 'socket.io-client';
 import { Transport } from './transport';
 
+const io = require('socket.io-client');
```
nicolodavis: What's the purpose of this change?

package.json
```diff
@@ -8,7 +8,7 @@
     "dev": "node scripts/install-web.js && npx concurrently npm:build:watch npm:dev:server npm:dev:client",
     "dev:client": "cd examples && cd react-web && npm start",
     "dev:server": "cross-env NODE_ENV=development nodemon -w src -w examples -e js,ts --exec babel-node --extensions '.ts,.js' --ignore 'src/**/*.test.ts' --presets @babel/preset-env examples/react-web/server.js",
-    "build": "cross-env BABEL_ENV=rollup rollup --config rollup.config.js",
+    "build": "cross-env BABEL_ENV=rollup rollup --config rollup.config.js --silent",
```
nicolodavis: How are we going to catch any actual warnings that Rollup throws if they are silenced?

package.json
```diff
@@ -148,10 +148,12 @@
     "koa-socket-2": "^1.0.17",
     "lru-cache": "^4.1.1",
     "prop-types": "^15.5.10",
+    "regenerator-runtime": "^0.13.3",
```
nicolodavis: Why is `regenerator-runtime` being added?

package.json
```diff
@@ -65,77 +65,77 @@
...
     "eslint-config-prettier": "^3.1.0",
     "eslint-plugin-jest": "^21.22.1",
     "eslint-plugin-prettier": "^2.6.2",
-    "eslint-plugin-react": "^7.11.1",
+    "eslint-plugin-react": "^7.18.3",
     "eslint-plugin-unicorn": "^6.0.1",
-    "firebase": "^7.5.2",
-    "firebase-admin": "^8.5.0",
-    "firebase-mock": "^2.2.10",
-    "graceful-fs": "https://github.com/mekwall/node-graceful-fs.git#patch-1",
```
nicolodavis: There are more changes to this file than are necessary to fix the problem at hand.  As such, I'm not comfortable approving this PR unless it is broken into smaller simpler changes.

Would merely removing `graceful-fs` be sufficient?  Is there a need to upgrade everything else?  (not opposed to upgrading them in a separate PR of course).

src/client/debug/log/LogEvent.svelte
ksjogo: They are unused and thus the svelte compiler warns on build.

src/client/debug/main/PlayerInfo.svelte
ksjogo: Same here.

src/client/transport/socketio.js
ksjogo: The problem was that the socket.io-client library is not having a proper default export which  might fail in some nested bundling contexts. Didn't find a better way for that for now.

package.json
ksjogo: Actual warnings will  still be on stdout. Just not the build info of what got build (which weirdly is also on stderr).

package.json
ksjogo: The build process was erring out without. Not sure what exactly requires it.

package.json
ksjogo: Oh, looks like I ran an npm upgrade too much. Yeah, I will break it down.
Just removing graceful-fs should be good enough.

src/client/debug/log/LogEvent.svelte
nicolodavis: I see.  Let's delete them, then.

ksjogo: Removed them.

ksjogo:
Would you like any further changes here for the moment?

src/client/transport/socketio.js
nicolodavis: I think this is also having the unintentional effect that `socket.io-client` no longer gets bundled into the minified bundle.  Try running `npm run build` with and without this PR you will notice a significant change in the bundle size.

nicolodavis: @ksjogo I think this is still not resolved.

ksjogo: Sorry, missed the comment, will have a look.

ksjogo: Changed it to keep the import syntax, builds fine with pnpm and npm now. And diff size seems roughly the same before/after.

package.json
nicolodavis: Hi @ksjogo.  Can you double-check this again?

What's the error message that you get when regenerator-runtime is not added?

ksjogo: Without that line I get these errors in a gatsby project depending on boardgame.io
```
ERROR #98123  WEBPACK

Generating JavaScript bundles failed

Can't resolve 'regenerator-runtime/runtime' in '/Users/johannes/dev/ksjogo/packages/boardgame.io/dist'

File: ../../packages/boardgame.io/dist/boardgameio.es.js


 ERROR #98123  WEBPACK

Generating JavaScript bundles failed

Can't resolve 'regenerator-runtime/runtime' in '/Users/johannes/dev/ksjogo/packages/boardgame.io/dist/esm'

File: ../../packages/boardgame.io/dist/esm/ai-601d4764.js


 ERROR #98123  WEBPACK

Generating JavaScript bundles failed

Can't resolve 'regenerator-runtime/runtime' in '/Users/johannes/dev/ksjogo/packages/boardgame.io/dist/esm'

File: ../../packages/boardgame.io/dist/esm/Debug-f72fc892.js

not finished run queries - 13.777s
```

ksjogo:
Is this iteration looking okay?

nicolodavis:
Hi @ksjogo.  Sorry about the delay in getting to this.  Been a bit busy.  Will try to get this PR on next week.

ksjogo:
No worries. Thank you.

package.json
nicolodavis: Is it possible that this is due to the way Gatsby's webpack config is set up?

The reason I'm picky about this is:

1. We don't transform async functions to generators: [source](https://github.com/nicolodavis/boardgame.io/blob/master/babel.config.js#L7)

2. A simple grep in `dist/` for `regenerator` returns no results.

So, my understanding is that boardgame.io has no need for `regenerator-runtime` (it just leaves async functions untouched).  What I suspect is happening is that the Babel config in your client project is transforming async functions and that's why you see the error (which can probably be fixed easy by adding `regenerator-runtime` as a dep to your Gatsby project.

Can you try this?

nicolodavis:
Hi Johannes, sorry about the delay.  Just took a look now.

I'm still unsure about the `regenerator-runtime` and left a comment.  Everything else looks good to go.

package.json
delucis: For what it’s worth, I’ve just run into the same error when trying to create a plain JS version of the TicTacToe multiplayer example: https://codesandbox.io/s/bgio-plain-js-multiplayer-server-742oh

Looks OK on CodeSandbox (until you click and it fails to connect to the server), but when downloaded and served using Parcel I see `ReferenceError: regeneratorRuntime is not defined` in the browser console.

delucis: This does appear to be Babel-related. There’s some more info here: https://github.com/babel/babel/issues/9849 but those suggestions don’t seem to help and replicating the `"exclude": ["transform-regenerator", "transform-async-to-generator"]` config boardgame.io uses, also didn’t help.

The only thing that does work is setting `browserslist` to [`> 0.4% and not ie 11`](https://browserl.ist/?q=%3E+0.4%25+and+not+ie+11) (excluding IE11 & other browsers that don’t support async).

ksjogo: Might it be the case that your dist has rolled regenerator runtime into your dist bundle and mine hasn't?
Will have another look.

delucis: @ksjogo No, this is a problem with conflicting build targets. boardgame.io builds without transforming async functions so regenerator runtime isn't needed. However if you try to compile boardgame.io to a different set of browser targets, Babel will transform async functions to support older browsers (IE11 and a few others). Not sure what the solution here would be other than indicating no support for those older browsers.

ksjogo:
Removed the regenerator-runtime now.

ksjogo:
Updated in regards to conflicts.

PR #556, opened 2020-03-06
fix(master): Remove credentials from action payloads after use

delucis:
Credentials are only used to check if an action is authorized in the master and should not then be passed on to the reducer etc.

This contributes towards #227, because now credentials should not leak beyond the `Master.onUpdate` method and won’t end up for example in the game log.

### DIFF
--- a/src/master/master.js
+++ b/src/master/master.js
@@ -83,6 +83,18 @@ export const isActionFromAuthenticPlayer = (
 };
 
 /**
+ * Remove player credentials from action payload
+ */
+const stripCredentialsFromAction = action => {
+  if ('payload' in action && 'credentials' in action.payload) {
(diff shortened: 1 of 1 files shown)


delucis:
**Side note:** I initially tried to tackle this in the reducer before realising it would be better to do this as early as possible. In the process, I worked a little on the typing around actions, which isn’t perfect yet, but in case it is of interest: https://github.com/nicolodavis/boardgame.io/compare/master...delucis:delucis/feat/improve-types

In general, are you planning to convert more of the source to Typescript?

nicolodavis:
@delucis Thanks!  Yes, I intend to convert everything to TypeScript so feel free to send me any types that you have added.

delucis:
OK, cool, I’ll see what needs to be done to make these types more useable.

PR #559, opened 2020-03-10
[WIP] feat: Improve types

delucis:
This PR improves some of the types, in particular around actions, but also in a few other places.

It does add [`ts-toolbelt`](https://github.com/pirix-gh/ts-toolbelt) as a development dependency, to help with manipulating some of the types and avoid manually writing out all the action variations. It’s a pretty nice tool allowing you to do things like manipulating nested object types:

```ts
import { Object } from 'ts-toolbelt'

interface Obj {
  x: {
    p1: string;
    p2: number;
  }
}

type FilteredObj = Object.P.Omit<Obj, ['x', 'p1']>
// FilteredObj is { x: { p2: number } }
```

If you think that’s overkill, I can remove it and type the actions manually.

Some other notes:

- `flatted` bundles types for >1.0.2. It might be possible to upgrade from the current 0.2.3.


src/core/action-creators.ts
```diff
@@ -65,33 +80,33 @@ export const sync = (state, log) => ({
  * @param {object} state - The state to restore.
  * @param {Array} deltalog - A log delta.
  */
-export const update = (state, deltalog) => ({
-  type: Actions.UPDATE,
+export const update = (state: any, deltalog: any[]) => ({
```
nicolodavis: Let's use the `State` type instead of `any` here?  There are a few other spots in this file.

nicolodavis:
I don't mind the ts-toolbelt dep, especially if it makes life easier.

src/core/action-creators.ts
delucis: Done :+1:

nicolodavis:
You might see some conflicts with https://github.com/nicolodavis/boardgame.io/pull/560, especially in `initialize.ts` and `types.ts`.  Happy to help with merging if it would make your life easier.

Also, I think the build error you see here will go away once merged with https://github.com/nicolodavis/boardgame.io/pull/560.

delucis:
I’ll give merging a go now — I had been tinkering to find a nice way to fix that error, so it’ll be great if this fixes it. Curious to check out the new plug-in API!

delucis:
Merge worked cleanly! There’s always more work to do, but I think this can be merged for now.

PR #568, opened 2020-03-18
feat(server): Type fetch so that it returns what the opts request

delucis:
This PR uses a generic type to model how `fetch` returns a selection of data fields based on the options object.

```ts
const data = db.fetch(gameID, { log: true });
// typeof data = { log: LogEntry[] }
```

It does add a little extra complexity in places where `let` is used before the fetch is called, but I think it might be a helpful tool.

Here’s some background: https://github.com/pirix-gh/ts-toolbelt/issues/98

---

As an aside, what is the value in maintaining async and sync APIs as separate interfaces? Can’t the in-memory database be implemented with async functions even if they are internally synchronous? It would simplify the codebase in quite a few places if it didn’t have to decide whether to `await` or not.

### DIFF
--- a/src/master/master.ts
+++ b/src/master/master.ts
@@ -207,7 +207,7 @@ export class Master {
     const key = gameID;
 
     let state: State;
-    let result: StorageAPI.FetchResult;
+    let result: StorageAPI.FetchResult<{ state: true }>;
     if (IsSynchronous(this.storageAPI)) {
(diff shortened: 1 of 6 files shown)


nicolodavis:
I think I had trouble getting the client to work in a synchronous way when I tried that before, which led to lots of async code in unit tests.  It might still be possible to retire the sync interface cleanly now that the codebase has evolved a bit.

Issue #576, opened 2020-03-24
Expose Typescript types


delucis:
@nicolodavis Following on from https://github.com/nicolodavis/boardgame.io/issues/573#issuecomment-603237109

For now, the bundle doesn’t include a `boardgame.io/types` package, so I can’t do

```ts
import X from 'boardgame.io/types'
```

I found the base database type by doing this, but that seems pretty fragile:

```ts
import { Async } from 'boardgame.io/dist/types/src/server/db/base'
```

I’m not familiar yet with how best to expose and bundle types. I _think_ the `package.json` files generated for `core`, `multiplayer` etc. should include a `types` field pointing to the correct place for general consumption ([TS docs](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package)).

Then when someone does `import X from 'boardgame.io/X'` Typescript will also know where to find the types for that module. (The server module doesn’t currently follow this pattern, so it might need a different approach?)

I’m also not sure what’s the nicest way to expose `base.ts` (because it’s not part of the main public-facing API), but I think it and all the related types should be somewhere like a `server-internals` package like the `internal` package, so database connectors can use them cleanly. (The `Async` abstract class is not only its types, because it actually provides functionality.)

nicolodavis:
> I think the package.json files generated for core, multiplayer etc. should include a types field pointing to the correct place for general consumption

You're right.  The fix is relatively straightforward:

1. We need to convert everything in packages/*.js to TS.

After this, dist/types/packages will contain one *.d.ts for each subpackage.

2. We need to edit scripts/proxy-dirs.js to add a types field to the package.json.

nicolodavis:
About where to expose `base.ts`, I think an `internal` package sounds good.

nicolodavis:
Try `v0.39.2`.

`Async` should be available in `boardgame.io/internal`.  Not every sub-package will have types (`client`, for example) because we haven't converted the relevant source files to TS.

delucis:
👍 From a quick check, it looks like it works as expected for packages that have been converted, but not for `server`. What’s the reason for the `server` sub-package being provided differently from the other sub-packages?

The other thing that would be handy would be to be able to access the basic types defined in `src/types.ts`. _Maybe_ the `types` field in boardgame.io’s top-level package.json could point to `./dist/types/src/types.d.ts` and then those could be imported directly from boardgame.io:

```ts
import type { State } from 'boardgame.io'
```

[...]

PR #578, opened 2020-03-25
Typescript improvements

delucis:
This PR makes three small improvements to how types are published and contributes towards #576. (I think only server types still need to be exposed.)

1. Moved `ts-toolbelt` from `devDependencies` to `dependencies`.

    While working with 0.39.2, I got errors because `ts-toolbelt` wasn’t installed and the exposed types depended on it.

2. Added a `types` field to the main boardgame.io package.json. Now the basic types in `src/types` can be consumed more simply:

    ```ts
    import { State } from 'boardgame.io';
    ```

3. Imported and exported the storage API types (`FetchResult` etc.) in `src/types` so they can be used in third-party database connectors:

    ```ts
    import { StorageAPI } from 'boardgame.io';
    const opts: StorageAPI.FetchOpts = { state: true };
    ```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,6 +1,9 @@
 import { Object } from 'ts-toolbelt';
 import * as ActionCreators from './core/action-creators';
 import { Flow } from './core/flow';
+import * as StorageAPI from './server/db/base';
+
+export { StorageAPI };
 
 export interface State {
   G: object;
--- a/package-lock.json
+++ b/package-lock.json
@@ -14623,10 +14623,9 @@
       }
     },
     "ts-toolbelt": {
-      "version": "6.3.5",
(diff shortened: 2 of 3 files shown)


src/types.ts
```diff
@@ -1,6 +1,9 @@
 import { Object } from 'ts-toolbelt';
 import * as ActionCreators from './core/action-creators';
 import { Flow } from './core/flow';
+import * as StorageAPI from './server/db/base';
+
+export { StorageAPI };
```
nicolodavis: Maybe let's remove Async from the `internal` package since it's now available via StorageAPI?

delucis: I did consider it, but exposing it via internal does allow someone to use it when not using Typescript.

nicolodavis: I see, OK.

nicolodavis:
Released in 0.39.3.

PR #579, opened 2020-03-25
add plugin types to ctx interface

janKir:
#### Checklist

When using the typings of `Ctx` I noticed it was missing the events. After having a look into the source code, I realized the `random` and `player` plugins also enhance the `Ctx` interface.

The name plugin implies that they are not necessarily defined, so I declared them optional.
Also it seems like specific event functions can be deactivated, so I declared those functions all optional as well. This has the downside, that you need to add checks or non-null assertions to your typescript code when calling those event functions.

### DIFF
--- a/src/plugins/plugin-player.ts
+++ b/src/plugins/plugin-player.ts
@@ -7,6 +7,9 @@
  */
 
 export interface PlayerAPI {
+  state: {
+    [playerId: string]: object;
+  };
   get(): any;
   set(value: any): any;
   opponent?: {
--- a/src/plugins/plugin-player.ts
+++ b/src/plugins/plugin-player.ts
@@ -6,6 +6,15 @@
  * https://opensource.org/licenses/MIT.
  */
 
+export interface PlayerAPI {
+  get(): any;
+  set(value: any): any;
+  opponent?: {
+    get(): any;
+    set(value: any): any;
+  };
+}
+
 /**
  * Plugin that maintains state for each player in G.players.
  * During a turn, G.player will contain the object for the current player.
--- a/src/plugins/plugin-player.ts
+++ b/src/plugins/plugin-player.ts
@@ -32,19 +32,18 @@ export default {
     return { players: api.state };
   },
 
-  api: ({ ctx, data }) => {
+  api: ({ ctx, data }): PlayerAPI => {
     let state = data.players;
-    let result = { state };
 
     const get = () => {
       return data.players[ctx.currentPlayer];
     };
-    result.get = get;
 
     const set = value => {
       return (state[ctx.currentPlayer] = value);
     };
-    result.set = set;
+
+    let result: PlayerAPI = { state, get, set };
 
     if (ctx.numPlayers === 2) {
       const other = ctx.currentPlayer === '0' ? '1' : '0';
--- a/src/types.ts
+++ b/src/types.ts
@@ -2,6 +2,9 @@ import { Object } from 'ts-toolbelt';
 import * as ActionCreators from './core/action-creators';
 import { Flow } from './core/flow';
 import * as StorageAPI from './server/db/base';
+import { EventsAPI } from './plugins/plugin-events';
+import { PlayerAPI } from './plugins/plugin-player';
+import { RandomAPI } from './plugins/plugin-random';
 
 export { StorageAPI };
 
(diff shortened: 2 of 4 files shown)


janKir:
@googlebot I signed it!

src/types.ts
```diff
@@ -42,6 +42,49 @@ export interface Ctx {
...
   };
+  // enhanced by events plugin
+  events?: EventsAPI;
+  // enhanced by player plugin
+  player?: PlayerAPI;
+  // enhanced by random plugin
+  random?: RandomAPI;
+}
+
+export interface EventsAPI {
```
nicolodavis: Let's move these to the relevant spots in `src/plugins`?

delucis:
I think it might be a good idea to make some of the types generics, so that users can do things like:

```ts
let ctx: Ctx // has a default value with default plugins enabled 
type CustomCtx = Ctx<{}> // has no plugins enabled
type ExtendedCtx = Ctx<DefaultPlugins & { x: value }> // extends default
```

Similarly for `State`, which could support arguments for `G` and `Ctx`:

```ts
// CustomState['G'] will have type MyG instead of any
type CustomState = State<MyG, CustomCtx>
```

It should then be possible for users to do something like the following and type check as expected:

```ts
type MyG = { val: boolean}

const myGame: GameConfig<MyG, CustomCtx> = {
  endIf: G => G.field // error: field does not exist in type MyG
}
```

I need to check how would be best to implement this, but the types for Koa work like this.

src/types.ts
janKir: You mean all in one `src/plugins/types.ts` file? Or somehow split for each plugin?

delucis: @janKir It might be better if the interfaces are declared in the same place as the actual plugins (should make it easier to keep them in sync later), so perhaps each directly in the `plugin-events` and `plugin-random` files? Then they can be imported in `src/types`.

janKir:
@delucis I think this should work for the `Ctx` type:

```typescript
export interface CtxBase {
  numPlayers: number;
  playOrder: Array<PlayerID>;
  playOrderPos: number;
  activePlayers: null | ActivePlayers;
  currentPlayer: PlayerID;
  numMoves?: number;
  gameover?: any;
  turn: number;
  phase: string;
  _activePlayersMoveLimit?: object;
  _activePlayersNumMoves?: object;
  _prevActivePlayers?: Array<object>;
  _random?: {
    seed: string | number;
  };
}

export type Ctx<
  T = { events: EventsAPI; player: PlayerAPI; random: RandomAPI }
> = CtxBase & T;
```
However, I guess this will break typing inside `boardgame.io` where the `Ctx` interface is already used  for objects without those plugins.

src/plugins/plugin-player.ts
```diff
@@ -6,6 +6,15 @@
...
 
+export interface PlayerAPI {
+  get(): any;
+  set(value: any): any;
+  opponent?: {
+    get(): any;
+    set(value: any): any;
+  };
+}
+
```
delucis: Thanks! Looks like converting `plugin-player` to Typescript introduced some errors.

Try declaring `result` on line 34 with type `PlayerAPI` and including the get and set properties directly:

```ts
    let result: PlayerAPI = {
      state,
      get: () => {
        return data.players[ctx.currentPlayer];
      },
      set: value => {
        return (state[ctx.currentPlayer] = value);
      },
    };
```

delucis:
> I guess this will break typing inside `boardgame.io` where the `Ctx` interface is already used for objects without those plugins. Instead of `Ctx` it would be necessary to use `CtxBase` or `Ctx<{}>` there.

Yes, it’ll need a bit of reorganising. The `Ctx` type should probably be the one without plugins and we can use a `CtxWithPlugins` internally and expose something else to users:

```ts
type Ctx<Plugins extends { [plugin: string]: any } = {}> = {
  numPlayers: number;
  // etc.
} & Plugins;

type DefaultPlugins = {
  events: EventsAPI;
  player: PlayerAPI;
  random: RandomAPI;
};

type CtxWithPlugins = Ctx<DefaultPlugins>;
```

It probably still needs some wider consideration to work out how to cascade the types down to all the bits of code that need them.

src/plugins/plugin-player.ts
janKir: Should be fixed now

janKir:
> It probably still needs some wider consideration to work out how to cascade the types down to all the bits of code that need them.

Maybe keep the changes simple for this PR and work on that on a follow-up PR?

delucis:
> Maybe keep the changes simple for this PR and work on that on a follow-up PR?

Sounds good to me!

PR #581, opened 2020-03-26
refactor(db): Make log handling explicit in StorageAPI.setState

### LINKED ISSUE #577: Storage API & Logs
While implementing the new Firebase connector, I noticed that `setState` in the [flat-file](https://github.com/nicolodavis/boardgame.io/blob/2ad3a82250512a7c51c056c95004de9d8a87dd91/src/server/db/flatfile.ts#L98-L102) and [in-memory](https://github.com/nicolodavis/boardgame.io/blob/2ad3a82250512a7c51c056c95004de9d8a87dd91/src/server/db/inmemory.ts#L54-L58) implementations handles concatenating logs. I’m not really familiar with what the log is used for, but isn’t this concatenation somewhat more than a storage layer concern?

As far as I can tell, the only place where state is currently persisted that will change log values is [in the master during `onUpdate`](https://github.com/nicolodavis/boardgame.io/blob/2ad3a82250512a7c51c056c95004de9d8a87dd91/src/master/master.ts#L299-L303). Would it make sense to concatenate logs there and add a `setLog` storage API method?

delucis:
~~Move the log concatenation logic out of `setState` and into a separate `appendLog` method.~~

Add a `deltalog` parameter to the `StorageAPI.setState` method to make it more obvious that `setState` should also handle appending entries to the game log.

Closes #577.

### DIFF
--- a/src/master/master.ts
+++ b/src/master/master.ts
@@ -299,13 +299,9 @@ export class Master {
     const { deltalog, ...stateWithoutDeltalog } = state;
 
     if (IsSynchronous(this.storageAPI)) {
-      this.storageAPI.setState(key, stateWithoutDeltalog);
-      this.storageAPI.appendLog(key, deltalog);
+      this.storageAPI.setState(key, stateWithoutDeltalog, deltalog);
     } else {
-      await Promise.all([
-        this.storageAPI.setState(key, stateWithoutDeltalog),
-        this.storageAPI.appendLog(key, deltalog),
-      ]);
+      await this.storageAPI.setState(key, stateWithoutDeltalog, deltalog);
     }
   }
 
--- a/src/master/master.ts
+++ b/src/master/master.ts
@@ -296,13 +296,15 @@ export class Master {
       };
     });
 
+    const { deltalog, ...stateWithoutDeltalog } = state;
+
     if (IsSynchronous(this.storageAPI)) {
-      this.storageAPI.setState(key, state);
-      this.storageAPI.appendLog(key, state.deltalog);
+      this.storageAPI.setState(key, stateWithoutDeltalog);
+      this.storageAPI.appendLog(key, deltalog);
     } else {
       await Promise.all([
-        this.storageAPI.setState(key, state),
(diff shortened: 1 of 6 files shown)


delucis:
Two quick details:

1. This only uses `appendLog` during `onUpdate` in the game master. I think this is the only place that will produce deltalogs, but wanted to flag it just in case a `setState` call elsewhere could have been adding to the log.

2. `deltalog` is still being included in the state object sent to `setState`. It might not be necessary to store it in state, but I wasn’t sure (and it’s usually a pretty small object in any case).

nicolodavis:
`deltalog` resides in the state object primarily because reducer.ts passes around a state object.  I'd love to separate it out if it's possible to do that in a clean way, but like you pointed out, it's a pretty small object.

delucis:
OK, I’ve stripped `deltalog` from the state that gets persisted. I’m going to refrain from stripping it out elsewhere for now, because I’m not very clear where it is needed (e.g. on the client).

src/master/master.ts
```diff
@@ -296,10 +296,16 @@ export class Master {
       };
     });
 
+    const { deltalog, ...stateWithoutDeltalog } = state;
+
     if (IsSynchronous(this.storageAPI)) {
-      this.storageAPI.setState(key, state);
+      this.storageAPI.setState(key, stateWithoutDeltalog);
```
nicolodavis: If the storage implementation wants to perform this update in one atomic operation (using a transaction at the database layer), splitting this into two calls will make it hard for it to do so.  Sorry I didn't catch this earlier.

How about we pass `deltalog` as an additional parameter to the `setState` call?  That way it's a bit more explicit than just piggybacking on the `state` object.

delucis: Hmm, OK, yes that does make sense. OK I’ll revert it back to more or less what it was, but with `deltalog` as an optional argument :-)

delucis: Done in 063721f.

delucis:
> Thanks, and sorry about the additional work to revert some of the changes!

No worries — it’s probably good to keep the API surface small too.

PR #583, opened 2020-03-26
Remove namespacing in gameIDs on client side

### LINKED ISSUE #582: Bug: Client side Transport still using namespaced gameID
I experienced few problems, especially missing `gameMetadata` after updating to the recent version `v0.39.0`. I did some debugging on data being sent from Client to Server and found a bug that client and server are using different format of `gameID`.

Commit 4157bc1  removed `gameID` namespacing in form of `<gameName>:<gameID>` from server side classes, but it's still used on client. This incompatibility causes problems in syncing states for games created from Lobby API. Server saves game state with simple `<gameID>` key, but then on every sync it receives `<gameName>:<gameID>` key from client and can't find appropriate state in the Storage and creates new game instance.

pastDexter:
Fixes #582

### DIFF
--- a/src/client/react-native.test.js
+++ b/src/client/react-native.test.js
@@ -122,14 +122,14 @@ test('update gameID / playerID', () => {
   const spy2 = jest.spyOn(m, 'updatePlayerID');
   const spy3 = jest.spyOn(g, 'updateCredentials');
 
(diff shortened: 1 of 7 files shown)


pastDexter:
@googlebot I signed it!

nicolodavis:
Thanks for the fix!

PR #589, opened 2020-03-28
fix(server): Correctly wait for server.listen event

### LINKED ISSUE #584: TypeError: Cannot read property 'port' of null
Hello, 

I use nest.js and I would like to add boardgame.io to it. But when I import the package and run `server.run(8000)`, I have the following error : 

 UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'port' of null. 

I think it's related to `async app.listen` in server.ts but I don't really know how to solve the issues currently. 

My code : 

```javascript
import { NestFactory } from '@nestjs/core';
import { AppModule } from './app.module';
import {Server} from 'boardgame.io/server';
import FuckTheDealer from './game/fuck-the-dealer';
import RedisIoAdapter from './live/RedisIOAdapter';

async function bootstrap() {
  const app = await NestFactory.create(AppModule);
  app.useWebSocketAdapter(new RedisIoAdapter(app));
  app.enableCors({ credentials: true, origin: ["http://localhost:8000"] });
  await app.listen(3000);

  const server = Server({
    games : [FuckTheDealer]
  });
  const PORT = process.env.PORT || 3001;

  try {
    server.run(PORT, () => {
      console.log(`Serving at: http://localhost:${PORT}/`);
    });
  }catch(e){
    console.log("error", e);
  }


  if ((module as any).hot) {
    (module as any).hot.accept();
    (module as any).hot.dispose(
      () => {
        app.close();
        server.kill();
      }
    );
  }
}
bootstrap();

```

delucis:
This PR changes how Koa’s `app.listen` method is used in the server package. Previously, the server awaited `app.listen` and “then” logged the server port to the console. Except `listen` doesn’t return a promise, so this was potentially buggy. I’ve never had an issue with this but, as reported in #584, it seems it could create an error if `server.address()` returned `null` (presumably because it was called too early). I’ve used `app.listen`’s callback to wait for the connection to complete before logging and returning.

I’ve also added `@types` packages for the various Koa modules and shortid, strengthening the typing of the server, as well as typing its options object.

Closes #584.


delucis:
Ah, I see that there is a build error when rollup tries to bundle the server code. I’ve tried installing the suggested JSON plugin, but came up against another error. I think this is because I switched to import statements instead of require for Koa etc. so rollup is processing them differently. I don’t have any experience with rollup. Do you have any pointers on how to resolve this?

nicolodavis:
The way Rollup works, it tries to bundle in all dependencies it can find via `import` statements.  By changing the `require` statements to `import`, you're making it pull in more stuff (including Koa internals from node_modules) into the final bundle.

The [fix](https://github.com/nicolodavis/boardgame.io/pull/589/commits/ca324a2f58dc3e46ce460f249e6106718e2df43c) is to instruct Rollup to ignore certain imports.

delucis:
Ah, fantastic! Thanks for that.

PR #595, opened 2020-03-29
Include credentials in undo/redo actions

### LINKED ISSUE #593: Undo/redo actions missing credentials
If playing on a remote server, actions are validated wrt player credentials that are assumed to be inside action.payload.credentials:
https://github.com/nicolodavis/boardgame.io/blob/v0.39.3/src/master/master.ts#L188

However, in case of undo and redo, the client sends an action that consists only of type:
https://github.com/nicolodavis/boardgame.io/blob/v0.39.3/src/core/action-creators.ts#L97

Is it just me looking at the code wrong or the undo/redo actions should also include credentials?

tom3q:
If playing on a remote server, actions are validated wrt player
credentials that are assumed to be inside action.payload.credentials:
https://github.com/nicolodavis/boardgame.io/blob/v0.39.3/src/master/master.ts#L188

However, in case of undo and redo, the client sends an action that
consists only of type:
https://github.com/nicolodavis/boardgame.io/blob/v0.39.3/src/core/action-creators.ts#L97

Make it possible to undo/redo with a remote server and fix #593 by
adding credentials to undo/redo actions.

#### Checklist

* [v] Use a separate branch in your local repo (not `master`).
* [v] Test coverage is 100% (or you have a story for why it's ok).

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -199,7 +199,7 @@ export namespace CredentialedActionShape {
     | ActionShape.Reset
     | ActionShape.Undo
     | ActionShape.Redo
-    | ActionShape.Plugin;
+    | Plugin;
 }
 
 export namespace ActionShape {
--- a/src/types.ts
+++ b/src/types.ts
@@ -190,15 +190,14 @@ export namespace CredentialedActionShape {
   export type AutomaticGameEvent = ReturnType<
     typeof ActionCreators.automaticGameEvent
   >;
+  export type Undo = ReturnType<typeof ActionCreators.undo>;
+  export type Redo = ReturnType<typeof ActionCreators.redo>;
   export type Any =
     | MakeMove
     | GameEvent
     | AutomaticGameEvent
-    | ActionShape.Sync
-    | ActionShape.Update
-    | ActionShape.Reset
-    | ActionShape.Undo
-    | ActionShape.Redo
+    | Undo
+    | Redo
     | Plugin;
 }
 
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -190,6 +190,8 @@ export namespace CredentialedActionShape {
   export type AutomaticGameEvent = ReturnType<
     typeof ActionCreators.automaticGameEvent
   >;
+  export type Undo = ReturnType<typeof ActionCreators.undo>;
+  export type Redo = ReturnType<typeof ActionCreators.redo>;
   export type Any =
```
nicolodavis: Can you also use the new `Undo / Redo` types in `Any` (rather than the credential-less versions being used there currently)?

nicolodavis:
Ah, looks like this [line](https://coveralls.io/builds/29755666/source?filename=src/master/master.ts#L117) is now uncovered and doesn't have a test covering it.  Do you mind adding a test?

tom3q:
Sure, I can take a look. Do you have an idea what could be tested to cover this line? It looks like all the remaining actions without credentials are client-only. Perhaps we should instead assert that payload and payload.credentials are in the action?

nicolodavis:
Perhaps you can just make the function accept `CredentialedActionShape.Any` (remove `ActionShape.Any`) and then you no longer need the `if` condition?

tom3q:
Let me check. Thanks!

beknazar1:
Hi, is there any chance if this is ready to be merged and released any time soon?

Thanks.

tom3q:
Updated the changes. Please take a look. :)

maciejmatu:
That's some good stuff, waiting for it to be merged! 😄

Issue #594, opened 2020-03-29
Improve documentation about plugins


adngdb:
Hi there,

Not sure if this is the best place to ask questions, so if there's a better one please point me there.

I'm trying to use plugins, and am failing hard. I'm making a few observations that I don't understand, thus I'm not sure if they are bugs or features. The documentation about plugins is quite terse, so hopefully those questions will help improving it afterwards (I'll be happy to open a PR for that).

1. I cannot access `ctx.events` from inside my plugin, thus my plugin's `api` cannot advance the game state.
2. I cannot access my plugin's data in client-side React components, `ctx.myplugin` is undefined. 
3. I don't understand how the data works. I've looked at the PlayerPlugin for reference, and it seems it copies the plugin's data in the `api` call, the exposes that data, then in `flush` it returns the data exposed by `api`. That seems overly complex to me (and is not explained in the docs). I've thus tried to simply use `data` and have `flush` return that, and it seems to work fine. Any reason not to do that? 

Once again, I can provide code examples if needed.

Thanks!

PS: I'm naming this "Improve documentation about plugins" because that's the outcome that I'd like to see here, and I'll do my best to make it happen. :-)

nicolodavis:
The plugin API is very much in flux, and I didn't want to document it too much until it settles down a bit.
Let me add some background about plugins, which should hopeful answer some of your questions (and we can add some of this info to the docs as well):

### Each plugin gets its own private state

If you have plugins `A` and `B`, the game state is initialized to:

```js
{
  G: {},
  ctx: {},
  plugins: {
    A: { data: {} },
    B: { data: {} },
  }
}
```

### The setup function initializes `data`

This is probably not that confusing, despite the terseness of the docs.

### `api` creates an ephemeral object (that is not persisted)

The game state looks like this after applying the `api` functions from all the plugins:

```js
{
  G: {},
  ctx: {},
  plugins: {
    A: { data: {}, api: {} },
    B: { data: {}, api: {} },
  }
}
```

These new ephemeral objects are made available in `ctx` as `ctx.A` and `ctx.B`.  The intention is that the user will interact with these objects, and it will update its own in-memory state in response to that.

### `flush` then persists any necessary data back into `data`

This completes one cycle.  This whole circus is to ensure that `api` does not get persisted, and it allows providing the user with a nice object in `ctx` to interact with.

### Specific Questions

> I cannot access ctx.events from inside my plugin, thus my plugin's api cannot advance the game state.

[...]

PR #604, opened 2020-04-01
feat(plugins): Make player plugin a factory function

### LINKED ISSUE #603: New Plugin API
Following on from working on #598, I have a question about plugin configuration.

The player plugin expects a `playerSetup` function in the game config. Do you think there should be a more specific namespace for plugin configuration? Or a way of cleanly associating plugins with their configuration, so that methods don’t clutter up the main game config object? For example, a `pluginsConfig` object could be keyed by plugin name, and hand the value of that field to the plugin methods:

```js
// game configuration
const game = {
  plugins: [PluginPlayer],
  pluginsConfig: {
    player: {
      setup: (playerID) => ({}),
    },
  },
}

// plugins/plugin-player
setup: ({ ctx, config }) => {
  // use config.setup()
},
```

This could also make it easier to enforce correct types for Typescript users. For now, [I allowed any arbitrary key–value pair in the game config](https://github.com/nicolodavis/boardgame.io/blob/78729ebe9425a79ce65d76d1ae72211f8a17e726/src/types.ts#L214), but that’s not great, because it reduces type-checking for misspelled keys.

An alternative might be to encourage a pattern where plugins that need configuring provide a plugin factory that takes the configuration options and returns a plugin, with no need for custom fields in the game config:

```js
const game = {
  plugins: [
    PluginPlayer(setupFn),
  ],
}
```

delucis:
This removes the `playerSetup` function from the game configuration object. Instead, `PluginPlayer` should be called with an options object containing the setup function and it will return the plugin object boardgame.io expects.

**BREAKING CHANGE:** The `PluginPlayer` is no longer a plain object that looks for `game.playerSetup` during its setup. Instead, you have to call it, passing your setup method in an options object to the plugin:

```js
import { PluginPlayer } from 'boardgame.io/plugins';

const playerSetup = (playerID) => ({ ... });

const game = {
  name: 'my-game',
  plugins: [
    PluginPlayer({ setup: playerSetup }),
  ],
};
```

Closes #603.

### DIFF
--- a/docs/documentation/plugins.md
+++ b/docs/documentation/plugins.md
@@ -70,6 +70,23 @@ const game = {
 !> Plugins are applied one after the other in the order
 that they are specified (from left to right).
 
+##### Configuring Plugins
+
+Some plugins may need a user to provide some configuration. The recommended way to do that is to design the plugin as a factory function that takes configuration as its arguments and returns a plugin object.
+
+```js
+import { ConfigurablePlugin } from './plugins';
+
+const game = {
+  name: 'my-game',
+  plugins: [
+    ConfigurablePlugin(options),
+  ],
+}
+```
+
+?> See `PluginPlayer` below for an example of this in practice.
+
 #### Available Plugins
 
 **PluginPlayer**
@@ -77,9 +94,14 @@ that they are specified (from left to right).
 ```js
 import { PluginPlayer } from 'boardgame.io/plugins';
 
+// define a function to initialize each player’s state
+const playerSetup = (playerID) => ({ ... });
+
 const game = {
-  playerSetup: (playerID) => ({ ... }),
-  plugins: [PluginPlayer],
+  plugins: [
+    // pass your function to the player plugin
+    PluginPlayer(playerSetup),
+  ],
 };
 ```
 
--- a/docs/documentation/plugins.md
+++ b/docs/documentation/plugins.md
@@ -100,7 +100,7 @@ const playerSetup = (playerID) => ({ ... });
 const game = {
   plugins: [
     // pass your function to the player plugin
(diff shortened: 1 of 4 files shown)


docs/documentation/plugins.md
```diff
@@ -70,16 +70,38 @@ const game = {
...
 
+// define a function to initialize each player’s state
+const playerSetup = (playerID) => ({ ... });
+
 const game = {
-  playerSetup: (playerID) => ({ ... }),
-  plugins: [PluginPlayer],
+  plugins: [
+    // pass your function to the player plugin
+    PluginPlayer(playerSetup),
```
nicolodavis, line 103: How about making it accept an options object like:

```
{
  setup: ...
}
```

That way it will becoming easy to add additional named keys without increasing the number of arguments.

delucis: Done in e7e677b

MathieuLoutre:
I've tried creating a new plugin (see #609) with that new pattern which I prefer. I changed the signature of `setup` to have access to `ctx` in the hope of accessing the random functions like `Shuffle` to prepare a deck of cards but it seems like at the moment, the default plugins aren't loaded when `ctx` is given to the plugin's setup. Is that an oversight or intended? Should I bring my own shuffle function in the plugin?

delucis:
Hey Mathieu! I don‘t think that plugins can access the random functions at the moment, because they are also implemented as a plugin: https://github.com/nicolodavis/boardgame.io/issues/594#issuecomment-605820957

MathieuLoutre:
Thanks @delucis! Indeed, that's the same issue. Happy it's been raised :)
I'll bring my own shuffle for now and see if it fixes my other issue in the meantime.

Issue #611, opened 2020-04-07
Suggestion to improve type definitions for PlayerAPI 


crhistianramirez:
## Problem I'd like to address

Currently the PlayerAPI is defined like so

```typescript
export interface PlayerAPI {
  state: Record<PlayerID, any>;
  get(): any;
  set(value: any): any;
  opponent?: {
    get(): any;
    set(value: any): any;
  };
}
```

And is by default a part of context as an optional property

```typescript
export interface Ctx {
	...
	player?: PlayerAPI
}
```

I understand the logic here - the plugin api isn't defined unless the user's implemented it. But can cause a few issues:

- `ctx` is unnecessarily bloated with definitions that may not apply
- If the API is implemented the implementor now has to perform typechecking to keep typescript happy. For example `ctx.player.get()` will throw a typescript error complaining that `player` is possibly undefined.
- player state isn't strongly typed

## Proposal for solution

I propose instead that `player` is not (by default) part of the `GameConfig` interface  but can be added by developers that implement it on their own. Additionally, it would be neat if it took a type parameter for the player state.

Something like this:

```typescript
export interface PlayerAPI<TPlayerState = any> {
  state: Record<PlayerID, TPlayerState >;
  get(): TPlayerState ;
  set(value: TPlayerState ): TPlayerState ;
  opponent?: {
    get(): TPlayerState ;
    set(value: TPlayerState ): TPlayerState ;
  };
}
```

Then a developer could do something like this:

```typescript
interface MyPlayerState {
    teamID: string
}

interface MyGameConfig extends GameConfig {
    player: PlayerAPI<MyPlayerState>
}

const MyGame: MyGameConfig = {
    moves:  {
        drawCard: (G, ctx) => {
            const player = ctx.player.get();
            const teamID = player.teamID; // strongly typed!
        }
    }
}
```

crhistianramirez:
It would actually be pretty neat if `GameConfig` had a type parameter too for game state

```
export interface GameConfig<TGameState = any> {
    endIf?: (G: TGameState, ctx: Ctx) => any;
	...
```

delucis:
@crhistianramirez These are good proposals. Linking an earlier discussion along these lines: https://github.com/nicolodavis/boardgame.io/pull/579#issuecomment-603736732

nicolodavis:
@crhistianramirez I really like the idea of passing in the type of the user's game state as a template.  Would you have time to work on a PR for this?

[...]

PR #612, opened 2020-04-08
Typescript enhancements

crhistianramirez:
#### Checklist

Please see issue #611 

Re: Test coverage. Were you guys looking for type tests on these? Didn't see any other ones.

### DIFF
--- a/src/plugins/plugin-player.ts
+++ b/src/plugins/plugin-player.ts
@@ -12,18 +12,22 @@ interface PlayerData {
   players: Record<PlayerID, any>;
 }
 
-export interface PlayerAPI {
-  state: Record<PlayerID, any>;
-  get(): any;
-  set(value: any): any;
+export interface PlayerAPI<PlayerState extends any = any> {
+  state: Record<PlayerID, PlayerState>;
+  get(): PlayerState;
+  set(value: PlayerState): PlayerState;
   opponent?: {
-    get(): any;
-    set(value: any): any;
+    get(): PlayerState;
+    set(value: PlayerState): PlayerState;
   };
 }
 
 interface PluginPlayerOpts {
   setup?: (playerID: string) => any;
+}
+
+export interface PlayerPlugin<PlayerState extends any = any> {
+  player: PlayerAPI<PlayerState>;
 }
 
 /**
--- a/src/plugins/plugin-player.ts
+++ b/src/plugins/plugin-player.ts
@@ -39,7 +39,7 @@ export interface PlayerPlugin<PlayerState extends any = any> {
  */
 const PlayerPlugin = <PlayerState extends any = any>({
   setup,
-}: PluginPlayerOpts = {}): Plugin<
+}: PluginPlayerOpts<PlayerState> = {}): Plugin<
   PlayerAPI<PlayerState>,
   PlayerData<PlayerState>
 > => ({
--- a/src/plugins/plugin-player.ts
+++ b/src/plugins/plugin-player.ts
@@ -8,8 +8,8 @@
 
 import { Plugin, PlayerID } from '../types';
 
-interface PlayerData {
-  players: Record<PlayerID, any>;
+interface PlayerData<PlayerState extends any = any> {
+  players: Record<PlayerID, PlayerState>;
 }
 
 export interface PlayerAPI<PlayerState extends any = any> {
@@ -22,8 +22,8 @@ export interface PlayerAPI<PlayerState extends any = any> {
   };
 }
 
-interface PluginPlayerOpts {
-  setup?: (playerID: string) => any;
+interface PluginPlayerOpts<PlayerState extends any = any> {
+  setup?: (playerID: string) => PlayerState;
 }
 
 export interface PlayerPlugin<PlayerState extends any = any> {
@@ -37,9 +37,11 @@ export interface PlayerPlugin<PlayerState extends any = any> {
  *
  * @param {function} initPlayerState - Function of type (playerID) => playerState.
  */
-const PlayerPlugin = ({ setup }: PluginPlayerOpts = {}): Plugin<
-  PlayerAPI,
-  PlayerData
+const PlayerPlugin = <PlayerState extends any = any>({
+  setup,
+}: PluginPlayerOpts = {}): Plugin<
+  PlayerAPI<PlayerState>,
+  PlayerData<PlayerState>
 > => ({
   name: 'player',
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -5,22 +5,21 @@ import { Flow } from './core/flow';
 import { INVALID_MOVE } from './core/reducer';
 import * as StorageAPI from './server/db/base';
 import { EventsAPI } from './plugins/plugin-events';
-import { PlayerAPI } from './plugins/plugin-player';
 import { RandomAPI } from './plugins/plugin-random';
 
 export { StorageAPI };
 
 export type AnyFn = (...args: any[]) => any;
 
-export interface State {
-  G: object;
-  ctx: Ctx;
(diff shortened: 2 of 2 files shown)


src/types.ts
```diff
@@ -66,11 +65,7 @@ export interface Ctx {
   _random?: {
     seed: string | number;
   };
-  // enhanced by events plugin
   events?: EventsAPI;
-  // enhanced by player plugin
-  player?: PlayerAPI;
-  // enhanced by random plugin
   random?: RandomAPI;
 }
```
crhistianramirez, line 70: Based on @delucis comments
> The Events & Random plugins are always included automatically in every game set-up, so these would be the default plugins

Tried making `events` and `random` not optional but it break 20 or so odd tests so I left it as optional. Perhaps our assumption was incorrect

delucis: Ah, my guess would be that this is because there are actually two different contexts: a POJO that is serialised and saved as state/transferred over web sockets, and an “enhanced” context that has the plug-in API methods attached. There’s probably a need internally to go through marking what’s using the plain context and what’s using the enhanced context.

src/types.ts
```diff
@@ -5,22 +5,21 @@ import { Flow } from './core/flow';
...
+export interface State<G extends any = any, CtxWithPlugins extends Ctx = Ctx> {
+  G: G;
+  ctx: CtxWithPlugins;
   deltalog?: Array<LogEntry>;
   plugins: {
     [pluginName: string]: PluginState;
   };
-  _undo: Array<Undo>;
-  _redo: Array<Undo>;
+  _undo: Array<Undo<G, CtxWithPlugins>>;
```
nicolodavis, line 21: The undo and redo state should just have Ctx without the plugins.

src/types.ts
```diff
@@ -5,22 +5,21 @@ import { Flow } from './core/flow';
...
 export { StorageAPI };
 
 export type AnyFn = (...args: any[]) => any;
 
-export interface State {
-  G: object;
-  ctx: Ctx;
+export interface State<G extends any = any, CtxWithPlugins extends Ctx = Ctx> {
+  G: G;
+  ctx: CtxWithPlugins;
```
nicolodavis, line 16: Should we mark this as `Ctx | CtxWithPlugins` in order to emphasize that the state is enhanced with plugins in some contexts, but bare in other?

crhistianramirez: Ooh I like that. I was thinking about that and wasn't sure how we could make it more semantic

src/types.ts
crhistianramirez: Got it, I won't include in this PR but I think two interface might make sense then. Maybe `InternalCtx` and `Ctx`

crhistianramirez:
@nicolodavis @delucis 

Updated with your feedback. Let me know if you'd like me to organize my commits differently prior to merging. I can squash the pr feedback into one of the first three commits if we want commit history to be cleaner.

src/plugins/plugin-player.ts
```diff
@@ -33,9 +37,11 @@ interface PluginPlayerOpts {
  *
  * @param {function} initPlayerState - Function of type (playerID) => playerState.
  */
-const PlayerPlugin = ({ setup }: PluginPlayerOpts = {}): Plugin<
-  PlayerAPI,
-  PlayerData
+const PlayerPlugin = <PlayerState extends any = any>({
+  setup,
+}: PluginPlayerOpts = {}): Plugin<
```
delucis, line 42: Shouldn’t this be `PluginPlayerOpts<PlayerState>` here?

src/types.ts
```diff
@@ -90,7 +94,7 @@ interface PluginContext<
   CtxWithPlugins extends Ctx = Ctx
 > {
   G: G;
-  ctx: CtxWithPlugins;
+  ctx: Ctx | CtxWithPlugins;
```
delucis, line 97: Sorry, following on from @nicolodavis’s comment about `Ctx`, at the moment, `ctx` inside plugins doesn’t currently include the other plugins’ API. (More details here: https://github.com/nicolodavis/boardgame.io/issues/594#issuecomment-605820957) So these should typed as only `Ctx` too.

src/types.ts
```diff
@@ -103,55 +107,59 @@ export interface Plugin<
...
+  }) => State<G, Ctx | CtxWithPlugins>;
+  fnWrap?: (
+    fn: AnyFn
+  ) => (G: G, ctx: Ctx | CtxWithPlugins, ...args: any[]) => any;
 }
 
 type MoveFn<G extends any = any, CtxWithPlugins extends Ctx = Ctx> = (
   G: G,
-  ctx: CtxWithPlugins,
+  ctx: Ctx | CtxWithPlugins,
```
delucis, line 140: Conversely, moves and hooks _will_ have the `CtxWithPlugins`. (This is effectively the plain and “enhanced” contexts mentioned previously.)

delucis: (Hope this all makes sense!)

src/plugins/plugin-player.ts
crhistianramirez: Yes, good catch, missed that

src/types.ts
crhistianramirez: Yes, that issue you linked was very insightful. I've updated my PR

delucis: Thanks Crhistian! What I meant here was that for moves and hooks we can actually type them as `(G: G, ctx: CtxWithPlugins) => any`, because the context will always be enhanced before they are called. Sorry not to be clearer. I can go through and tidy up the places where we can use `CtxWithPlugins` instead of `Ctx | CtxWithPlugins` if necessary.

crhistianramirez: Ohhhhhh got it. Yeah, I can update it no problem

delucis: Awesome, thanks for hanging in there!

crhistianramirez: @delucis so if I understand correctly I think I want to revert my changes regarding `Ctx | CtxWithPlugins` and make just `State` be `ctx: Ctx | CtxWithPlugins` and have everything else be `ctx: CtxWithPlugins` does that seem right?

If not, I might need an assist. Not sure if I have a good enough understanding of the library yet to make the subtle distinctions

delucis: Yes, after a quick look through, I think that’s correct. The only one I’m not 100% on is the `processMove` method. I’ll have a look at the source for that.

src/types.ts
```diff
@@ -202,18 +236,22 @@ export interface Game {
...
-  onEnd?: (G: any, ctx: Ctx) => any;
-  playerView?: (G: any, ctx: Ctx, playerID: PlayerID) => any;
-  plugins?: Array<Plugin>;
+  endIf?: (G: G, ctx: Ctx | CtxWithPlugins) => any;
+  onEnd?: (G: G, ctx: Ctx | CtxWithPlugins) => any;
+  playerView?: (G: G, ctx: Ctx | CtxWithPlugins, playerID: PlayerID) => any;
+  plugins?: Array<Plugin<any, any, G>>;
   processMove?: (
-    state: State,
+    state: State<G, Ctx | CtxWithPlugins>,
```
delucis, line 244: I think you can leave this one as it is. (In any case, `processMove` is an internal field that is not intended for public consumption.)

src/types.ts
crhistianramirez: Looks like based on the source code `processMove` should also have `CtxWithPlugins` so I updated my PR as such

```
// src/core/reducer.ts

// Execute plugins.
state = plugins.Enhance(state, {
  game,
  isClient,
  playerID: action.payload.playerID,
});

// Process the move.
let G = game.processMove(state, action.payload);
```

crhistianramirez:
All requested changes should be addressed now

Issue #605, opened 2020-04-01
Plugin APIs no longer available in turn order methods


delucis:
(This is completely edited as I misdiagnosed this problem initially.)

I was using this in my initial turn order:

```js
first: (G, ctx) => ctx.random.Die(ctx.numPlayers) - 1
```

But after updating to 0.39.4, the randomness API is no longer available here. (It _is_ available during `setup`.) I guess this makes sense in the case of the Events API for example, but it would be nice to be able to randomise things here as was previously possible.

<details>
<summary>Original misdiagnosed issue report</summary>

Previously I used `ctx.random` to randomise various bits of game state in my game’s `setup` function. (As I outlined in https://github.com/nicolodavis/boardgame.io/issues/601#issuecomment-607126633.) I’ve just realised that the new plugin code doesn’t make the plugin APIs available in setup. Can we restore that previous behaviour?

Currently a plugin’s `setup` method notionally receives `G`, but none of the current plugin implementations use it, and I feel like it’s reasonable for a plugin not to depend on game-specific state at start-up. That would make it possible for the randomness API to be available as previously.

There are some questions, e.g. what happens to the events API if it gets called during setup? Or should there be some way for a plug-in to indicate if it is initialised before or after game state?

</details>

nicolodavis:
Ah this must have been an oversight on my part when I migrated the events and random API`s to plugins.  Will take a look.

flamecoals:
+1 I also could not do ctx.events.endPhase() there

delucis:
@flamecoals That is related, but can’t you use `endIf` in your phase config instead of using an event in the turn order?

flamecoals:
@delucis good point, I will try that. What I am dealing with is the turn.next on takesix code, the most popular gmae we have, returns undefined which _used_ to finish the phase:

https://github.com/freeboardgames/FreeBoardGames.org/blob/d0904895c80d4ee55dcabee49833b4f7ba592279/src/games/takesix/game.ts#L151

But now returning undefined throws an error/warning by the framework, even though it still works:
https://github.com/nicolodavis/boardgame.io/blob/45f4f432718c20a1a0f052fe0c2c25dc41e34a0b/src/core/turn-order.ts#L294

I am trying to "fix" it by replacing with equivalent behavior, hence I tried `ctx.events.endPhase()`, but it didnt work.

I will try your suggestion to see if I can find a criteria that fits when the turn should be ended.

delucis:
[...]

PR #614, opened 2020-04-11
fix(core): Don’t error if turn.order.next returns undefined

delucis:
Fixes a bug raised by @flamecoals in #605 where an error is logged if `turn.order.next` returns undefined, which is permitted as a way of ending the phase.

I’ve also amended the turn order interface to indicate that `next` can return undefined.

### DIFF
--- a/src/core/turn-order.ts
+++ b/src/core/turn-order.ts
@@ -291,7 +291,7 @@ export function UpdateTurnOrderState(
     const type = typeof t;
     if (t !== undefined && type !== 'number') {
       logging.error(
-        `invalid value returned by turn.order.next — expected number got ${type} “${t}”.`
+        `invalid value returned by turn.order.next — expected number or undefined, got ${type} “${t}”.`
       );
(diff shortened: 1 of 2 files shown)


nicolodavis:
Perhaps we should change the interface (in the future) to return a constant END_PHASE to make it more obvious.

jasonharrison:
I ran into this on v0.39.6 (where the changelog https://github.com/nicolodavis/boardgame.io/blob/master/docs/documentation/CHANGELOG.md says this was fixed)

It looks like @delucis 's commit was reverted in https://github.com/nicolodavis/boardgame.io/commit/eb1e06070eabe7d2e723b3f35fd553100183621a#diff-ffe51a82f97e27f74a79b6b8382009e3L292

Was this accidental?

nicolodavis:
Ah, that looks like a bad merge.

nicolodavis:
Fixed and released in 0.39.7.

Issue #460, opened 2019-09-16
Out of the box in game chat support


haoyangnz:
Any plan for out of the box in game chat support?
This would be very helpful.

nicolodavis:
I'm not opposed to this idea, especially if it can be configured nicely so that users that don't need it don't incur any cost.

We'd just need someone to sign up for this as I have my hands full with other areas of the framework.

haoyangnz:
@nicolodavis If I were to do this, do you have a recommended approach in mind?

nicolodavis:
We currently store game data in two areas:

1. Game State - this is where `G` and `ctx` are stored.
2. Game Metadata - this is where lobby info and other player related metadata are stored.

The areas above are two [key spaces](https://github.com/nicolodavis/boardgame.io/blob/master/src/server/api.ts#L59) in the same table (although in theory it is possible to change the storage adapters to use two entirely different tables for this).

For chat, I'd propose using a separate table that is able to store messages between players per game in a nice normalized flat structure.

```
key: <gameID>:<messageID>
value: { from: <playerID>, to: <playerID>, message: ' ... ' }
```

Links:
1. Storage Adapters - [link](https://github.com/nicolodavis/boardgame.io/tree/master/src/server/db)
2. Lobby API - [link](https://github.com/nicolodavis/boardgame.io/blob/master/src/server/api.ts)

blunket:
Yesss, I want this. Can this be opt-in, though? Preferably on a per-game basis?

I assume we'd have a pre-built chat component for the front-end while offering options to implement your own front-end (similar to the current lobby discussion).

nicolodavis:
Yes I'd hope that we made this purely opt in.  We can also implement (as a first step) a purely ephemeral chat that uses the existing socket layer.  That way you don't have to worry about storage, which might be exactly what some people want.

flamecoals:
I want to implement this, it is one of the most important features missing on FBG right now. 

Cant we really re-use the current store ? I was thinking something on the sorts of:

For sending messages:
```
this.props.chat.send('User message'); 
```

For listing all messages for all players:
```
this.props.chat.get();
```

Each message could be a specific action type on the store/reducer, and change the state on a separate part of the state object. This will mean that if a user reloads the page, the messages are still going to be there (retrieved from the store).

In fact, this is so close to what Plugins offer that I thought I could do it in a plugin, but I had trouble with the Plugins API not being exposed to the Board

delucis:
[...]

PR #629, opened 2020-04-20
Proof of concept for Text chat using BGIO plugins

flamecoals:
Hi, from our discussion on #460 ... I tried to make this work using plugins, but I got stuck for hours...

It seems like I can access plugin's state using `this.props.plugins.chat`, however, I could not find a way to reliably trigger an event to change it (calling its api).

@nicolodavis you pointed out to use the client's dispatcher, and I tried to use it here:
``this.props.client.plugins.chat()``, but this seems to be a really raw function that does not provide the plugin API, creating the Action for the store directly. I tried to search how this action should look like but I had no luck, I couldnt fully understand how plugins work. Can you point out how to use the API from the client?

Thanks for all your help!

#### Checklist

* [ x ] Use a separate branch in your local repo (not `master`).
* [x ] Test coverage is 100% (or you have a story for why it's ok).

### DIFF
--- a/examples/snippets/src/chat/index.html
+++ b/examples/snippets/src/chat/index.html
@@ -1,0 +1,26 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<style>
+body {
+  padding: 20px;
+}
+.msg {
+  position: absolute;
+  bottom: 0;
+  left: 20px;
+  color: #aaa;
+  font-size: 12px;
+  margin-bottom: 20px;
+}
+</style>
+</head>
+
+<body>
+  <div id="app"></div>
+
+  <script type="text/javascript" src="./index.js"></script>
+</body>
+
+</html>
--- a/examples/snippets/src/chat/index.js
+++ b/examples/snippets/src/chat/index.js
@@ -42,6 +42,12 @@ const ChatPlugin = {
   flush: ({ api }) => {
     return { messages: api.get() };
   },
+  action: (data, payload) => {
+    const message = { content: payload.args[0], playerID: payload.playerID };
+    return {
+      messages: [...data.messages, message],
+    };
+  },
 };
 
 const TicTacToe = {
@@ -51,8 +57,6 @@ const TicTacToe = {
 
   moves: {
     clickCell(G, ctx, id) {
-      debugger;
-      ctx.chat.send('hello');
       const cells = [...G.cells];
 
       if (cells[id] === null) {
@@ -133,7 +137,7 @@ class TicTacToeBoard extends React.Component {
         <pre>{JSON.stringify(this.props.plugins.chat, null, 2)}</pre>
         <button
           onClick={() => {
-            this.props.client.plugins.chat();
+            this.props.pluginsDispatcher.chat(prompt('Message?'));
           }}
         >
           Send
@@ -146,7 +150,7 @@ class TicTacToeBoard extends React.Component {
 var TicTacToeClient = Client({
   board: TicTacToeBoard,
   game: TicTacToe,
-  debug: true,
+  debug: false,
   multiplayer: Local(),
 });
 
--- a/examples/snippets/src/chat/index.js
+++ b/examples/snippets/src/chat/index.js
@@ -1,0 +1,167 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { Client } from 'boardgame.io/react';
+import { Local } from 'boardgame.io/multiplayer';
+
+function IsVictory(cells) {
+  const positions = [
+    [0, 1, 2],
+    [3, 4, 5],
+    [6, 7, 8],
+    [0, 3, 6],
+    [1, 4, 7],
+    [2, 5, 8],
+    [0, 4, 8],
+    [2, 4, 6],
+  ];
+
+  for (let pos of positions) {
+    const symbol = cells[pos[0]];
+    let winner = symbol;
+    for (let i of pos) {
+      if (cells[i] != symbol) {
+        winner = null;
+        break;
+      }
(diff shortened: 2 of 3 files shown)


flamecoals:
I also tried to use ` const ctxWithAPI = plugin.EnhanceCtx(state);` on the Client's store. However, that did not work because somehow the state that the clients receives does not have the api part of the plugin (only the plugin data).

I think it would be much better if we could rely on using client's `ctx` too, as it would be consistent with the usage on moves.

delucis:
Hey @flamecoals! Thanks so much for taking the time to try and get this working.

I think the plugin documentation still needs some improvements, but also there were some things not ready yet:

- A plugin’s `api` method is only for attaching the plugin API to `ctx` within moves, like `ctx.random.D6()`, which you can call in a move or hook, but not from the client.

    This is a little confusing, because the client does have additional dispatchers for events which mirror the events plugin’s API, so it can seem like that API is available from the client, but `ctx` on the client is just the plain state object.

- Dispatching an action like @nicolodavis mentioned was possible, is intercepted by the plugin’s action property, but this isn’t documented yet:

    ```js
    const plugin = {
      action: (pluginData, payload) => {
        // use action payload to potentially mutate pluginData
        return pluginData
      }
    }
    ```

- To dispatch the action, you should be able to call `plugins[pluginName]()`, but I found that this wasn’t yet exposed to the React client. Working on fixing that in #632, but it would clash with `props.plugins` at the moment (which contains plugins’ state) and I want to add some tests.

I still need to do some more experiments to understand exactly how a plugin’s `action` method receives an action sent from the client, but once that is working I think you could do something like:

```js
const ChatPlugin = {
  name: 'chat',

  setup: () => ({ messages: [] }),

  api: ({ data, playerID }) => ({
    get: () => data.messages,
    send: content => {
      data.messages = [...data.messages, { playerID, content }];
    },
  }),

  flush: ({ api }) => {
    return { messages: api.get() };
  },

  // very minimal action handler — you’d probably want some
  // validation or reformatting logic here
  action: (data, payload) => ({
    messages: [...data.messages, payload],
  }),
};
```

Then from moves and hooks, the game could use `ctx.chat.send(content)` to post bot messages etc. and from the client you would do `client.plugins.chat(content)`. Or something similar. In theory, the plugin definition could be designed in a way that would allow more structured APIs, rather than a single action handler that receives a payload, or we can stick to the action handler and let plugins use an action creator pattern.

> I think it would be much better if we could rely on using client's ctx too, as it would be consistent with the usage on moves.

This is definitely a point worth considering. I remember when I was starting out with the framework I found it confusing what was attached to `ctx` in moves compared to on the client. The one note of caution I would raise is that `ctx` in game logic and on the client have quite different powers. Exposing the randomness API on the client would probably not be very helpful and a chat message triggered by game code vs from a human player probably need to be handled differently, so in a way separating these two APIs into two places helps indicate that they’re not the same thing, even if they’ll often mirror each other closely.

nicolodavis:
`ctx` on the client is a read-only object (analogous to `G`).  The semantics would be somewhat confusing if people could affect the game state through it on the client.

I think the real problem is that we even attach an API to `ctx`.  The only reason that we do this is lack of real-estate in a move's signature:

```js
move: (G, ctx  /* convenient spot to piggyback on */, ...args) => {
}
```

However, I think a better solution would be to actually convert move signatures to:

```js
move: ({ G, ctx }, ...args) => {
}
```

That way, you can continue attaching other objects to the first argument and what you are using becomes more explicit (almost like an import):

```js
move: ({ G, ctx, random }, ...args) => {
  const roll = random.D6();
}

move: ({ G, ctx, events }, ...args) => {
  events.endTurn();
}
```

This will also avoid the confusion of [two ctx contexts](https://github.com/nicolodavis/boardgame.io/blob/master/src/types.ts#L70) throughout the codebase (it can stay a plain object everywhere).

The only downside is that this is backwards incompatible, but I think compatibility can be fixed in the short term via a plugin.  People can use this plugin to have instant compatibility while they migrate their code slowly:

```js
fnWrap: (fn) => ({ G, ctx, ...api }, ...args) => {
  const ctxWithAPI = { ...ctx, ...api };
  return fn(G, ctxWithAPI, ...args);
},
```

flamecoals:
Hey folks, thanks so much for the discussion and the help! I updated the proof of concept and it _almost_ works!

The state now is correctly changed after the user sends a message for BOTH players on this multiplayer game, but only if the player that sent the message is active player (player 0).

In the case I try to send the text message with the other player, I get:
```
ERROR: player not active - playerID=[1]
```

Any ideas on how we could solve that?

PS: I've added `this.props.pluginsDispatcher` in order to send the events as mentioned by @delucis

nicolodavis:
> but only if the player that sent the message is active player

You're running into the primary assumption made by the framework about plugins: that they affect game state.  This is why we run all the usual checks like we do for moves / events and disallow actions from unauthorized players.

We could relax the checks for plugin actions.  However, the downside is that if someone wants to implement a plugin to simulate an auction (say), they will need to duplicate all the turn order checks inside the plugin implementation which is not what we want.

We should probably add a boolean flag (in a separate PR) to relax this check for plugin actions that are meant to be made without any checks: https://github.com/nicolodavis/boardgame.io/blob/master/src/master/master.ts#L235

nicolodavis:
Hi @flamecoals, what's the status of this PR?

flamecoals:
This was just a proof of concept. After we discussed I'm not sure if it is
the best approach to do the text chat, we might implement a separate
service for that.

nicolodavis:
OK. I'll close this PR then.

Let us know how it goes.

Issue #322, opened 2018-12-20
nginx config for server 


pociej:
Im trying to setup server on DigitalOcean and im really not sure how should i config nginx to make this working. I will share if i will find the way but maybe someone can help by providing instruction how to config nginx for this case? Don't you that would be nice to add some deployment instructions to docs?

pociej:
I opened question with details on stackoverflow : https://stackoverflow.com/questions/53866211/nginx-config-for-socket-io

flamecoals:
I have a working nginx for turnato.com. I will send it soon here

flamecoals:
I recommend setting up a SSL protocol on your nginx. I recommend LetsEncrypt. Besides that, your nginx configuration file seems right. Also, [I do not overwrite the server URL in turnato](https://github.com/flamecoals/turnato/blob/c23f75953769299bfe4eabc257527aaee94eaeeb/src/components/games/chess/index.tsx#L38), as it is running on the same domain and the default ports.

This is the full nginx configuration file for [turnato.com](https://turnato.com):
`server {

    server_name turnato.com;

    location / {
        proxy_pass http://localhost:8001;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection 'upgrade';
        proxy_set_header Host $host;
        proxy_cache_bypass $http_upgrade;
    }

    location /blog {
        autoindex on;
        root /home/turnato-server;
    }

    listen 443 ssl; # managed by Certbot
    ssl_certificate /etc/letsencrypt/live/turnato.com/fullchain.pem; # managed by Certbot
    ssl_certificate_key /etc/letsencrypt/live/turnato.com/privkey.pem; # managed by Certbot
    include /etc/letsencrypt/options-ssl-nginx.conf; # managed by Certbot
    ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem; # managed by Certbot
}
server {
    server_name www.turnato.com;
    return 301 https://turnato.com;

    listen 443 ssl; # managed by Certbot
    ssl_certificate /etc/letsencrypt/live/turnato.com/fullchain.pem; # managed by Certbot
    ssl_certificate_key /etc/letsencrypt/live/turnato.com/privkey.pem; # managed by Certbot
    include /etc/letsencrypt/options-ssl-nginx.conf; # managed by Certbot
    ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem; # managed by Certbot
}
server {
[...]

PR #631, opened 2020-04-20
Add option to run server over HTTPS

c-w:
This commit adds the option to run the boardgame.io server over HTTPS by passing through [koa-socket-2 SSL options](https://github.com/ambelovsky/koa-socket-2/blob/3c4433e/README.md#https-example) to the SocketIO transport.

While it's also possible to achieve HTTPS in other ways, e.g. via nginx as discussed in https://github.com/nicolodavis/boardgame.io/issues/322, enabling first-party support for HTTPS makes it much simpler to spin up such services since it removes the need to install and configure an external SSL terminator and proxy.

#### Checklist

### DIFF
--- a/src/server/transport/socketio.js
+++ b/src/server/transport/socketio.js
@@ -62,6 +62,7 @@ export function SocketIO({
   clientInfo = new Map(),
   roomInfo = new Map(),
   auth = true,
+  https,
 } = {}) {
   return {
     init: (app, games) => {
@@ -73,7 +74,7 @@ export function SocketIO({
       });
 
       app.context.io = io;
-      io.attach(app);
+      io.attach(app, !!https, https);
 
       for (const game of games) {
         const nsp = app._io.of(game.name);
--- a/docs/documentation/api/Server.md
+++ b/docs/documentation/api/Server.md
@@ -62,6 +62,24 @@ server.run(8000);
 server.run(8000, () => console.log("server running..."));
 ```
 
+##### With HTTPS
+
+```js
+const { Server } = require('boardgame.io/server');
+const fs = require('fs');
+
+const server = Server({
+  games: [game1, game2, ...],
+
+  https: {
+    cert: fs.readFileSync('/path/to/cert'),
+    key: fs.readFileSync('/path/to/key'),
+  },
+});
+
+server.run(8000);
+```
+
 ##### With custom authentication
 
 `generateCredentials` is called when a player joins a game with:
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -26,6 +26,11 @@ interface ServerConfig {
   };
 }
 
+interface HttpsOptions {
+  cert: string;
+  key: string;
+}
+
 /**
  * Build config object from server run arguments.
  */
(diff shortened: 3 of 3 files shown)


delucis:
@c-w Thanks so much for this! Do you think it would make sense to also include the Koa apps in this as well?

Currently we run the Lobby and API servers using Koa’s `listen` method:

https://github.com/nicolodavis/boardgame.io/blob/475930512ca51f037aa8c5a8cb803d5958265045/src/server/index.ts#L122

https://github.com/nicolodavis/boardgame.io/blob/475930512ca51f037aa8c5a8cb803d5958265045/src/server/index.ts#L131

These use `http` internally, but [the Koa docs](https://github.com/koajs/koa/blob/master/docs/api/index.md#applisten) describe how to use Node’s `https` module instead. If we implemented this, the certificates could be used to secure the whole server, not just the socket transport.

Let me know what you think, and if it’s too much for this PR, I’ll get this merged and we can add it to the to-do list.

c-w:
@delucis Thanks for the review. I definitely also want to include the Koa apps in the HTTPS setup.

My initial implementation of this feature actually was very similar to your suggestions, namely a refactor of the `Server.run` method to return both Koa application objects so that calling code could then run them in their own http or https server. However, it turns out that koa-socket-2 actually already creates a http or https server for us in [attach(app, https, opts)](https://github.com/ambelovsky/koa-socket-2/blob/3c4433e/index.js#L106-L115). As such, all that we need to do to host the socket server and API servers on HTTPS is to set up HTTPS on the socket.

The code I pushed to the pull request doesn't make this setup obvious and we technically have no guarantee that any transport implementation provided by a caller maintains the invariant that `app.server` is undefined by the time that SocketIO is initialized. As such, I pushed an update to the pull request to encapsulate this logic: we now have a `https` argument in the Server function that sets up the wiring as required.

src/server/transport/socketio.js
```diff
@@ -62,6 +62,7 @@ export function SocketIO({
   clientInfo = new Map(),
   roomInfo = new Map(),
   auth = true,
+  https = undefined,
```
delucis, line 65: This can just be:

```js
https,
```

The default value will always be undefined.

src/server/transport/socketio.js
```diff
@@ -73,7 +74,7 @@ export function SocketIO({
       });
 
       app.context.io = io;
-      io.attach(app);
+      io.attach(app, https != null, https);
```
delucis, line 77: Could you use `!!https` for the boolean flag here?

src/server/transport/socketio.js
c-w: Done.

src/server/transport/socketio.js
c-w: Done.

delucis:
Thanks Clemens! Especially for including the documentation update with this.

c-w:
Thanks for the merge! Any chance I could get a patch release for this too? I'm using this feature on a game I'm working on currently and would love to remove my local package override hack.

delucis:
@nicolodavis is managing the release flow, but I’ll ask!

nicolodavis:
Will release this soon.

Let me also point out that NPM supports GitHub installs so it's very easy to just do:

```
npm i --save nicolodavis/boardgame.io
```

and get the latest code rather than waiting for an official NPM package.

PR #632, opened 2020-04-20
fix(client): Expose plugin action dispatchers in React client

delucis:
`client.plugins` contains dispatcher for plugin actions, but these aren’t currently exposed to the React client.

## To-do

- [ ] Expose plugin dispatchers in props. The `plugins` namespace is already used by `state.plugins`, because `state` is destructured into props, so some other API might be needed.

- [ ] Add tests for the plugin dispatchers.

### DIFF
--- a/src/client/react.js
+++ b/src/client/react.js
@@ -124,6 +124,7 @@ export function Client(opts) {
           isMultiplayer: !!multiplayer,
           moves: this.client.moves,
           events: this.client.events,
+          plugins: this.client.plugins,
           gameID: this.client.gameID,
           playerID: this.client.playerID,
           reset: this.client.reset,


delucis:
@nicolodavis Do you have any opinions on how to make the plugin dispatchers available in the React props. I’m currently trialling mixing the dispatchers into the `plugins` prop as a method named `dispatch` for each plugin so that you would do:

```jsx
function Board(props) {
  const { data, dispatch } = props.plugins[pluginName];
  const onClick = () => dispatch('Go');
  return (
    <button onClick={onClick}>Go</button>
  );
}
```

It feels nice to keep all the plugin-related stuff under the same path, but it does mix methods and state in a way the rest of the props don’t. It also then diverges from the vanilla client API, which isn’t super great either.

nicolodavis:
I don't have a strong preference between either pattern, but I definitely do think that all of this needs to happen on the raw client.  The React layer should be very thin, merely wrapping an object that exists on the raw client.  This will also ensure that other implementations (Angular etc.) can stay in sync more easily and provide a consistent interface.

The current way of dispatching events is analogous to the way moves and events are dispatched:
```js
client.moves.moveName();
client.events.eventName();
client.plugins.pluginName();
```

But I can see the benefit of also having the plugin data available under a single object rather than having to fish it out of the state.  If we do this, we should also erase it from the state before we send it to the client in order to avoid confusion that it is available in two places (this will also introduce a new type for the state object, so that needs consideration).

flamecoals:
Just FYI, there is already a "plugins" prop coming for the react components (from the `...state`), and this would overwrite it.

Also it was with this prop that I was able to get the plugin state. So for my proof of concept I created another prop `pluginsDispatcher` for this.

Probably a good idea to try to merge both on client `getState()` ?

delucis:
A few more follow-up thoughts:

1. Putting plugins alongside moves and events does emphasise how much less expressive a plugin’s client-side API would be. Would the plan be to encourage a plugin to provide action creators if it needed structured action payloads?

    ```js
    import { Send } from 'PluginChat'

    const action = Send('Hello', playerID)
    client.plugins.chat(action)
    ```
 
    If we did want to let plugins define a client-side API, how might we expose that?

2. Should we create a plugin dispatcher for a plugin if it doesn’t declare an `action` method?

3. If we want to preserve the vanilla client API as much as possible, should we stop destructuring state in the React board props? It’s a significant breaking change that would probably impact 100% of users using React, but the migration path is simple and for most people would mean modifying a single file and doing something like `const { G, ctx, plugins } = props.state`. This would resolve the namespace clash between `client.plugins` and `state.plugins`. I also kind of like how this might clarify what is coming from the game state to users.

nicolodavis:
> Putting plugins alongside moves and events does emphasise how much less expressive a plugin’s client-side API would be. Would the plan be to encourage a plugin to provide action creators if it needed structured action payloads?

Yes, I think just exposing a single dispatch method gives the author more flexibility to use any action subdivision method that they would like to use.

> Should we create a plugin dispatcher for a plugin if it doesn’t declare an action method?

Ah, we should probably not.  I didn't originally consider this.

> should we stop destructuring state in the React board props?

I'm not opposed to this idea, but maybe we should kick this off as a RFC on the issue tracker and get some more opinions?  Maybe we can bunch this together with a slightly broader API overhaul.  For example, I'd also like to explore the idea I proposed here: https://github.com/nicolodavis/boardgame.io/pull/629#issuecomment-616900537

delucis:
@nicolodavis Thanks for the feedback. Combining this with your proposed changes in #629 could make sense. Especially if we bundle a plugin like you describe to help smooth migration. 

I'll probably be too busy this week to put together an RFC, so if you have time, please go ahead. 

While we're on this topic I've been wondering about client plugins. I've been tinkering with a game plugin that pairs well with a client plugin to handle its state (in this case to create an event emitter). Currently though I would have to write different code for React boards and the vanilla client. It would be nice to support some kind of client plugin API that could be accessed on both the vanilla client and client wrappers (e.g. via a `plugins` option to the React client factory).

nicolodavis:
> While we're on this topic I've been wondering about client plugins.

Oh I see.  Do you have a link to your client plugin?  Let's explore how we can add support for this.  Sounds like a good idea.

delucis:
@nicolodavis I haven’t finished work on this, or even decided exactly what the best client API would be, but here’s the WIP: https://github.com/delucis/bgio-effects

I’ve tried to document as I go to clarify usage to myself, so the README should explain the general idea.

Issue #618, opened 2020-04-13
UX issues with contenteditable editors in the debugger


will-hart:
There are a few UX issues with the `contenteditable`s in the debug panel.

In particular hotkeys are still applied while a "move" argument is being edited. As a result of this, if you type an argument such as `abcd` then it will change tabs when the `b` key is pressed. This makes it impossible to type the input in.

Copying and pasting is also a bit challenging because `Ctrl+C` is captured and prevented.

Chrome 81, Win 10

nicolodavis:
Thanks for the report @will-hart!

nicolodavis:
@will-hart I was not able to reproduce the copy / paste issue.  Can you describe the issue you ran into?

will-hart:
When I have an action in the debug view which has a `c` short cut, then trying to `ctrl+c` copy anywhere in the debugger triggers the `c` action and seems to `preventDefault` on the copy. Therefore `ctrl+c` doesn't work in the debug view.

I'm not really familiar with the code, but does https://github.com/nicolodavis/boardgame.io/blob/master/src/client/debug/main/Hotkey.svelte#L26 need to check for `e.altKey` or `e.ctrlKey` and ignore if they're pressed?

PR #640, opened 2020-04-23
fix UX issues in move debugger

nicolodavis:
#### Checklist

Fixes https://github.com/nicolodavis/boardgame.io/issues/618


src/client/debug/main/InteractiveFunction.svelte
```diff
@@ -69,10 +71,14 @@
...
 
 <div class="move" class:active on:click={Activate}>
-  {name}(<span
+  <span>{name}</span>
+  <span>(</span>
+  <span
     class="arg-field"
     bind:this={span}
     on:blur={Deactivate}
+    on:keypress|stopPropagation={() => {}}
```
delucis, line 80: First time reviewing Svelte code, but am I understanding correctly that the important change is this line preventing `on:keypress` listeners firing elsewhere while editing the contenteditable span and that that span is displayed conditionally somehow?

delucis:
This is an additional change, but given it’s in the same scope:

When the debug panel is hidden using the <kbd>.</kbd> shortcut, the hotkeys for toggling which panel is displayed (<kbd>m</kbd>, <kbd>l</kbd>, <kbd>i</kbd> & <kbd>a</kbd>) are still active. This is a minimal bug, because the controls within the panels seem to be disabled correctly, but might be a simple fix to include here to ensure the panels aren’t toggled while hidden.

delucis:
Did some local testing and this looks good for disabling shortcuts inside the move content editables 👍 

To fix the copy/paste issue, I think we can add checks here:

https://github.com/nicolodavis/boardgame.io/blob/06122c7e5b4e0f2f0cfe13ce8bb899c6e37903ad/src/client/debug/main/Hotkey.svelte#L26

Something like:

```js
    if (
      !$disableHotkeys && !disable &&
      !e.ctrlKey && !e.metaKey &&
      e.key == value
    ) { 
```

I don’t think we need to check for alt or shift keys as they tend to produce different `e.key` values?

To fix toggling the displayed pane while the debug panel is hidden, I think we need to move this code inside an if statement that checks if `visible` is true:

https://github.com/nicolodavis/boardgame.io/blob/06122c7e5b4e0f2f0cfe13ce8bb899c6e37903ad/src/client/debug/Debug.svelte#L37-L41

nicolodavis:
Thanks @delucis!

PR #642, opened 2020-04-23
Add an endpoint to the Lobby  to add additional information  to a player

arome:
#### Checklist

Ok, so my coverage isn't 100%, I'm not too familiar with unit testing (I know, not cool), I did the API part but there's this one line that isn't covered (api.test.js:342)

### DIFF
--- a/src/master/master.ts
+++ b/src/master/master.ts
@@ -334,8 +334,8 @@ export class Master {
 
     if (gameMetadata) {
       filteredMetadata = Object.values(gameMetadata.players).map(player => {
-        const { credentials, ...fileteredData } = player;
-        return fileteredData;
+        const { credentials, ...filteredData } = player;
+        return filteredData;
       });
     }
 
--- a/src/master/master.ts
+++ b/src/master/master.ts
@@ -334,7 +334,8 @@ export class Master {
 
     if (gameMetadata) {
       filteredMetadata = Object.values(gameMetadata.players).map(player => {
-        return { id: player.id, name: player.name };
+        const { credentials, ...fileteredData } = player;
+        return fileteredData;
       });
     }
 
--- a/src/server/api.test.ts
+++ b/src/server/api.test.ts
@@ -258,7 +258,7 @@ describe('.createApiServer', () => {
             });
             response = await request(app.callback())
               .post('/games/foo/1/join')
-              .send('playerID=0&playerName=alice');
+              .send({ playerID: 0, playerName: 'alice', data: 99 });
           });
 
           test('is successful', async () => {
@@ -276,6 +276,18 @@ describe('.createApiServer', () => {
                 players: expect.objectContaining({
                   '0': expect.objectContaining({
                     name: 'alice',
+                  }),
+                }),
+              })
+            );
+          });
+          test('updates the player data', async () => {
+            expect(db.mocks.setMetadata).toHaveBeenCalledWith(
+              '1',
+              expect.objectContaining({
+                players: expect.objectContaining({
+                  '0': expect.objectContaining({
+                    data: 99,
                   }),
                 }),
               })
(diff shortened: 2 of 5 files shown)


arome:
> Thanks for this @arome! I have a few suggestions:
> 
> * [x]  Can we rename `additionalInfo` to `data`? For the type I would suggest just having it as `any` rather than an indexed type in case someone just wants to store a primitive.
> * [x]  Can we accept and set a player’s `data` along with their `name` also in the `/join` endpoint if provided?
> * [x]  Can we consolidate the `rename` and `update` endpoints?
>   I would suggest doing this as follows:
>   
>   * Write the endpoint handler to accept `newName` and/or `data`, merging that into the existing metadata object.
>   * Expose the same handler as `/games/:name/:id/update` _and_  `/games/:name/:id/rename`, but make calls to the `/rename` endpoint log a warning that it is deprecated and `/update` should be used instead.
>   * Remove documentation for the `/rename` endpoint, replacing it entirely with `/update`
> 
> How does this sound? I can take a look at test coverage once we’re closer to a final implementation.

Sounds great, made the changes. 
About logging the warning that it is deprecated, I currently logged it in the console, but I think what is needed is to actually return it to the client-side, the problem is, it seems like the [Warning](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Warning) header response is soon to be deprecated, so how do you suggest I should proceed?
Also, for some reason, the tests done locally and the ones done on Semaphore aren't producing the same result? I don't get any error running them using npm run test/test:coverage. But it fails during CI

src/master/master.ts
```diff
@@ -334,7 +334,8 @@ export class Master {
 
     if (gameMetadata) {
       filteredMetadata = Object.values(gameMetadata.players).map(player => {
-        return { id: player.id, name: player.name };
+        const { credentials, ...fileteredData } = player;
+        return fileteredData;
```
delucis, line 338: Typo here, should be `filteredData`.

src/server/api.ts
```diff
@@ -15,6 +15,7 @@ import cors from '@koa/cors';
 import { InitializeGame } from '../core/initialize';
 import * as StorageAPI from './db/base';
 import { Server, Game } from '../types';
+import { update } from '../../dist/types/src/core/action-creators';
```
delucis, line 18: Is this used somewhere? This is breaking the CI tests because you’re trying to import from `dist` in `src`. In any case, I don’t think this is being used, so it’s probably a case of just deleting this line.

src/server/api.ts
```diff
@@ -285,19 +292,20 @@ export const addApiToServer = ({
...
+    const data = ctx.request.body.data;
     const { metadata } = await (db as StorageAPI.Async).fetch(gameID, {
       metadata: true,
     });
     if (typeof playerID === 'undefined') {
       ctx.throw(403, 'playerID is required');
     }
-    if (!newName) {
-      ctx.throw(403, 'newName is required');
+    if (!data && !newName) {
```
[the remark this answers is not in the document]
arome: Yep makes sense

src/server/api.ts
arome: Oh, I probably pressed tab accidentally while writing the update function and typescript auto imported this. Definitely not using it, removed it, thanks.

src/server/api.ts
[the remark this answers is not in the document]
arome: So what happens if the user sends data and a new name but the new name is not a string? Do I fail the whole operation, or partially update?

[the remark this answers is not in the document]
delucis: I think it’s OK to fail the whole thing. If the payload is partially wrong, it’s probably best not to leave the metadata in some half-updated state.

src/server/api.test.ts
```diff
@@ -472,6 +484,269 @@ describe('.createApiServer', () => {
...
+            test('newName is required', async () => {
+              expect(response.text).toEqual('newName or data is required');
+            });
+          });
+        });
+      });
+    });
+  });
+
+  describe('updating', () => {
```
delucis, line 617: Let’s name this to match the handler you wrote, e.g. “updating player metadata”.

delucis:
Great work @arome! This should be a nice improvement for the next release.

arome:
Awesome thanks to you for your guidance and input! The question I have now is, when is the next release expected?

delucis:
@arome There’s no fixed release schedule. There are a few open PRs it would be good to get merged and then I imagine a release might follow shortly after — they’re all small progressive improvements — but it might take a few days.

PR #646, opened 2020-04-24
feat(core): Support moves that don’t contribute to numMoves

### LINKED ISSUE #644: Feature request: excluding a move from the moveLimit.
I think the title speaks for itself 😏. Here is an example of how would I see it. 

```js
const someGame = {
    setup : () => {
        // some code here
    },

    moves : {
        someMove      : ( G, ctx, id ) => {},
        limitFreeMove : {
            move              : ( G, ctx, id ) => {},
            moveLimitExcluded : true,
        },
    },
};
```

delucis:
This implements a long-form move syntax that prevents a move from incrementing `numMoves` and by extension triggering move limits. It would close #644.

This is a slightly naïve way of implementing this `noLimit` feature because as a consequence we completely lose track of those moves in the move counts. In theory we could track `numMoves` and `limitNumMoves` separately and trigger move limits using the latter, but it would make state and logic more complex. Let me know if you think it’s acceptable as is or if we should implement the more robust version.

### DIFF
--- a/docs/documentation/api/Game.md
+++ b/docs/documentation/api/Game.md
@@ -20,6 +20,7 @@
       undoable: false,  // prevents undoing the move.
       redact: true,     // prevents the move arguments from showing up in the log.
       client: false,    // prevents the move from running on the client.
+      noLimit: true,    // prevents the move counting towards a player’s number of moves.
     },
   },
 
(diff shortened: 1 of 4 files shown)


tibo1902:
Hi,

Would it be possible to modify this functionality as follows: 

add a moveCost properety with a default value of 1. If it's set to 0, it will work as noLimit: true, with the possibility of setting higher costs for certain moves.

PR #648, opened 2020-04-25
Add Azure Storage database connector

c-w:
This pull request refactors the FlatFile database connector into an arbitrary key-value connector and uses the new abstraction to implement a database connector that leverages [Azure Storage](https://azure.microsoft.com/en-us/services/storage/) for persistence. In the future this model can be easily extended to other object storage persistence layers such as AWS S3, Google Cloud Storage, etc.

#### Checklist

### DIFF
--- a/docs/documentation/storage.md
+++ b/docs/documentation/storage.md
@@ -40,6 +40,10 @@ server.run(8000);
 
 Instructions at https://github.com/delucis/bgio-firebase.
 
+#### Azure Storage
+
+Instructions at https://github.com/c-w/bgio-azure-storage.
+
 #### Postgres
 
 Coming soon.


delucis:
Hi Clemens, thanks for working on this!

A couple of things:

1. I think the plan is to encourage users to maintain database connectors as separate packages. (For example, I maintain [a connector for Cloud Firestore](https://github.com/delucis/bgio-firebase).) It’s challenging to maintain and test backend implementations unless you’re familiar with the service and potentially actively using it, so this would allow community members to contribute with their expertise and maintain connectors independently. Would you be happy to publish a `bgio-azure-storage` package? Then we would link to it from the storage docs.

2. We could consider adding the abstract `KeyValueStore` if you think it’s necessary, but in general the `StorageAPI.Async` class is designed to allow backend implementations to make atomic changes, which the `get`/`set` per-item approach doesn’t permit. For example, some database methods write or read multiple bits of data and by grouping those operations into a single method the implementation can ensure that they either all succeed or all fail. I’m not familiar with the Azure API, so maybe that’s not relevant, but with Firestore for example, you can use a transaction to do this. As a further example, `FlateFile` is pretty basic in how it implements `fetch`, dispatching each read synchronously. In a production implementation, you would probably want to dispatch all those requests in parallel.

What do you think? If you have any questions about publishing a separate package, let me know!

c-w:
Thanks for the review. I moved the connector to a separate package ([Github](https://github.com/c-w/bgio-azure-storage), [NPM](https://www.npmjs.com/package/bgio-azure-storage)) as requested.

delucis:
Awesome, thank you! Feel free to open a PR linking to your module in the storage docs, or I can take care of that when I’m back at a computer.

c-w:
@delucis I force pushed this branch to include the documentation update. Could we reopen this?

PR #653, opened 2020-04-27
feat(server): Add option to exclude rooms from public listing

c-w:
This commit adds an option to game room creation to mark the room as unlisted. Unlisted rooms are excluded from the list of all rooms. The setting is useful for situations where someone wants to host a room for friends and therefore doesn't want arbitrary people to be able to join their room via the lobby.

#### Checklist


delucis:
Thanks for this!

PR #654, opened 2020-04-27
Use `/update` endpoint in unit tests instead of deprecated `/rename`.

adngdb:
This is trivial, it just removes some warnings when running tests, following the recent renaming of the `/rename` Lobby endpoint to `/update` in #642.

#### Checklist

### DIFF
--- a/src/server/api.test.ts
+++ b/src/server/api.test.ts
@@ -388,7 +388,7 @@ describe('.createApiServer', () => {
           });
           const app = createApiServer({ db, games });
           response = await request(app.callback())
-            .post('/games/foo/1/rename')
+            .post('/games/foo/1/update')
             .send('playerID=0&playerName=alice&newName=ali');
           expect(response.status).toEqual(404);
         });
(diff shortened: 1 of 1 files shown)


arome:
So I kept these tests for users who would still be using the deprecated rename endpoint. Otherwise, if it's not necessary, we should remove the whole 'rename with deprecated endpoint' section from the tests, since the rename tests are done in the 'updating player metadata' section.

adngdb:
Note that my opinion doesn't matter much, but I like that there's a redirection from `/rename` to `/update` with a deprecation notice (it allows not breaking existing games and allowing a smooth transition), but I don't see the point of showing those warnings during test runs. If the point was to have a test that `/rename` redirects to `/update`, then I believe that should be a specific test dedicated to just that.

delucis:
Hi — thanks for noticing the `console.warn` output during the tests. I think we should keep the `rename` test suite as is. My reasoning was that, when we do remove the `rename` endpoint, we will also need to remove the `rename` tests, which will make it clearer that that’s a breaking change. And as long as the endpoint is included, we should test it.

The actual annoyance is that we don’t want to `console.warn` during testing. I would suggest we mock `console.warn` for the `rename` tests:

```js
console.warn = jest.fn()
```

This would stop the log output and as a bonus would also allow us to add a test asserting that console warn is called with the deprecation message.

adngdb:
Aaah I missed that what I changed were tests specific for the `/rename` endpoint. OK, that makes sense then.

PR #656, opened 2020-04-29
[#623, #588] Make plugins available in onBegin/onEnd when called from a move.

### LINKED ISSUE #588: Random is undefined when manually starting new phase
When starting new phase using either `endPhase` or `setPhase` callback, the `random` object isn't defined in `ctx` passed to `onBegin` hook. However, when previous phase ends from `endIf` hook everything works as expected.
I've prepared a minimal example showing this bug, available here https://codesandbox.io/s/boardgameio-t8b29

### LINKED ISSUE #623: Plugins not accessible in turn's `onEnd` when `endTurn` called from a move
When you call `ctx.events.endTurn()` from inside a move, during the subsequent call to `onEnd`, it seems that plugins are not accessible. However, if you simply call `ctx.events.endTurn()` directly from your input, it works just fine.

## Library version

I reproduced this behavior with versions 0.39.5 and 0.39.7.

## Steps to reproduce

1. Use `ctx.events.endTurn()` in a move function.
2. Use `ctx.random.Die()` or any plugin methods inside the turn's `onEnd` callback.
3. Call your move function.

## Expected result

The plugin method is available and produces the expected results.

## Actual result

An exception is raised: ` TypeError: Cannot read property 'Die' of undefined`.

## Test case

https://codesandbox.io/s/boardgameio-uosrp

adngdb:
I'm pretty sure this is the wrong solution to this problem, but I'm blocked by this issue so I'm doing my best to move it forward. Please send help! :pleading_face: 

Here's what I believe happens:
1. a move happens that triggers the end of a phase or turn;
2. that move resolves, then plugins are flushed;
3. during the flushing of the events plugin, the onBegin or onEnd trigger is called;
4. right before that trigger runs, a wrapper enhances `ctx` with plugins, but because that happens while plugins are being flushed, the `random` plugin doesn't exist (it's not in `state.plugins`). Thus, the trigger is called without the random plugin's API.

Now this code fixes my unit tests, but it breaks another one (`basic / does not make it into undo state`). I don't really understand that test but it looks a bit fishy to me, so I'm wondering if it is actually doing what it's suppose to do? Why would calling `undo()` remove the `test` plugin from `ctx`? That doesn't make sense to me.

Any help to move this forward will be greatly appreciated! :-)

---

Closes #623, Closes #588


delucis:
Hi Adrian, thanks for looking into this. I’m afraid Nicolo and I are quite busy this week, so it may take a little longer than usual to get an actual fix for this. Your tests look good though — thanks for starting with those. I’ll try in any case to give some quick feedback, hoping it might be helpful!

The [`basic > does not make it into undo state` test](https://github.com/nicolodavis/boardgame.io/blob/a4c4c7c2b10dc9a5080383062714d84de63e56e4/src/plugins/main.test.js#L107-L112) is correct as far as I can tell:

- The game is using a custom plugin called “test”.

- In the reducer, the plugin’s API will be attached to `ctx` (“enhancing” it).

- But when the state object is returned to the client, `ctx` should no longer contain the API. It should be a plain object with no functions etc. (this is the “flushing” step).

- This test makes sure that the “flushing” also happens when a move is undone.

It is failing here because the API methods are still found in the `ctx` returned to the client. We can see in the error message that not only `test`, but also `random` & `events` are still attached to `ctx`:

```
FAIL src/plugins/main.test.js
● basic › does not make it into undo state

  expect(received).not.toContain(expected) // indexOf

  Expected value: not "test"
  Received array:     [
    "numPlayers", "turn", "currentPlayer", "playOrder", "playOrderPos", "phase",
    "activePlayers", "_activePlayersMoveLimit", "_activePlayersNumMoves",
    "_prevActivePlayers", "_nextActivePlayers", "numMoves", "random", "test", "events"
  ]
```

The issue with your proposed fix is that it doesn’t guarantee plugins will be flushed after the new `EnhanceCtx` call. Basically we have to make sure to always pair an `EnhanceCtx` call with a `Flush` call, wrapping the code in between. My guess is that instead of adding additional enhance calls as you’ve done here, we will need to move `Flush` later in the process (after the events have been processed). Does this make sense?

adngdb:
@delucis thank you for your input! You helped me figure out the right solution. :relaxed: The problem is that the events are actually triggered during the `Flush` call, so I couldn't "move it later". But, that made me try to explain the problem in my head: when it comes to flushing the `events` plugin, the `random` plugin has already been flushed, and thus is unavailable. Thinking about it like that, the solution seems obvious: we just need to revert the flushing order of plugins, so that `events` comes before `random`. It even makes sense: if you consider that a plugin should know about other plugins declared before it, then when removing them we should do it backwards.

So, a simple call to `reverse()` on the plugins array in `Flush` did the trick! 

Now, I assume my tests are not in the correct place, or written correctly. If you could tell me where to put them, that'd be great! Thanks!

nicolodavis:
Hi @adngdb, thanks for looking into this.  The tests can go in `src/plugins/main.test.js`.

adngdb:
Alright, done! I think it's ready for a review. :relaxed:

adngdb:
I do love me some symmetry, and I had forgotten about `turn.onBegin`. That's done now! :)

delucis:
@adngdb Nice work!

PR #657, opened 2020-04-30
Make package.json scripts work on Windows

c-w:
Currently there are several bashisms in the various package.json scripts which makes it impossible to work on boardgame.io natively on Windows without using a Linux emulation layer such as WSL. This commit replaces the bashisms with cross-platform Node commands.

#### Checklist


nicolodavis:
Thanks for taking the time to do this!!  Will take a closer look in a few days (busy this week).

package.json
```diff
@@ -110,6 +113,7 @@
     "lint-staged": "^8.1.0",
     "node-persist": "^3.0.4",
     "nodemon": "^1.18.9",
+    "npm-run-all": "^4.1.5",
```
ksjogo, line 116: Does one need npm-run-all and concurrently? They seem to cover pretty much the same functionality as far as I can tell.

scripts/changelog.js
```diff
@@ -0,0 +1,36 @@
+// Generates docs/CHANGELOG.md
+// Run this right after the npm version command.
+
+const { EOL } = require('os');
+const shell = require('shelljs');
```
ksjogo, line 5: Why have I not been using shelljs before? :D

package.json
ksjogo, line 116: Does one need npm-run-all and concurrently? They seem to cover pretty much the same functionality as far as I can tell.

c-w: Replaced `concurrently` with `run-p` in 7a0c04f since `concurrently -m1` didn't approximate `run-s` for me.

nicolodavis:
@c-w Made a few changes.  Let me know if they look good to you.

The main addition is to run "build" in prestart so you can just do:
```
npm install
npm start
```

when you first clone the repo.

c-w:
@nicolodavis Changes LGTM.

delucis:
Thanks Clemens!

PR #658, opened 2020-04-30
feat(server): Enable passing through SocketIO options

c-w:
This pull request (1) exposes the SocketIO object in the server package so that clients can construct their own instantiation of the transport and pass it through to the Server constructor and (2) enables passing through configuration options to the underlying socket.io server.

These two changes combined enable running boardgame.io in hosting environments such as Azure App Service which require setting certain SocketIO configuration options to be set like `perMessageDeflate` (see [Azure docs](https://docs.microsoft.com/en-us/azure/app-service/containers/app-service-linux-faq#language-support)).

#### Checklist


src/server/index.ts
```diff
@@ -65,6 +66,7 @@ interface ServerOpts {
   authenticateCredentials?: ServerTypes.AuthenticateCredentials;
   generateCredentials?: ServerTypes.GenerateCredentials;
   https?: HttpsOptions;
+  socketOpts?: SocketOptions;
```
nicolodavis, line 69: This should probably be a part of the transport option on line 65.  The server doesn't have to use sockets, so we shouldn't bake in options that assume that it does.

c-w: That works for me too. I updated this pull request to expose SocketIO in the package so that clients can configure it themselves and pass it through to the Server constructor.

c-w: Oh, I saw that this is already exposed in the main package. Closing this PR.

c-w: Note that with the SocketIO transport exposed, it would also be possible to remove the `https` and `authenticateCredentials` arguments to the Server constructor. Let me know if you'd like me to make this change and update the documentation accordingly.

delucis: @c-w I think it’s OK to leave these here for now. It’s especially nice to have `authenticateCredentials` alongside `generateCredentials` without needing to manually instantiate the transport class.

Issue #663, opened 2020-05-02
stop destructuring state on the React client


nicolodavis:
We currently destructure the state object on the React client: [source](https://github.com/nicolodavis/boardgame.io/blob/master/src/client/react.js#L122).

This allows you to access `G` as `props.G` rather than having to do `props.state.G`.  However, it does cause name clashes between properties in the game state and other props that are passed to the React client (for example `plugins`).

The proposal is to stop destructuring the state object.

nicolodavis:
One thing that could be confusing about this is that `state` means something in React, so `props.state` is confusing.  We'd have to rename it to something like `gameState`, which could also be confused with just the user-controlled game state `G`.

adngdb:
Maybe use a sort of namespace instead of `state`? Something like `bgio`?

delucis:
I personally don’t mind overly the fact that `props.state` is similar to the React naming convention. It’s most confusing in class components where you could have `this.state` and `props.state`, but with function components it’s less likely there will be a variable actually named `state`.

This is also a prop that will very likely be destructured early in the component tree, so users will likely not be passing a `state` prop between lots of components. In fact, I would suggest showing something like this in documentation:

```js
function Board (props) {
  const { G, ctx } = props.state;
  return (
    <div>
      <Score value={G.score} />
      <TurnCounter count={ctx.turn} />
    </div>
  )
}
```

If you do think an alternative name is essential, might `store` be viable?

It might also make sense to discuss how to structure namespaces for plugins more generally before deciding for sure. Currently the situation we’re trying to resolve is:

```js
{
  plugins: {}, // plugin data
  plugins: {}, // plugin dispatchers
}
```

If we add support for client plugins as briefly mentioned in #632, we might end up with a _third_ `plugins` namespace. This RFC would resolve the current conflict, but how would we resolve that?

delucis:
Another question around the interface for plugins on the client:

Currently state is available at `plugins[name].data`, which mirrors the data/api distinction of plugins, but if we’re _not_ going to combine these on the client — e.g. `{ data, dispatch, ...etc }` — the `data` namespace might be unnecessarily verbose. `state.plugins[name]` should probably point directly to the data.

JaredReisinger:
[...]

Issue #662, opened 2020-05-02
move signature change


nicolodavis:
There is currently a [proposal](https://github.com/nicolodavis/boardgame.io/pull/629#issuecomment-616900537) to change the signature of moves that could benefit from some feedback from the community:

###### current

```
move: (G, ctx, ...args) => {},
```

###### proposed

```
move: ({ G, ctx }, ...args) => {},
```

The rationale for this change is that it is currently difficult to add new API's without polluting `ctx`.  This causes difficulties in the codebase (resulting in `ctx` carrying API's like `random` or `events` in some contexts, but being a plain object in others).

Under the new proposal, all API's could just be added to the first arg passed to the move, so you would access the `random` API like this:

```
move: ({ G, ctx, random }, ...args) => {},
```

While this is a breaking change, compatibility can be ensured in the short term via a simple plugin that wraps incompatible moves:

```
fnWrap: (fn) => ({ G, ctx, ...api }, ...args) => {
  const ctxWithAPI = { ...ctx, ...api };
  return fn(G, ctxWithAPI, ...args);
},
```

adngdb:
That looks good to me. I like moving plugins outside of `ctx`, especially now that I had to take a look at how it is implemented under the hood. Having `ctx` be a plain object with a structure that does not vary seems like a very good thing to do (also, it will make it easier to document).

As for a backward incompatibility, that comes with the "beta" status of the engine, and I hope everyone using it is fine with that. It just needs to be made super clear. As a reference, I really like how [create-react-app](https://github.com/facebook/create-react-app/releases) handles changelogs, it's really easy to see what you have to do when updating dependencies. 

Maybe we could ship the compatibility plugin in code, but that might backfire, with people enabling the plugin and never actually doing the migration work, leading to a point where it's even more broken for them and harder to migrate. I think it would be good to show those 4 lines of code in the migration documentation, while enforcing that it should be a temporary solution and folks should migrate their move functions asap.

delucis:
A few extra thoughts:

- We should probably extend this to other values that are not strictly part of `ctx` — I can think of `playerID` for now — for consistency with the client.

- Should we destructure the plugins or have an interface like `{ G, ctx, plugins }`?

nicolodavis:
I'm in favor of destructuring plugins.  We can add a check to prevent plugins from being named 'G' or 'ctx' if that's a concern.

+1 to also moving `playerID` out of `ctx`.

PR #666, opened 2020-05-02
fix(api): Expose gameover metadata in rooms endpoints.

### LINKED ISSUE #665: Game over metadata is not exposed by API
This is a follow-up to #645, where the information that a game is over was stored in the associated room's metadata. However, that data is never sent to the client via the API, so it's not usable. 

This covers adding the `gameover` metadata of each room to the return value of the `/games/:name` route.

adngdb:
This adds the `gameover` metadata, if it exists, to both the all rooms and single room query endpoints.

Fixes #665.

#### Checklist

### DIFF
--- a/src/server/api.test.ts
+++ b/src/server/api.test.ts
@@ -1137,6 +1137,7 @@ describe('.createApiServer', () => {
                 },
               },
               unlisted: gameID === 'bar-4',
+              gameover: gameID === 'bar-3' ? { winner: 0 } : undefined,
             },
           };
         },
@@ -1156,6 +1157,7 @@ describe('.createApiServer', () => {
         },
       });
     });
+
     describe('when given 2 rooms', () => {
       let response;
       let rooms;
@@ -1179,6 +1181,11 @@ describe('.createApiServer', () => {
         expect(rooms[0].players).toEqual([{ id: 0 }, { id: 1 }]);
         expect(rooms[1].players).toEqual([{ id: 0 }, { id: 1 }]);
       });
+
+      test('returns gameover data for ended game', async () => {
+        expect(rooms[0].gameover).toBeUndefined();
+        expect(rooms[1].gameover).toEqual({ winner: 0 });
+      });
     });
   });
 
@@ -1200,6 +1207,7 @@ describe('.createApiServer', () => {
                   credentials: 'SECRET2',
                 },
               },
+              gameover: { winner: 1 },
             },
           };
         },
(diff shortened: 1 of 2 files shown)


adngdb:
I'm wondering about standardizing the room identifier variable to `gameID`, as that makes it confusing from an external point of view. The API has a concept of "game", which is a type of game that can be played, and of "room" which is one instance of a given game. By talking about `gameID` when listing or creating a room, it becomes confusing for users, in my opinion. So, if we're going to introduce a breaking change anyway, I'd rather push for `roomID` and use that everywhere in the Lobby / API.

Note though that maybe what we actually mean to use in this case is the concept of a "match" instead of room / game?

nicolodavis:
I think some of the confusion stems partly from the idea of wanting to replay a game with the same set of people, which suggests some sort of room.

I like the idea of standardizing on match. It's also used in other game servers like Colyseus and Nakama. You don't need the concept of a room because you can just transport users from their existing match to the next one (which is what the current implementation does anyway).

delucis:
Sounds good to me! Then let’s go with `matchID`. This is a slightly more extensive change as we’ll need to update the React Lobby connection class and should probably also use `nextMatchID` in the play again endpoint (instead of `nextRoomID`). Let me know if you have any questions about those, Adrian.

adngdb:
Hey, I'm coming back to this. I want to split the work into two parts, because I think they are very different. This PR ensures the exposure of the gameover metadata (and all metadata, actually) and refactors that into a function. 

It "solves" the naming issue by simply exposing all three possible identifiers: `gameID, `roomID` and `matchID`. I think that is a fine transition until I get to the work of using `matchID` everywhere, which I intend to do later. 

What do you think about this approach?

delucis:
Thanks Adrian! I think this is OK for the interim, but we probably shouldn’t actually release a version in this state.

@nicolodavis What do you think about the following: release v0.39.11 with the currently unreleased patches, then plan the following breaking changes for v0.40.0:

- Standardise to `matchID`, merging this PR
- Convert move signatures from `(G, ctx, ..args)` to `({ G, ctx, ...plugins }, ...args)` (#662)
- Decide on changes to the React client props (#663)

nicolodavis:
Yes, that sounds good. I'll get out a release now.

nicolodavis:
Just noticed the following warning when releasing:
```
(!) Circular dependency: src/plugins/main.ts -> src/plugins/plugin-
immer.ts -> src/core/reducer.ts -> src/plugins/main.ts
(!) Circular dependency: src/core/game.ts -> src/plugins/main.ts ->
 src/plugins/plugin-immer.ts -> src/core/reducer.ts -> src/core/gam
e.ts
```

Haven't had time to investigate which commit introduced this.

adngdb:
Ah, I see the base branch is now `0.40.0`, can we thus merge this? It will be easier for to work on the follow-up if we do.

delucis:
@adngdb Done! Thanks for your patience. We can keep working towards these breaking changes on 0.40.0 if you’d like to tackle the standardisation to `matchID`.

PR #667, opened 2020-05-04
docs(phases): Expand documentation of phase.endIf

delucis:
This adds more detail to the “Moving between phases” section of the documentation to clarify how to return a `{ next }` object from a phase’s `endIf`.

I’ve also updated the types to reflect that `endIf` can return `{ next }`.


delucis:
Semaphore is failing with seemingly no log output, but everything builds/passes locally — I’m guessing there’s a problem on their end.

PR #685, opened 2020-05-12
Add pagination plugin to the docs

### LINKED ISSUE #683: Add link to the next page in the documentation
I am a new user so I am currently going through the documentation. When I was done reading the concepts, I wanted to jump to the next step but didn't find any pointers at the end of the page.

I would suggest adding a link to the next page at the bottom of the documentation's pages.

Let me know what you think, I would be happy to look into it and submit a PR if you are ok with that.

Zhouzi:
This PR adds the pagination plugin to the docs, following the conversation in #683.

While testing the docs manually, I noticed that the pagination was not showing up when accessing the documentation's `homepage` from the "Documentation" link on the homepage.

So I spent some time figuring it out and the problem comes from the fact that the documentation's `homepage` has two URLs: `/` and `/README`. This issue has been reported in [docsifyjs#1131](https://github.com/docsifyjs/docsify/issues/1131) and the recommended solution is to change the link to the homepage in the sidebar from `concepts.md` to `/`.

I updated the sidebar and it does work now. I also tested the old URL (`/README`) to make sure that it still works and it does. The pagination is not showing up on `/README` though. It's probably fine as there are no references to that URL in the documentation anymore. I think it's ok for people still accessing this URL for some reason to not have the plugin.

Here's a screenshot of the bottom of the concepts page:

localhost_3000_documentation_

Closes #683

### DIFF
--- a/docs/documentation/index.html
+++ b/docs/documentation/index.html
@@ -54,6 +54,7 @@
     subMaxLevel: 2,
     auto2top: true,
     search: 'auto', // default
+    pagination: {}, // default
     markdown: {
       renderer: {
         code: function(code, lang) {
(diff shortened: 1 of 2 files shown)


delucis:
@Zhouzi Thanks!

PR #688, opened 2020-05-12
feat(plugins): Let moves return INVALID_MOVE after mutating G

### LINKED ISSUE #687: Discard changes to G if a move is invalid
Immer is super handy when making complex changes in a move rather than using immutable update patterns in a deep nested object. However, I have a few occasions when I'd like to discard some changes I've made if it turns out the move is invalid.

For instance, when a card is played in a particular way, a corresponding token is put on the board in a starting space and then is moved to another space on the board. It's easier to put the token on the board and then check if the move is valid so I modify G and run the check, but if it's invalid and I return `INVALID_MOVE`, Immer rightfully complains.

I tried messing with Immer's `original` function but it doesn't seem to do what I want. Ideally, before I return `INVALID_MOVE` I'd like to make as if I hadn't modified G at all and reset it to its previous state. That sounds like something that should be possible but I can't figure out how.

delucis:
Closes #687

This changes the immer plugin to handle returning `INVALID_MOVE` after mutations have already been applied to `G`. This is different from the approach I initially considered in #687, but is actually conceptually a little simpler.

Basically, if a move returns `INVALID_MOVE` we return `undefined` to immer, which therefore accepts any state mutations, but still return `INVALID_MOVE` instead of the bad state update to the reducer.

### DIFF
--- a/src/plugins/plugin-immer.ts
+++ b/src/plugins/plugin-immer.ts
@@ -7,7 +7,8 @@
  */
 
 import produce from 'immer';
-import { Plugin } from '../types';
+import { AnyFn, Ctx, Plugin } from '../types';
+import { INVALID_MOVE } from '../core/reducer';
 
 /**
  * Plugin that allows using Immer to make immutable changes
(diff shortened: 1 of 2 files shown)


MathieuLoutre:
I'm so glad we found a solution for this! Thanks @delucis :)

src/plugins/plugin-immer.ts
```diff
@@ -7,15 +7,29 @@
  */
 
 import produce from 'immer';
-import { Plugin } from '../types';
+import { AnyFn, Ctx, Plugin } from '../types';
+import { INVALID_MOVE } from '../core/reducer';
```
nicolodavis, line 11: We should move this constant into it's own file to avoid the circular dependency:

```
(!) Circular dependency: src/plugins/main.ts -> src/plugins/plugin-
immer.ts -> src/core/reducer.ts -> src/plugins/main.ts
(!) Circular dependency: src/core/game.ts -> src/plugins/main.ts ->
 src/plugins/plugin-immer.ts -> src/core/reducer.ts -> src/core/gam
e.ts
```

Issue #552, opened 2020-03-04
Problem with LocalMaster using previous instantiation of a Client


julienroulle:
I've had some trouble with the Local use of the lib.

When I first instantiate a Local Game like that with bots coming from another screen:
```
bots = {'0': MyBot}
var GameApp = new Client({
        game: MyGame,
        board: MyBoard,
        multiplayer: new Local({
          bots,
        }),
      });
```
When going back to my "parameters screen" and changing the turn of my bot, meaning changing the following var:
```
bots: {'1': MyBot}
```

I discovered that my new GameApp will use the previous parameters of my 1st instantiation.

I managed to find the cause of the problem, but I'm not sure why this is causing a problem.
It comes from this [if-else statement](https://github.com/nicolodavis/boardgame.io/blob/9fc13f15af3b095fa6e57ccd2dd851477024ab86/src/client/transport/local.js#L213-L221).

By commenting out the first block of the if-statement, I managed to make it work as I wanted to be (meaning instantiating a new Local Master with my new given parameters). But as I am not really familiar with this part of the lib, I'm not sure what this if-statement is here for, and what damaged I could do by removing it.

Please share with me your knowledge about that if possible :)

julienroulle:
Looks like commenting out this block breaks a test for Local Multiplayer without bots ..
So I don't know yet how to tackle this problem.

julienroulle:
I managed it make it work by replacing the following:
```
const localMasters = new Map();
export function Local(opts) {
  return transportOpts => {
    let master;

    if (localMasters.has(transportOpts.gameKey)) {
      if (
        localMasters
          .get(transportOpts.gameKey)
          .storageAPI.games.has(transportOpts.gameID)
      ) {
        master = localMasters.get(transportOpts.gameKey);
      } else {
        master = new LocalMaster({
          game: transportOpts.game,
          bots: opts && opts.bots,
        });
        localMasters.set(transportOpts.gameKey, master);
      }
    } else {
      master = new LocalMaster({
        game: transportOpts.game,
        bots: opts && opts.bots,
      });
      localMasters.set(transportOpts.gameKey, master);
    }

    return new LocalTransport({ master, ...transportOpts });
  };
}
```

But for some reasons it makes the `multiplayer interactions` test of `client.test.js` fail ..

julienroulle:
Doing this:
```
const localMasters = new Map();
export function Local(opts) {
  return transportOpts => {
    let master;

[...]

PR #691, opened 2020-05-14
Add localStorage support (close #679)

### LINKED ISSUE #679: localStorage as Local storage.
What do you think about idea to allow localStorage to be storage for localMaster? Sometimes it could be useful to be able to back to same game after accidental tab close or refresh without need of server in backend.

pociej:
Closes #679 

* Local can now be called with Local({persist : true }) and use localStorage

* Add documentation note

#### Checklist

#### TODO : 

- In case of local master reset in debug UI doesn't work regardless of localStorage usage. Without localStorage is wasn't that bad as one can simply refresh the browser but with localStorage usage it can be problem.


src/client/transport/local.js
```diff
@@ -208,13 +210,13 @@ const localMasters = new Map();
 export function Local(opts) {
   return transportOpts => {
     let master;
-
-    if (localMasters.has(transportOpts.gameKey) & !opts) {
```
nicolodavis, line 212: See https://github.com/nicolodavis/boardgame.io/issues/552 for the context of why `!opts` was added originally.

src/client/transport/local.js
```diff
@@ -208,13 +210,13 @@ const localMasters = new Map();
...
+    if (localMasters.has(transportOpts.gameKey)) {
       master = localMasters.get(transportOpts.gameKey);
     } else {
       master = new LocalMaster({
         game: transportOpts.game,
         bots: opts && opts.bots,
+        persist: opts && opts.persist,
```
nicolodavis, line 219: Maybe allow passing the localstorage key to be used here? That way if the user wants to "wipe" the data they can just switch the key to a new value.

src/server/db/localstorage.ts
```diff
@@ -0,0 +1,53 @@
...
+  constructor(key: string) {
+    super();
+    const inCache = initFromCache(key);
+    inCache.forEach((pair: any[]) => {
+      this.set(pair[0], pair[1]);
+    });
+    this.key = getStorageKey(key);
+  }
+  sync(): void {
+    globalAny.localStorage.setItem(
```
nicolodavis, line 16: What's the purpose of `globalAny`? Is it just to avoid adding an interface for `localStorage`?

nicolodavis (dismissed):

src/client/transport/local.js
pociej: Still no reason to have `!opts` guard. Should be eventually `!(opts.bots)` with explanation in comment why this is here.

src/client/transport/local.js
pociej: Hmm, will think probably you are right it could be useful.

src/server/db/localstorage.ts
pociej: Ah, thats because im typescript novice, and when i implemeted this outside the repo i had some error from ts about `locaStorage` usage most probably related to something in `tsconfig`. Will remove.

src/client/transport/local.js
pociej: @nicolodavis @delucis i would be happy to see your comment to above so i will be able to finish this PR. Thx.

nicolodavis: Let's change it to `!opts.bots` for now and we can eventually add a `forceCreate / createNewMaster` option later.  Thanks!

nicolodavis:
@pociej What's the status of this PR? No hurry, just wanted to make sure you're not waiting on me.

pociej:
@nicolodavis i just pushed things you requested to change. Hard to say why from three hours i see "coverage/coveralls Expected — Waiting for status to be reported".

pociej:
@nicolodavis @delucis What is status of review? No rush, just wanted to know.

nicolodavis:
> @nicolodavis @delucis What is status of review? No rush, just wanted to know.

Hi @pociej, I've taken note of your change. I won't be able to look into it until the weekend (a bit busy till then).

Also, a quick look suggests that there are some conflicting files. Can you sync your branch?

delucis:
@pociej Re: the pending Coveralls check, if you can resolve conflicts and sync your branch as Nicolo suggested, that should also transfer over all the new GitHub Action configurations and run the checks successfully 🤞😅

src/server/db/localstorage.ts
```diff
@@ -0,0 +1,61 @@
...
+  constructor(key: string) {
+    super();
+    const inCache = initFromCache(key);
+    inCache.forEach((pair: any[]) => {
+      this.set(pair[0], pair[1]);
+    });
+    this.key = getStorageKey(key);
+  }
+  sync(): void {
+    localStorage.setItem(this.key, JSON.stringify(Array.from(this.entries())));
```
HydraOrc, line 14: `this.key` is `undefined `on the first call before even the game inits
should add something like
```
if (!this.key) {
  return;
}
```

HydraOrc:
I have a global question about the feature.
Let's imagine that you want to have only one active game overall at the same time and the old one comes to an end (game over).
What exactly would you call to clear the storage so you will be able to restart your game once again?

src/server/db/localstorage.ts
pociej: You are right. Is fixed.

pociej:
> I have a global question about the feature.
> Let's imagine that you want to have only one active game overall at the same time and the old one comes to an end (game over).
> What exactly would you call to clear the storage so you will be able to restart your game once again?
> I know that I can check for gameOver and the do
> 
> ```
> localStorage.removeItem('bgio_log');
> localStorage.removeItem('bgio_state');
> ```
> 
> but what if the persist key is different and dynamicly generated?

You are able to provide `storageKey` so `Local({ persist:  true : storageKey : `myPrefix`})` and then in local storage you will see prefixed keys so `bgio_myPrefix_log` `bgio_myPrefix_state` and so on. Is that what you are looking for? I don't see good reason to have cleaning mechanism to be built in the library, you can easy achieve it on your own. Or i don't understand your request properly.

pociej:
@delucis im not familiar with coveralls and therefore im not sure what means last failing check, and how to fix it. All the tests seems to be ok, so why it is not mergable? Can you close now the PR?

delucis:
@pociej Coveralls reports on test coverage and errors if coverage decreases. Looks like there’s [a line in `localstorage.ts` that isn’t covered by any tests](https://coveralls.io/builds/34105590/source?filename=src/server/db/localstorage.ts#L9) and also [an uncovered branch in `appendStorageKeyPrefix`](https://coveralls.io/builds/34105590/source?filename=src/server/db/localstorage.ts#L64). You would need to add tests for those to fix that failing check.

I’ll try to go through all these changes soon!

delucis (changes requested):
Hi @pociej, here a few small changes to make. I’ve pushed a commit simplifying the `LocalStorage` classes quite a bit — let me know what you think.

There are still two lines missing in test coverage:

1. A check to see if `LocalStorage` or `InMemory` should be used in the local master ([Coveralls](https://coveralls.io/builds/34135068/source?filename=src%2Fclient%2Ftransport%2Flocal.ts#L94)).

2. Using the `LocalStorage` *without* passing a storage key.

package-lock.json
```diff
@@ -7590,8 +7590,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
```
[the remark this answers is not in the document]
pociej: hmm its not intentional, you are right it shouldnt be part of PR.

src/client/transport/local.ts
```diff
@@ -44,6 +45,8 @@ export function GetBotPlayer(state: State, bots: Record<PlayerID, any>) {
 interface LocalMasterOpts {
   game: Game;
   bots: Record<PlayerID, any>;
+  persist: boolean;
+  storageKey: string;
```
delucis, line 49: Oh, one more thing. `bots`, `persist` and `storageKey` should all be optional here:

```ts
interface LocalMasterOpts {
  game: Game;
  bots?: Record<PlayerID, any>;
  persist?: boolean;
  storageKey?: string;
}
```

pociej:
@delucis  i pulled last changes you pushed and it started trowing error : 4df5725a80c1fd6be11e5750f62fd0607a08db48.js?meteor_js_resource=true:9 Uncaught (in promise) TypeError: Cannot perform 'get' on a proxy that has been revoked
    at JSON.stringify (<anonymous>)
    at Map.sync (4df5725a80c1fd6be11e5750f62fd0607a08db48.js?meteor_js_resource=true:9)
    at Map.set (4df5725a80c1fd6be11e5750f62fd0607a08db48.js?meteor_js_resource=true:9)
    at u.setState (4df5725a80c1fd6be11e5750f62fd0607a08db48.js?meteor_js_resource=true:9)
    at f.onUpdate (4df5725a80c1fd6be11e5750f62fd0607a08db48.js?meteor_js_resource=true:9)
    at n.e (4df5725a80c1fd6be11e5750f62fd0607a08db48.js?meteor_js_resource=true:9)
    at 4df5725a80c1fd6be11e5750f62fd0607a08db48.js?meteor_js_resource=true:9
    at Object.dispatch (4df5725a80c1fd6be11e5750f62fd0607a08db48.js?meteor_js_resource=true:9)
    at Object.t.<computed> [as setInitialSetup] (4df5725a80c1fd6be11e5750f62fd0607a08db48.js?meteor_js_resource=true:9)
    at 4df5725a80c1fd6be11e5750f62fd0607a08db48.js?meteor_js_resource=true:169 

hmm....

delucis:
@pociej Weird. The `Cannot perform 'get' on a proxy that has been revoked` error usually shows up when users try to use a value from a move function “late”, because Immer uses proxies to wrap `G` which once they expire can’t be accessed. I don’t get why a proxy would be showing up in the storage class though. At that point it should all be plain objects, I would have thought.

pociej:
@delucis this kinda wow. Its not up to your changes i guess as I did checkout previous version that was working even on production. Now situation is that locally it works as it should but on other enviroment ( Digital Ocean droplet) built with same version of bgio it throws error mentioned above. It is called when i try to call move from another component to which i passed moved as props. Sounds like in meantime G has be changed. So it is like this : 

board component 
```class Board extends React.Component {
  render() {
    return (
      <div>
        ...
        <AntoherComponent moves={this.props.moves} />
        ...
      </div>
    );
  }
}```

and now inside `<AnotherComponent>` on some click i call `this.props.moves.myMove()` it throw `Cannot perform 'get' on a proxy that has been revoked`. Crazy unstability, will try to create minimal reproduction example.

pociej:
@delucis btw can you merge this PR so i can work against master version on boardgame.io? Hard to say why now coveralls fails after your last changes.

delucis:
@pociej There are still some things that need fixing before merging this, see [my review above](https://github.com/boardgameio/boardgame.io/pull/691#pullrequestreview-506964675). Coveralls is failing because some lines added aren’t covered in tests (see above), plus you need to revert some changes to package-lock.json.

Not sure what’s going on in your example above (especially weird if there’s a difference between environments).

delucis (dismissed):

delucis (approved):

Issue #693, opened 2020-05-15
feat: Use authenticateCredentials in Lobby API endpoints


delucis:
Currently the server supports options for custom `generateCredentials` and `authenticateCredentials` methods ([docs](https://boardgame.io/documentation/#/api/Server?id=with-custom-authentication)).

`generateCredentials` is used [when a player joins a match](https://github.com/nicolodavis/boardgame.io/blob/cfeaf6781be9fdc11f6a489dded29faca95b9395/src/server/api.ts#L197-L198), permitting for example a client to send a JSON Web Token in the request headers and have the server generate a public token to store in credentials for later.

If provided, `authenticateCredentials` is currently only passed to the game master to authenticate actions. We should also use it when authentication is needed in the lobby API. Currently this is a simple equality check, e.g.

https://github.com/boardgameio/boardgame.io/blob/cfeaf6781be9fdc11f6a489dded29faca95b9395/src/server/api.ts#L224-L226

Creating a default value for `authenticateCredentials` will have the bonus that we can centralise this check instead of having each endpoint decide to do a certain equality comparison.

PR #698, opened 2020-05-16
feat(server): Expose API router

c-w:
This pull request refactors `Server.run` to separate setting up the API routes from running the API server. Splitting these responsibilities enables developers to more easily customize hosting of the app, e.g. customizing the CORS setup, mounting the API routes at a path prefix, etc.

#### Checklist

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -288,8 +288,6 @@ export namespace Server {
   export interface LobbyConfig {
     uuid?: () => string;
     generateCredentials?: GenerateCredentials;
-    apiPort?: number;
-    apiCallback?: () => void;
   }
 }
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -289,11 +289,6 @@ export namespace Server {
     uuid?: () => string;
     generateCredentials?: GenerateCredentials;
   }
-
-  export interface LobbyRuntimeSettings {
-    apiPort?: number;
-    apiCallback?: () => void;
-  }
 }
 
 export namespace CredentialedActionShape {
--- a/src/types.ts
+++ b/src/types.ts
@@ -289,6 +289,11 @@ export namespace Server {
     uuid?: () => string;
     generateCredentials?: GenerateCredentials;
   }
+
+  export interface LobbyRuntimeSettings {
+    apiPort?: number;
+    apiCallback?: () => void;
+  }
 }
 
 export namespace CredentialedActionShape {
--- a/src/types.ts
+++ b/src/types.ts
@@ -284,11 +284,6 @@ export namespace Server {
     nextRoomID?: string;
     unlisted?: boolean;
   }
-
-  export interface LobbyConfig {
-    uuid?: () => string;
-    generateCredentials?: GenerateCredentials;
-  }
 }
 
 export namespace CredentialedActionShape {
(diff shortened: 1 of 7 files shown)


src/types.ts
```diff
@@ -288,8 +288,6 @@ export namespace Server {
   export interface LobbyConfig {
     uuid?: () => string;
     generateCredentials?: GenerateCredentials;
-    apiPort?: number;
-    apiCallback?: () => void;
```
delucis, line 292: Why are `apiPort` and `apiCallback` removed here? They’re still in use as far as I can tell.

c-w: The only place where these values seem to be used is in `Server.run` and `createServerRunConfig` (which makes sense since `apiPort` and `apiCallback` are very much runtime considerations as opposed to generic configurations). However, `createServerRunConfig` has its own type definition for `apiPort` and `apiCallback` in the `lobbyConfig` member of `ServerConfig`. This means that the values could be safely removed from the more generic `LobbyConfig` interface.

To put this in another way: previously `LobbyConfig` was really two distinct interfaces that we could have called `LobbySettings` (`uuid` and `generateCredentials`, used by the framework code) and `LobbyRuntimeSettings` (`apiPort` and `apiCallback`, to my understanding only used by `Server.run`). This pull request encapsulates the `LobbyRuntimeSettings` in `Server.run`.

delucis: Ahh, got it! Thanks. Could we type the `portOrConfig` argument to `run` on line 119 to reflect this? (It’s currently just `number | object`.)

delucis: Oh, and we should probably update the documentation with the `lobbyConfig` option to `Server`.

c-w: Done.

c-w: Note that there are now quite a few redundant arguments in the Server function arguments:

- `https` and `authenticateCredentials` are less specific than a custom `transport`.
- `generateCredentials` is less specific than `lobbyConfig`.

If you're open to breaking API changes, I would recommend to deprecate `https`, `authenticateCredentials` and `generateCredentials` in favor of `transport` and `lobbyConfig` to reduce future maintenance burden and confusion with the function arguments (e.g. which argument takes precedence). If you'd like, I can make a PR for these changes.

delucis: Thanks! Aren’t these already breaking changes as `uuid` is moved from the `lobbyConfig` passed to `server.run` to the `Server` constructor options?

I’ve done some more refactoring here to reduce some of the specificity issues you raise: https://github.com/nicolodavis/boardgame.io/tree/delucis/pr/698

If you think that looks good, I’ll add it to this PR and get this merged.

In terms of further changes, do you have any views on #693?

c-w: While I believe it was technically possible to pass `uuid` and `generateCredentials` via the `Server.run` argument, this wasn't expressed in the type system. I'll make a change to make this fully backwards compatible.

delucis: Typescript is a fairly recent addition to the library, so there are a few places where the types don’t fully reflect potential usage out there.

That said, we’re actually planning to release several breaking changes in 0.40.0 so I think this is a reasonable time to clean this up as long as we document the migration path clearly. I’m going to merge your earlier version into a new development branch from which we can then release this.

Thanks again!

PR #704, opened 2020-05-18
Fix #703 - Use the term "match" in the API and Lobby

### LINKED ISSUE #703: [API, Lobby] Standardize "room" and "game" terms to "match"
As part of #665 we discovered that, in the Lobby and API, "rooms" where sometimes called "games". Upon discussing how to standardize that, we agreed that renaming to "match" was the correct move, as it makes more sense, and is also used in other similar tools.

This issue covers:
- removing all instances of `room`, `roomID`, `game` and `gameID` from the Lobby and API;
- replacing them with either `match` or `matchID`;
- updating the documentation to match (*pun intended*) these changes.

adngdb:
Here's where I'm at with this work. I've tried to do it slowly, with a lot of small, focused commits, to make this easier to review. I know how painful this type of change is to review, sorry in advance! :)

I believe this covers all of the instances of "room" in the api and lobby, as well as all the cases where "game" was used but it actually meant "match". It is, however, likely that I have missed some instances in some corners of the app. I'm counting on you to spot them! :grin: 

Regarding documentation, I guess we'll want to explain the changes very clearly. Should that be in the changelog? I want to wait until these changes are validated before writing that doc though.

### DIFF
--- a/src/master/master.test.ts
+++ b/src/master/master.test.ts
@@ -12,7 +12,7 @@ import {
   Master,
   redactLog,
   getPlayerMetadata,
-  doesGameRequireAuthentication,
+  doesMatchRequireAuthentication,
   isActionFromAuthenticPlayer,
 } from './master';
 import { error } from '../core/logger';
@@ -619,10 +619,10 @@ describe('getPlayerMetadata', () => {
   });
 });
 
-describe('doesGameRequireAuthentication', () => {
+describe('doesMatchRequireAuthentication', () => {
   describe('when game metadata is not found', () => {
     test('then authentication is not required', () => {
-      const result = doesGameRequireAuthentication();
+      const result = doesMatchRequireAuthentication();
       expect(result).toBe(false);
     });
   });
@@ -636,7 +636,7 @@ describe('doesGameRequireAuthentication', () => {
           '0': { id: 1 },
         },
       };
-      const result = doesGameRequireAuthentication(matchMetadata);
+      const result = doesMatchRequireAuthentication(matchMetadata);
       expect(result).toBe(false);
     });
   });
@@ -653,7 +653,7 @@ describe('doesGameRequireAuthentication', () => {
           },
         },
       };
-      const result = doesGameRequireAuthentication(matchMetadata);
+      const result = doesMatchRequireAuthentication(matchMetadata);
       expect(result).toBe(true);
     });
   });
--- a/src/master/master.test.ts
+++ b/src/master/master.test.ts
@@ -590,7 +590,9 @@ describe('getPlayerMetadata', () => {
 
   describe('when metadata does not contain players field', () => {
     test('then playerMetadata is undefined', () => {
-      expect(getPlayerMetadata({} as Server.GameMetadata, '0')).toBeUndefined();
+      expect(
+        getPlayerMetadata({} as Server.MatchMetadata, '0')
+      ).toBeUndefined();
     });
   });
 
@@ -627,21 +629,21 @@ describe('doesGameRequireAuthentication', () => {
 
   describe('when game has no credentials', () => {
     test('then authentication is not required', () => {
-      const gameMetadata = {
+      const matchMetadata = {
         gameName: '',
         setupData: {},
         players: {
           '0': { id: 1 },
         },
       };
-      const result = doesGameRequireAuthentication(gameMetadata);
+      const result = doesGameRequireAuthentication(matchMetadata);
       expect(result).toBe(false);
     });
   });
 
   describe('when game has credentials', () => {
     test('then authentication is required', () => {
-      const gameMetadata = {
+      const matchMetadata = {
         gameName: '',
         setupData: {},
         players: {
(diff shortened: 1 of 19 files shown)


src/master/master.test.ts
```diff
@@ -627,21 +629,21 @@ describe('doesGameRequireAuthentication', () => {
 
   describe('when game has no credentials', () => {
```
delucis, line 630:
```suggestion
  describe('when match has no credentials', () => {
```

src/master/master.test.ts
```diff
@@ -627,21 +629,21 @@ describe('doesGameRequireAuthentication', () => {
...
           '0': { id: 1 },
         },
       };
-      const result = doesGameRequireAuthentication(gameMetadata);
+      const result = doesGameRequireAuthentication(matchMetadata);
       expect(result).toBe(false);
     });
   });
 
   describe('when game has credentials', () => {
```
delucis, line 644:
```suggestion
  describe('when match has credentials', () => {
```

src/master/master.ts
```diff
@@ -81,10 +81,10 @@ export function redactLog(log: LogEntry[], playerID: PlayerID) {
  * Verifies that the game has metadata and is using credentials.
```
delucis, line 81:
```suggestion
 * Verifies that the match has metadata and is using credentials.
```

src/master/master.ts
```diff
@@ -81,10 +81,10 @@ export function redactLog(log: LogEntry[], playerID: PlayerID) {
...
  */
 export const doesGameRequireAuthentication = (
```
delucis, line 83: This will require refactoring the call sites in `master.ts` and `master.test.ts`:

```suggestion
export const doesMatchRequireAuthentication = (
```

src/server/api.test.ts
```diff
@@ -137,7 +137,7 @@ describe('.createRouter', () => {
       });
 
       test('returns game id', () => {
```
delucis, line 139:
```suggestion
      test('returns match id', () => {
```

src/server/api.test.ts
```diff
@@ -1080,7 +1080,7 @@ describe('.createRouter', () => {
       response = await request(app.callback())
         .post('/games/foo/1/playAgain')
         .send('playerID=0&credentials=SECRET1');
-      expect(response.body.nextRoomID).toBe('12345');
+      expect(response.body.nextMatchID).toBe('12345');
     });
 
     test('when the game does not exist throws a "not found" error', async () => {
```
delucis, line 1086:
```suggestion
    test('when the match does not exist throws a "not found" error', async () => {
```

src/server/api.test.ts
```diff
@@ -1167,33 +1167,33 @@ describe('.createRouter', () => {
...
+        matches = JSON.parse(response.text).matches;
       });
 
-      test('returns rooms for the selected game', async () => {
-        expect(rooms).toHaveLength(2);
+      test('returns matches for the selected game', async () => {
+        expect(matches).toHaveLength(2);
       });
 
       test('returns room ids', async () => {
```
delucis, line 1184:
```suggestion
      test('returns match ids', async () => {
```

src/server/api.test.ts
```diff
@@ -1167,33 +1167,33 @@ describe('.createRouter', () => {
...
       });
 
       test('returns player names', async () => {
-        expect(rooms[0].players).toEqual([{ id: 0 }, { id: 1 }]);
-        expect(rooms[1].players).toEqual([{ id: 0 }, { id: 1 }]);
+        expect(matches[0].players).toEqual([{ id: 0 }, { id: 1 }]);
+        expect(matches[1].players).toEqual([{ id: 0 }, { id: 1 }]);
       });
 
       test('returns gameover data for ended game', async () => {
```
delucis, line 1194:
```suggestion
      test('returns gameover data for ended match', async () => {
```

src/server/db/base.ts
```diff
@@ -88,7 +88,7 @@ export abstract class Async {
    */
   abstract setMetadata(
     gameID: string,
```
delucis, line 90: Ideally, we should probably rename all the `gameID` parameters in this file to `matchID`.

```suggestion
    matchID: string,
```

src/server/db/base.ts
```diff
@@ -146,7 +146,7 @@ export abstract class Sync {
   /**
    * Update the game metadata.
```
delucis, line 147:
```suggestion
   * Update the match metadata.
```

src/server/db/base.ts
adngdb: I am unsure about this, because it looks like this is going to open a huge can of worms, where I'll have to go through the entire code base to determine where the word "game" is correctly used and where it isn't. I was quite content to do that only in the API / Lobby domains, as that's a smaller part and one that is not too hard to understand for me.

But maybe I'm wrong in assuming that those `db` functions are used all over the code base?

delucis: The naming here is purely cosmetic as the ID is a string parameter (not a property in an object) so I don’t think you’d need to change the call sites, but we can also leave this as is for now and tidy up when we come to it.

adngdb: Ah, right, it's an interface file! Alright then, on it. :-)

delucis:
Thanks again for this Adrian!

Issue #534, opened 2019-12-26
Horizontal scaling


JoHuang:
Hi

Does this framework support horizontal scaling?
Which means
1. stateless multiple node.js instances behind a load balancer.
2. central state storage (ex. MongoDB, redis , ...)

It seems that single point of failure might happen in this framework?
Correct me if I'm wrong.
Thanks

jasonharrison:
Hi Jo,

We do support external databases such as MongoDB (they are set through ENV).  See:

jasonharrison:
Hi Jo,

1) it is possible to run multiple BGIO servers behind a load balancer, however, you may run into issues with socket.io's websocket broadcasts.  It may be possible to overcome this (for example, with https://github.com/socketio/socket.io-redis ), but I have not looked into what this would entail.

Note that games and content can be served separately from BGIO (and served on multiple servers without issue); the only point of failure could be with multiplayer mode which uses BGIO's server.

For 2), BGIO's server does support external databases such as MongoDB (these can be set through environment variables).  See:

https://github.com/nicolodavis/boardgame.io/tree/master/src/server/db
https://github.com/nicolodavis/boardgame.io/blob/master/src/server/db/index.js

JoHuang:
For 1), if I have A, B, C 3 servers behind a load balancer, players a, b, c connected to A, B, C respectively. Can these 3 players play the same game?

If the answer is true. I have another question about 2), especially for MongoDB implementation.
There is a LRU for preventing race condition, but the LRU is only used locally. Is that be able to prevent race condition caused by different servers?

Thanks

nicolodavis:
@JoHuang There is another issue open to discuss this: #277 

You can try to use a load balancer with sticky connections to ensure that all the traffic for a particular game goes to the same server, but I think it would be better if we added support for something like SocketCluster.

nicolodavis:
@JoHuang Is this something that you'd like to contribute to?  We could definitely use some help in this area.

JoHuang:
Thanks for reply.
So the MongoDB implementation is only for record / backup, players have to connect to the server/instance to join the same game.

I'm still evaluating this framework. For now I'm not sure if I would put effect on it.

To achieve horizontal scaling, there are 2 todo items:

1. Stateless node.js instances behind load balancer. (load & save to database for every move)
2. Atomic update game states in a central state storage (Mongodb atomic ops like findAndModify).
[...]

PR #706, opened 2020-05-20
Add support for socket.io-adapter implementations

zdraganov:
Allows using socket adapter, that's one possible solution for horizontal scaling issues.

The issue where this is requested: https://github.com/nicolodavis/boardgame.io/issues/534

More information about socket.io-adapter option: https://socket.io/docs/server-api/#server-adapter-value
#### Checklist

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -65,7 +65,6 @@ interface ServerOpts {
   authenticateCredentials?: ServerTypes.AuthenticateCredentials;
   generateCredentials?: ServerTypes.GenerateCredentials;
   https?: HttpsOptions;
-  socketAdapter?;
 }
 
 /**
@@ -85,7 +84,6 @@ export function Server({
   authenticateCredentials,
   generateCredentials,
   https,
-  socketAdapter,
 }: ServerOpts) {
   const app = new Koa();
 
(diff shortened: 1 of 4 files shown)


src/server/index.ts
```diff
@@ -84,6 +85,7 @@ export function Server({
   authenticateCredentials,
   generateCredentials,
   https,
+  socketAdapter,
```
nicolodavis, line 88: There's no need to pass this to the `Server` constructor because the `SocketIO` class below is exposed to the user. So, if they want to customize the options sent they can just do this:

```js
import { SocketIO, Server } from 'boardgame.io/server';

const server = Server({
  transport: new SocketIO({ ... })
})

mariadb-zdraganov: Ah, I see. I don't think we need this at all, if we can pass the SocketIO instance to the transport. I will close this one.

nicolodavis:
Oh, it also looks like there is an uncovered line here: https://coveralls.io/builds/30941173/source?filename=src/server/transport/socketio.ts#L103

Would you mind adding a test for this?

zdraganov:
Yep, I will try to add.

PR #714, opened 2020-05-25
docs: Overhaul docs site & add vanilla JS to tutorial

delucis:
This is a pretty heavy PR — I may have gone overboard, let me know if you’d like this broken up into smaller pieces.

Here are the key changes:

- I’ve rewritten the Tic-Tac-Toe tutorial. It now features tabs with both the React code we already had and example code using the vanilla JS client. This is the thing that will need the closest review — a read-through and potentially a follow along would be ideal. (I’ve done that and I think it’s all good, but extra eyes are always good.)

    As this is also the first documentation of the vanilla client, thoughts on how it is presented would also be appreciated. Does it seem like a reasonable minimal implementation?

- In working on this and adding the tabs, I felt there were quite a few improvements possible to layout and readability on the docs site. I’ve switched the theme to [docsify-themeable](https://jhildenbiddle.github.io/docsify-themeable/#/introduction). This has a narrower column width and is generally more readable.

- There’s now an “Edit on GitHub” link on each page.

- I’ve tidied up some CSS and made various other small tweaks.


nicolodavis:
Chris, this is excellent.

Couple of high level comments as I continue to inspect the details:

1. I would suggest getting rid of `App.js` entirely and just inlining the script in the HTML file (while still retaining a separate `Game.js`). `createBoard` can probably go as well if we just add the markup directly to the HTML rather than creating it in JS. This will allow a better emphasis on the JS that the tutorial is actually trying to communicate.

2. The number of steps in the tutorial is now 8 (vs. 5 previously). Should we hide some of them as sub-steps so that they don't show up on the navigation as sub headings? This may seem inconsequential, but I think it does have a psychological effect on the reader.

delucis:
Glad you’re OK with it — got a little carried away once I started tweaking things!

1. I tried to structure the plain JS the same as the React examples to keep them as close to each other as possible, but I think I can make it work like you suggest.

2. We can certainly collapse “Validating Moves” and “Victory Condition” under one heading — “Game Improvements” or something similar. I’ll think about whether there are other possibilities for the same. I like how now the tutorial winds back and forth from client/UI to game code, so that can maybe be the points where sections are divided.

delucis:
OK, I think this is ready @nicolodavis!

I’ve also updated the Multiplayer guide with vanilla JS code examples and documented the vanilla client API.

One quick question. The React client docs mention an AI `step` method, but I couldn’t find anywhere this was created in the client code. Is it implemented?

https://github.com/nicolodavis/boardgame.io/blob/4e889011453f8623b511c24d83cbab8961144f7a/docs/documentation/api/Client.md#L122-L123

nicolodavis:
Ah, we don't expose the step function to the client anymore, so that's out of date.

Sorry, was busy last week. Will take a look at this PR again in a day or two.

delucis:
> Ah, we don't expose the step function to the client anymore, so that's out of date.

Got it — removed that line from the docs in that case. And no worries — I know this a lot to get through!

Issue #715, opened 2020-05-25
state.plugins is undefined after restoring state from debug control panel


SmileyJames:
Back again 😆 

To reproduce using tic-tac-toe example:
- Make some moves
- Save
- Restore
- Make another move

Causes TypeError: state.plugins is undefined

SmileyJames:
So only G and ctx are saved, not the whole state
https://github.com/nicolodavis/boardgame.io/blob/master/src/client/debug/main/Controls.svelte#L9
Is there a reason for this? Can we save the whole state? Can we save plugins aswell as G and ctx.

delucis:
Closed in #716

PR #716, opened 2020-05-25
Save all state to localStorage, not just G and ctx

SmileyJames:
#### Checklist

Solves #715

Consider approach, is there a reason not to commit all state to localStorage?

### DIFF
--- a/src/client/debug/main/Controls.svelte
+++ b/src/client/debug/main/Controls.svelte
@@ -6,7 +6,10 @@
   import { parse, stringify } from 'flatted';
 
   function Save() {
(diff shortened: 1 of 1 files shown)


JGrzybowski:
It sounds resonable. As Save/Restore is mainly used to store the game state for later use. We can assume that the player after some time would not be interested to restore the game state and immediately undo the move.

SmileyJames:
@delucis Thanks for wrapping this up, apologies for my lack of input - been busy with work :)

delucis:
No worries — I saw we were about to package a release and I thought it would be nice to have this included in that. Thanks again for kicking this off.

Issue #723, opened 2020-06-05
Improve client error handling


delucis:
This is a proposal for how we could improve error handling in boardgame.io clients.

**Related:** #592 #218

## Current Situation

Errors happen implicitly and there is no way for a client to handle them or even know they happened.

Errors are logged in two key places:

1. The reducer logs errors and returns the original `state` when it encounters them.

2. When in use, the game master does various validation and logs errors too, bailing when it encounters them.

A lot of this validation logic is actually duplicated across master and reducer, which may not be necessary.

## Goals

- A client should know if an error happened and why.

- Errors should be handled in local single-player instances as well as server-connected multiplayer scenarios.

- If possible, the action that triggered an error should be able to receive the error at its call site.

## Core Proposal

1. If the reducer encounters an error, instead of returning the unchanged `state` it should return `{ ...state, error }`.

2. The `error` field should never be persisted — it is a transient side effect of the most recent action specific to a single client.

3. When a master is being used, if there is an error, the `update` should not complete by sending an `update` event with the full state to all clients. Instead, the master should send a new `error` event type to the requesting client. (`error` would be a third event type alongside `update` and `sync`.)

4. The `error` field would have a shape of `{ type, payload? }`, where `type` is one of a range of boardgame.io error codes (see “Error Types” below) and `payload` is additional information where relevant.

5. If possible, client action dispatchers should return a promise that is resolved by the client once the action has been processed. If the action produces an error, the dispatcher would resolve with `{ error }`. See “Errors at call sites” below for more details.

### Error Types

<details>

<summary>A catalogue of potential error types based on current error logging, with suggested error codes.</summary>

| type                         | description                                                |
|:-----------------------------|:-----------------------------------------------------------|
| `update/unauthorized_action` | The action’s `credentials` were missing or invalid         |
| `update/match_not_found`     | The action’s `matchID` was not found                       |
| `action/stale_state_id`      | The action contained a stale state ID                      |
| `action/unavailable_move`    | The requested move is unknown or not currently available   |
| `action/invalid_move`        | The move declared it was invalid (`INVALID_MOVE` constant) |
| `action/inactive_player`     | The player making the action is not currently active       |
| `action/gameover`            | The game has finished                                      |
| `action/action_disabled`     | The requested action is disabled (e.g. undo/redo, events)  |

</details>

### Errors at call sites

<details>
<summary>Could an action dispatcher offer an async response at the call site?</summary>

[...]

PR #727, opened 2020-06-06
Fixes #713

### LINKED ISSUE #713: server sync problem
Seems that there is problem with reconnection. Im running game with local server and sometimes completly randomly without any error runs only on current client, server has no info no move is played. Only way is to restart the server (client restart ends up with 'connecting...' loader in place of client) it is like 10% of cases. Game is pretty complex so i will need to create small repo for reproduction, but seems that multiplayer is broken.  Will provide more detail once i will track down way of data from client to server ( debuggin this library is not that simple thing) but smells like problem with socket disconnection and then reconnection.

flamecoals:
Fixes #713

#### Checklist

### DIFF
--- a/src/client/transport/socketio.ts
+++ b/src/client/transport/socketio.ts
@@ -129,11 +129,10 @@ export class SocketIOTransport extends Transport {
       }
     });
 
-    // Initial sync to get game state.
-    this.socket.emit('sync', this.gameID, this.playerID, this.numPlayers);
-
     // Keep track of connection status.
     this.socket.on('connect', () => {
+      // Initial sync to get game state.
+      this.socket.emit('sync', this.gameID, this.playerID, this.numPlayers);
       this.isConnected = true;
       this.callback();
     });


jabza:
Possible to get this released soon? cheers :)

nicolodavis:
Thanks @flamecoals!

Does this fix the issue for you (I'm guessing you use an internal fork and aren't blocked by a release from us)?

I think there might still be a possibility of a desync if:
- The internet connection is slow, but the socket itself hasn't disconnected.

I think the eventual solution should probably be something like:
- The server just sends a sync response back whenever it receives a stale response from the client. Right now it just [rejects](https://github.com/nicolodavis/boardgame.io/blob/master/src/master/master.ts#L272) the response without any further action.

flamecoals:
It fixed for the case that the internet drops completely, I agree with you that it wont fix for the case of it being slow. It is quite bit hard to simulate these scenarios though, because Chrome does not apply the network settings to Websockets.

We got rid of our internal fork in order to reduce complexity, so we are still waiting for release, it would be great if it could be released :)

nicolodavis:
> Chrome does not apply the network settings to Websockets.

Yeah, I find this quite annoying as well!

> We got rid of our internal fork in order to reduce complexity, so we are still waiting for release, it would be great if it could be released :)

This is released as `0.39.13`.

delucis:
If we implement something like #723, issuing a `sync` in response to stale state would conflict with that model, but if the server can reliably fix errors like this, that would be a better experience. Perhaps #723 should consider if there any other situations like this that can “self heal”?

flamecoals:
Thanks for the release!

Issue #735, opened 2020-06-08
API: allow to query only active games


janKir:
Currently, the `GET /games/:game` API endpoint lists all existing games. It includes also games that are already finished and also outdated games that were started days ago but never finished.
It would be nice to allow handing query parameters to the endpoint to only receive active / recent games.

I know this filtering could be done on the client but this doesn't resolve the issue that with every new game, this response is getting bigger and bigger.

delucis:
Yes, this would be nice. There are some things that are relatively easy to add like filtering out completed matches by checking the metadata’s `gameover` field. How would you suggest measuring active vs inactive? We would probably need to add some kind of `lastActive` or `modified` timestamp to the metadata and update it on every game action.

Ideally, this would also be implemented in the Storage API as most backends would probably support filtering entries more efficiently than looping over the full list on the server.

janKir:
I think it would make sense anyway to add timestamps like `createdAt`, `updatedAt` and `finishedAt`.
Doing it in the storage sounds good, but how should we handle the potential incompatiblity to existing storage packages?

delucis:
We can add additional fields to the metadata during the game creation process here:

https://github.com/nicolodavis/boardgame.io/blob/41d77da5c846a9a375e1be510985edd678010c21/src/server/api.ts#L47-L51

And update the type here:

https://github.com/nicolodavis/boardgame.io/blob/41d77da5c846a9a375e1be510985edd678010c21/src/types.ts#L284-L292

In terms of filtering, it should be possible to add this without breaking other storage implementations if we do it by adding to the options object passed to `db.listGames`:

https://github.com/nicolodavis/boardgame.io/blob/41d77da5c846a9a375e1be510985edd678010c21/src/server/db/base.ts#L37-L39

Implementations that haven’t been updated will simply return all games instead of filtering, which is probably OK until they can be updated. (This is actually already the case with the bundled `FlatFile` which [ignores the listGames options](https://github.com/nicolodavis/boardgame.io/blob/0.40.0/src/server/db/flatfile.ts#L151-L157).)

Would you be interested in trying to make this work? If possible, it would be best to base your work on the [`0.40.0` branch](https://github.com/nicolodavis/boardgame.io/tree/0.40.0) as there are already some breaking changes to the REST API there.

janKir:
What do you think how the query options should look like? And how extensive should they be?
I can think of something like this, but not sure if we really need all of those:
```typescript
export interface ListGamesOpts {
[...]

PR #740, opened 2020-06-13
allow to filter listGames query

janKir:
This PR ( based on issue #735 ) adds `createdAt` and `updatedAt` fields to the metadata containing numeric timestamps. Furthermore, the following query parameters were given to the listGames endpoint: `isGameover`, `updatedBefore`, `updatedAfter`.

#### remaining tasks

* [x] implement filter in the included storage adapters

### DIFF
--- a/src/server/api.ts
+++ b/src/server/api.ts
@@ -48,6 +48,8 @@ export const CreateMatch = async ({
     gameName: game.name,
     unlisted: !!unlisted,
     players: {},
+    createdAt: Date.now(),
+    updatedAt: Date.now(),
   };
   if (setupData !== undefined) metadata.setupData = setupData;
   for (let playerIndex = 0; playerIndex < numPlayers; playerIndex++) {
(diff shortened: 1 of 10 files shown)


src/server/api.ts
```diff
@@ -155,7 +157,42 @@ export const createRouter = ({
    */
   router.get('/games/:name', async ctx => {
     const gameName = ctx.params.name;
-    const matchList = await db.listGames({ gameName });
+    const {
+      isGameover: isGameoverString,
+      updatedBefore: updatedBeforeString,
+      updatedAfter: updatedAfterString,
+    } = ctx.query;
```
delucis, line 164:
```suggestion
    } = ctx.request.body;
```

The other API endpoints use JSON body parameters instead of the query string, so we should be consistent here. This should also simplify the parameter parsing logic that follows.

janKir: Is this even possible for get requests?

delucis: Oh, right… sorry, I should have noticed that — of course everything else is using JSON body, but those are all POST requests, so the query string makes sense here! In that case, I think we’re all good. 👍

PR #745, opened 2020-06-15
build: Fix installing boardgame.io as a git dependency

delucis:
When installing as a git dependency, e.g. `npm i nicolodavis/boardgame.io`, the prepack script is run but will error because development dependencies aren’t available. This PR adds an `npm install` step to the prepack pipeline, which admittedly will generally be redundant, but I don’t see a way to only run it for git dependency installs and it shouldn’t cause any unwanted behaviour in other scenarios.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "lint:fix": "eslint --fix .",
     "prepublishOnly": "npm run clean",
     "proxydirs": "node scripts/proxy-dirs.js",
-    "prepack": "run-s build proxydirs",
+    "prepack": "npm i && run-s build proxydirs",
     "postpack": "npm run clean",
     "prettier": "prettier --write \"{examples,src,benchmark}/**/*.{ts,js,css,md}\"",
     "changelog": "node ./scripts/changelog.js",


iamsumanth:
This PR doesn't solve completely. Fails at,
```
> cross-env BABEL_ENV=rollup rollup --config rollup.config.js
[!] (plugin babel) Error: Cannot find module '@babel/compat-data/corejs3-shipped-proposals'
```

delucis:
@iamsumanth Hmm, interesting — I retried the install with an empty cache and got the same error. Ran some more tests and the build seems to succeed under Node v10 but not 12 or 14, so this could be an environment issue although I’m not 100% sure.

Sorry but I think this is probably the most we can do for now short of a release and there can always be delays with those. If the git dependency isn’t working you might have better luck building the repo manually using `npm pack` and sharing the archive somehow, but I can’t guarantee it.

iamsumanth:
Thanks a lot of trying @delucis, I appreciate your help  :) 
I will try something like npm pack or some other option, anyway looks like release would happen in a few days. So I can wait for that too.

package.json
```diff
@@ -28,7 +28,7 @@
     "lint:fix": "eslint --fix .",
     "prepublishOnly": "npm run clean",
     "proxydirs": "node scripts/proxy-dirs.js",
-    "prepack": "run-s build proxydirs",
+    "prepack": "npm i && run-s build proxydirs",
```
nicolodavis, line 31: This would also run `npm install` prior to a `npm publish` I think? Not sure if that's necessarily a good idea (a dependency might get updated that causes something to break that you didn't catch while testing, for example).

delucis: Yes, it would. There doesn’t seem to be a simple way to do this only for git dependency installs without potentially interfering with the actual build/publish steps, so perhaps it’s best just to accept that git installs aren’t supported for now and encourage linking for local development where necessary.

PR #749, opened 2020-06-20
build: Fix installing boardgame.io from GitHub

nicolodavis:
#### Checklist

Safer version of https://github.com/boardgameio/boardgame.io/pull/745.


package.json
```diff
@@ -28,7 +28,8 @@
     "lint:fix": "eslint --fix .",
     "prepublishOnly": "npm run clean",
     "proxydirs": "node scripts/proxy-dirs.js",
-    "prepack": "run-s build proxydirs",
+    "github:deps": "npm i --no-save npm-run-all",
+    "prepack": "npm run github:deps && run-s build proxydirs",
```
delucis, line 32: Tested this out and it does seem to work although I don’t quite understand why we don’t need to install cross-env, rollup etc. that are used in the build script.

Builds still fail for me under Node 12 & 14 — `(plugin babel) Error: Cannot find module '@babel/compat-data/corejs3-shipped-proposals'` — but it works under Node 10. (This was the same in #745.)

If it’s as simple as needing npm-run-all, would the following do the same and simplify this?

```suggestion
    "prepack": "npx run-s build proxydirs",
```

nicolodavis: Interesting.

After it applies this fix, it now complains about `cross-env` not being installed. I think what's happening is that `cross-env` is pulled in by `npm-run-all`.

nicolodavis: I've reverted the change.

delucis:
Some further thoughts given that the origin of some dependencies seems a bit mysterious when just installing `npm-run-all`:

As far as I understand, `npm i` in a project folder where `node_modules` matches what is specified in `package-lock.json` is a no-op. If we did include `npm i` in the prepack script (as in #745), it would only actually trigger a package update on publish if `package-lock.json` had been updated upstream and `npm i` was never run on the publishing machine. In this situation, we probably shouldn’t be building & publishing anyway.

If the main concern about a full install is potential discrepancy between tested and published code, could we run the tests as part of the prepack pipeline? It would slow down installing as a git dependency but the case where `npm publish` triggers a package install that would fail tests would be caught.

Issue #763, opened 2020-07-18
Improve consistency around “match” vs “game”


delucis:
There is still some work needed to make the framework fully consistent around using the term “match” to describe an instance of a specific game (see #709). Should we aim to get this done as part of 0.40.0?

- [x] Database connectors (would require updating the various third-party connectors):

  - [x] `createGame` method should be renamed `createMatch`

  - [x] `listGames` method should be renamed `listMatches`

  Done in #806

- [x] `gameID` → `matchID`

  - [x] in React client

  - [x] in plain JS client

  - [x] in debug panel

  - [x] in master

  - [x] in server transport (cosmetic argument renaming)

- [x] `gameMetadata` → `matchMetadata`

  - [x] in React client

  - [x] in plain JS client

  (This had in fact been done in #704, because the 0.40.0 docs have been updated to show `matchMetadata`, but I think I may have overwritten those changes when merging in the Typescript clients.)

- [x] `updateGameID` → `updateMatchID` (plain JS client)

There may be some more I’ve missed, but I think that’s most of it.

evandroabukamel:
Hey, @delucis , I'm working on that if you don't mind. I'm opening a PR.

PR #765, opened 2020-07-19
feat: Improve consistency around “match” instead of “game”

delucis:
This PR addresses #763, updating client and transport implementations to use `matchID`, `matchData`, `updateMatchID`, `subscribeMatchData`, etc. instead of using “game”.

So far I didn’t include the refactoring of the Storage API as that might require some further consideration, so I’ll await feedback before doing anything about that.

#### Clients should now be passed a `matchID` instead of `gameID`

```js
import { Client } from 'boardgame.io/react';

const App = Client({ game });

export default () => <App matchID="xyz" />;
```

```js
import { Client } from 'boardgame.io/client';

const client = Client({
  matchID: 'xyz',
  // ...
});

client.updateMatchID('abc');
```

#### React board components will receive `matchData` instead of `gameMetadata`

```js
const Board = ({ matchData }) => {
  return <div>{matchData}</div>
};
```

### DIFF
--- a/docs/documentation/api/Client.md
+++ b/docs/documentation/api/Client.md
@@ -100,7 +100,7 @@ The following properties are available on a client instance:
 - `credentials`: Multiplayer authentication credentials for this player.
 
 
-- `matchMetadata`: An array containing the players that have joined
+- `matchData`: An array containing the players that have joined
   the current match via the [Lobby API](/api/Lobby.md).
 
   Example:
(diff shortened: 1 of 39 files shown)


nicolodavis:
@delucis: Do you think it's worth shortening `matchMetadata` to `matchData`?

I feel like it's unambiguous enough (unlike `gameData`).

delucis:
> Do you think it's worth shortening `matchMetadata` to `matchData`?

Sure, I don’t see why not.

delucis:
Done in 3cde0a7 & 15bd0ab.

PR #767, opened 2020-07-21
build: Exclude svelte from Rollup’s external module list

### LINKED ISSUE #766: error loading react client
after 72ed5913 i cant load the react client:

```
Error: Function called outside component initialization
```

which traces back to svelte calling `get_current_component` after importing svelte in [src/client.js](https://github.com/boardgameio/boardgame.io/blob/72ed591310bec944cbb3b7eba458726b810b3a1d/src/client/client.ts#L9)

delucis:
Fixes #766

@nicolodavis I’ve found it a little tricky to test this, but looking through the bundle output, this removes the Svelte imports introduced in 0.39.15, and I could get a demo project to build. Sorry for all the false starts getting the Svelte–Typescript integration working, I’ve been learning the hard way to double check my work…

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -17,8 +17,9 @@ import pkg from './package.json';
 import typescript from 'rollup-plugin-typescript2';
 const subpackages = require('./subpackages');
 
+const internalDeps = ['svelte'];
 const external = [
-  ...Object.keys(pkg.dependencies),
+  ...Object.keys(pkg.dependencies).filter(name => !internalDeps.includes(name)),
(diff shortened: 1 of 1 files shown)


nicolodavis:
@delucis Taking a step back, what are we losing by keeping 'svelte' in `devDependencies`?

delucis:
Typescript support for consumers. Users would have to manually install Svelte so that its ambient type declaration for `'*.svelte'`, which the client type declaration depends on, is available.

PR #780, opened 2020-08-19
Enforce that the debug menu is always on the top of the screen

### LINKED ISSUE #778: Increase z-index of the debug panel
At the moment, elements with a high z-index can be rendered over the side debug panel.  It would be useful to increase it to some high value so that it's always on top.

popey456963:
Only change here is adding `z-index` to alter the debug-menu render height.

Fixes #778


delucis (approved):

delucis:
Thanks @popey456963!

PR #781, opened 2020-08-19
Replaces the side debug bar with a live object view

### LINKED ISSUE #775: Hide some parameters in debug menu
We pass some fairly large objects through the global state variable.  Some items, such as the map, contain hundreds of entries.  It would be useful to have a way to filter items seen through the debug menu.

popey456963:
Fixes #775

This pull request replaces with the `JSON.stringify()` state and ctx sections with a live `JSONTree` view, which allows you to minimise / maximise attributes.  This should help cases where the state is very large.

We use a custom package that is two lines different from the default to:

- Expand objects by default.
- Expand arrays that are less than 1024 characters in length by default.

Otherwise `ctx` and `state` start minimised even if they only contain a few elements which is infuriating.  A ticket has been raised on the original project to ask if they can provide a solution in-package.

> Note: This PR is currently a draft because it causes tests to fail.


popey456963:
The latest commit replaces the `import JSONTree from 'svelte-json-tree-auto';` line with `import JSONTree from 'svelte-json-tree';`, which is the default `svelte-json-tree` package available from the svelte team themselves.  This is just to ensure that the issue is not caused by my packaging of the package.

src/client/debug/main/Main.svelte
```diff
@@ -1,6 +1,8 @@
 <script>
   export let client;
 
+  import JSONTree from 'svelte-json-tree';
```
delucis, line 4: If you take a look at [the `dist` code from `svelte-json-tree`](https://unpkg.com/browse/svelte-json-tree@0.1.0/dist/index.js), it’s a CommonJS module (little hard to see on the minimised output, but prettify it and scroll to the bottom and you’ll find `module.exports = Root` with `Root` being the entry component).

I think normally when the Svelte compiler is run on the Debug panel, it resolves `svelte-json-tree` by looking at its `package.json` and following the `svelte` field there (which points to `src/Root.svelte`). All good. However, my suspicion is that under Jest for some reason `svelte-json-tree` gets resolved as a regular non-Svelte module, so we now hit `dist/index.js`, but because that’s a CommonJS module and not an ES module, the import returns `undefined` and throws a “not a constructor” error when the code tries to instantiate `JSONTree`.

So, a solution:

```suggestion
  const JSONTree = require('svelte-json-tree');
```

I think this assures that we import the CommonJS module everywhere (local tests, build & examples work). I don’t know if this is the *ideal* fix, because perhaps depending on this results in a less optimised bundle than compiling the source Svelte component (I guess there’s some boilerplate shared code in the CommonJS transpiled version).

@nicolodavis Do you have any thoughts here? Is there a way we can get Jest to do the right thing in terms of resolving this import?

delucis (changes requested):
Thanks for this @popey456963. And I think I’ve figured it out 🤞

src/client/debug/main/Main.svelte
delucis: Hmm, I hadn’t noticed that plugin, but it doesn’t seem to be possible to disable it just for a certain file or import. (I’m also not 100% sure if it comes into play. `Main.svelte` is presumably handled by `jest-transform-svelte`, which already outputs CJS modules, so perhaps the import transformation is happening there?) For now it seems like the `require` fix is acceptable. If there is some minimal bump in bundle size, this is only going to be in the development environment in any case.

package.json
```diff
@@ -165,6 +165,8 @@
     "socket.io": "^2.1.1",
     "socket.io-client": "^2.3.0",
     "svelte": "^3.24.0",
+    "svelte-json-tree": "^0.1.0",
```
delucis, line 168: Let’s remove `svelte-json-tree` and go with `svelte-json-tree-auto` until https://github.com/tanhauhau/svelte-json-tree/issues/8 gets addressed. (Thanks for creating that fork!)

package.json
```diff
@@ -190,7 +192,7 @@
       "^.+\\.svelte$": "jest-transform-svelte"
     },
     "transformIgnorePatterns": [
-      "node_modules/(?!(boardgame.io|flatted|svelte-icons)/)"
+      "node_modules/(?!(boardgame.io|flatted|svelte-icons|svelte-json-tree-auto|sveslte-json-tree)/)"
```
delucis, line 195:
```suggestion
      "node_modules/(?!(boardgame.io|flatted|svelte-icons|svelte-json-tree-auto)/)"
```

src/client/debug/main/Main.svelte
```diff
@@ -1,6 +1,8 @@
 <script>
   export let client;
 
+  const JSONTree = require('svelte-json-tree');
```
delucis, line 4:
```suggestion
  const JSONTree = require('svelte-json-tree-auto');
```

src/client/debug/main/Main.svelte
```diff
@@ -1,6 +1,8 @@
...
+  // import JSONTree from 'svelte-json-tree-auto';
```
delucis, line 5:
```suggestion
```

src/client/debug/main/Main.svelte
```diff
@@ -36,6 +38,11 @@
     color: #888;
```
delucis, line 38: Let’s remove the `.json` style block (and classes) as it’s no longer needed with the JSON tree styling.

src/client/debug/main/Main.svelte
```diff
@@ -107,12 +114,14 @@
 
 <section>
   <label>G</label>
-  <pre class="json">{JSON.stringify(G, null, 2)}</pre>
+  <div class="json tree">
```
delucis, line 117:
```suggestion
  <div class="tree">
```

src/client/debug/main/Main.svelte
```diff
@@ -107,12 +114,14 @@
...
+    <JSONTree value={G} />
+  </div>
 </section>
 
 <section>
   <label>ctx</label>
-  <pre class="json">
-    {JSON.stringify(SanitizeCtx(ctx), null, 2)}
-  </pre>
+  <div class="json tree">
```
delucis, line 124:
```suggestion
  <div class="tree">
```

delucis (changes requested):
Hi @popey456963 — I’ve merged in that change to use `require` instead of `import` when including the JSON Tree component, which has fixed the failing tests. Here are a few small changes to tidy this up and get it ready to merge. Cheers!

delucis (approved):

Issue #558, opened 2020-03-09
"ctx.events.setPhase('MyPhase')" seems to have issues


Oliboy50:
I try to [migrate my playing card game](https://github.com/Oliboy50/coinche/pull/11) from boargame.io `0.32.1` to `0.37.2` (first my target was `0.33.0` but it behaves exactly the same).

So I followed the migration guide, but after a lot of debugging, I can't understand why it doesn't work now...

My game is split into several phases (Deal, Talk, PlayCard, CountPoint, etc.) and a lot of logic is done through `onBegin` phase hooks (Deal and CountPoint phases for instance, doesn't contain any move).

Here is a **simple** reproduction of my main issue:
```ts
{
  // ...
  phases: {
    Deal: {
      start: true,
      onBegin: (G, ctx) => {
        console.log('phase Deal');
        ctx.events.setPhase('Talk');
      },
    },
    Talk: {
      onBegin: (G, ctx) => {
        console.log('phase Talk');
        ctx.events.setPhase('PlayCards');
        console.log('TalkEnd');
      },
    },
    PlayCards: {
      onBegin: (G, ctx) => {
        console.log('phase PlayCards');
        ctx.events.setPhase('Talk');
        console.log('PlayCardsEnd');
      },
    },
  },
}
```

then my logs are:
```
phase Deal
phase Talk
TalkEnd
```

it just doesn't want to immediately go to `PlayCards` phase after the `Talk` one and I don't understand why... (it worked fine in version `0.32.1`)

could you help?

(maybe this is relative to https://github.com/nicolodavis/boardgame.io/issues/526)

delucis:
From the sample of code you shared, it looks like you’re ending `PlayCards` immediately and returning to the `Talk` phase. So something like this happens:

[...]

PR #796, opened 2020-09-08
chore(deps): upgrade Immer from 1.8.0 to 7.0.4

adngdb:
Hey folks,

I noticed that the version of Immer used here was very outdated (almost 2 years old), and I was hoping to use a feature introduced in version 7 (`current`). I thus figured I could give upgrading Immer a try, and see if it broke anything. As far as tests are concerned, not much broke, and it was very easy to fix. So, here's a PR to upgrade it! 

However, I am very unsure of this. I thought you might have reasons for not upgrading, and maybe I'm missing some very important details? 

In order to help with this, I compiled below the list of all breaking changes in Immer since version 1.8.0:

### [7.0.0](https://github.com/immerjs/immer/compare/v6.0.9...v7.0.0) (2020-06-10)

- [BREAKING CHANGE] getters and setters are now handled consistently:  own getters and setters will always by copied into fields (like  Object.assign does), inherited getters and setters will be left as-is.  This should allow using Immer directly on objects that trap their  fields, like done in Vue or MobX. Fixes [#584](https://github.com/immerjs/immer/issues/584), [#439](https://github.com/immerjs/immer/issues/439), [#593](https://github.com/immerjs/immer/pull/593), [#558](https://github.com/immerjs/immer/pull/558)
- [BREAKING CHANGE] produce no longer accepts  non-draftable objects as first argument
- [BREAKING CHANGE] original can only be called on drafts and will throw otherwise (fixes [#605](https://github.com/immerjs/immer/issues/605))
- [BREAKING CHANGE] non-enumerable and symbolic fields will never be frozen
- [BREAKING CHANGE] the patches for arrays are now computed differently to fix some scenarios in which they were incorrect. In some cases they  will be more optimal now, in other cases less. Especially splicing /  unshifting items into an existing array might result in a lot of  patches. Fixes [#468](https://github.com/immerjs/immer/issues/468)

### [6.0.0](https://github.com/immerjs/immer/compare/v5.3.6...v6.0.0) (2020-03-03)

- Support for ES5, patches and Map/Set collections has to be *explicitly* enable now: https://immerjs.github.io/immer/docs/installation
- Custom serialization hooks are no longer supported

### [5.0.0](https://github.com/immerjs/immer/compare/v4.0.2...v5.0.0) (2019-10-30)

BREAKING CHANGE: Maps and Sets are treated differently now; they will no longer directly mutated when updated inside a draft

### [4.0.0](https://github.com/immerjs/immer/compare/v3.3.0...v4.0.0) (2019-09-11)

- any value returned from produce will be deeply frozen in DEV, even when it wasn't frozen before the producer started. Fixes [#412](https://github.com/immerjs/immer/issues/412) [#363](https://github.com/immerjs/immer/issues/363) [#260](https://github.com/immerjs/immer/issues/260) [#230](https://github.com/immerjs/immer/issues/230) [#313](https://github.com/immerjs/immer/issues/313) [#229](https://github.com/immerjs/immer/issues/229) through [#417](https://github.com/immerjs/immer/pull/417)

### [3.0.0](https://github.com/immerjs/immer/compare/v2.1.5...v3.0.0) (2019-04-17)

- **ts:** typescript@3.4.0 is now the minimum required version
- `this` in recipe is no longer the draft, but unbound ([3f40298](https://github.com/immerjs/immer/commit/3f40298))
- Dropped node 6 support ([eecc773](https://github.com/immerjs/immer/commit/eecc773))

### [2.0.0](https://github.com/mweststrate/immer/compare/v1.12.1...v2.0.0) (2019-02-02)

- added migration guide: Promises returned from producer will be evaluated

### Checklist


nicolodavis:
Hi @adngdb, thanks a lot for this!

nicolodavis (approved):

Issue #810, opened 2020-09-26
Moves from debug panel are disallowed when playing with local master


jorgebg:
## Sumary
When running the game with a local master the moves triggered from the debug panel raise the following error:

```
ERROR: disallowed move: moveName
```

## Steps to reproduce:

1. Go to the local master docs and open the example in SandBox
https://boardgame.io/documentation/#/multiplayer?id=local-master

2. In the `Client`, set `debug` to `true` and reload the app:
https://codesandbox.io/s/boardgameio-dibw3?file=/src/index.js:551-556

3. Open the developer console.

4. Select player 0 by clicking on it from the debug panel (otherwise it will raise `ERROR: player not active - playerID=[1]` as player 1 is selected by default, not sure why):

![image](https://user-images.githubusercontent.com/606597/94339619-7a2f0f00-fffb-11ea-8bc8-90fa975ce16b.png)

5. Click on the `clickCell` move from the debug panel, enter `0` (or leave it empty), and press `enter`:

![image](https://user-images.githubusercontent.com/606597/94339585-389e6400-fffb-11ea-8cad-c5b130218a5e.png)

6. Observe the error in the developer console:

```
ERROR: disallowed move: clickCell
```

I forked the example project following the steps to reproduce so you can take a look right away:
https://codesandbox.io/s/boardgameio-forked-1yihh

delucis:
Thanks for the clear report and reproduction steps!

This is actually a symptom of another bug/error: in this case each client is rendering the Debug Panel (one on top of the other). That’s why player 1 is selected by default — because player 1’s debug panel is rendered second and appears on top. We actually need to fix this some other way by rendering one debug panel per _master_ instead of per client (or potentially have some way of toggling between debug panel instances if there are several).

jorgebg:
Thank you, the fix worked! I also activated the debug panel only in the first client, as you suggested, and then everything worked as expected. Thanks again!

PR #812, opened 2020-09-27
fix(core): Fix EndPhase deltalog

jimu:
A side effect of calling `EndPhase()` is an automatic call to `EndTurn()`. LogEntries for `endPhase` and `endTurn` log entries are added to `deltalog`.

When the user mouses over `LogEntries` in the **Debug Panel**, the state is regenerated/reduced from the deltalogs. `EndPhase()` *again* makes an automatic call to `EndTurn()`, and then the `endTurn` in the `deltalog` is also called, resulting in
one-too-many calls to `EndTurn()`.

This issue affects games using `DEFAULT` turn order and causes playback fail when using the `Debug:Log` Panel.

This error is generated to the browser console:
```
   ERROR: disallowed move: MYMOVENAME
```
where *MYMOVENAME* is the name of the move that fails.

Appears to correct issue #810

Issue also reported:
 * [Gitter 2020-06-23 07:05](https://gitter.im/boardgame-io/General?at=5ef1f017b8152d34845bace8)
 * [Gitter 2020-07-08 02:16](https://gitter.im/boardgame-io/General?at=5f05778fc7d15f7d0f7b1e01)

### DIFF
--- a/src/core/flow.ts
+++ b/src/core/flow.ts
@@ -438,7 +438,7 @@ export function Flow({
 
   function EndPhase(state: State, { arg, next, turn, automatic }: any): State {
     // End the turn first.
-    state = EndTurn(state, { turn, force: true });
+    state = EndTurn(state, { turn, force: true, automatic: true });
 
     let G = state.G;
     let ctx = state.ctx;


delucis (approved):
Good catch @jimu!

PR #816, opened 2020-10-06
feat: Add global manager to track client & debug panel instances

### LINKED ISSUE #670: Debug UI buggy ....
... in case of multiple clints on local master. 
Screenshot 2020-05-06 at 11 05 53

When i click 'End Phase' it throws error `ERROR: player not active - playerID=[3]` 
This is becuase my table component is like this : 

```    
    <div id="table">
      <Client gameID="123" playerID="0" />
      <Client gameID="123" playerID="1" />
      <Client gameID="123" playerID="2" />
      <Client gameID="123" playerID="3" />
    </div>
```
When i change it to be 
```  
    <div id="table">
      <Client gameID="123" playerID="3" />
      <Client gameID="123" playerID="2" />
      <Client gameID="123" playerID="1" />
      <Client gameID="123" playerID="0" />
    </div>
```

Then it works, it first case i have to make few moves, and wait till 'activePlayer' and 'currentPlayer' is same. Whatever the difference between active and current is ( i dont se it yet ) it cant be like that ability to end phase depends on elements orders in DOM. Also when i click on 'End Phase' when it is allowed phase is not finished properly. `ctx.phase`shows updated value but, list of allowed moves is not updated, this is for sure related to https://github.com/nicolodavis/boardgame.io/issues/669

### LINKED ISSUE #810: Moves from debug panel are disallowed when playing with local master

delucis:
This PR fixes bugs created by rendering multiple debug panels when there are multiple clients. It does so by:

- Adding a `ClientManager` class, a global instance of which is used by all clients to register/unregister as they start/stop.

- Rendering the debug panel is delegated to this manager, which can therefore ensure only one panel at a time is rendered.

- I’ve added a client “switcher” to the bottom of the main debug pane whenever there is more than one active client to allow users to toggle the client the debug panel is connected to. (You can see this in action by running the examples and checking out the tic-tac-toe multiplayer example.)

  Screen capture showing use of the client switcher

  There’s no simple way of identifying a client, so the menu shows various bits of client info to try to help clarify which one’s which.

- When switching player IDs, the manager also checks to see if there’s an active client for that player and switches clients before toggling the `playerID` for the current client.

The main positive here is that confusion around the debug panel when using multiple clients should be solved. For example, this fixes #810 and closes #670.

The main drawback is that the `ClientManager` class is only currently used to bridge between the clients and the debug panel, but it will be bundled and instantiated even in production builds where (in most cases) the debug panel won’t be used. Also, it’s a global instance. I’m not sure if there’s another way to do this, but would be interested in feedback or thoughts for different approaches.


nicolodavis:
Looks good @delucis!

I can't think of another way to do the client switching without storing global state.

nicolodavis (approved):

PR #819, opened 2020-10-07
fix(client): Fix playerView executed twice do not strip again if is multiplayer

### LINKED ISSUE #818: playerView executed twice for multiplayer game
When using playerView to calculate results from secret info, boardgame.io crashes.

Here is a minimal reproduce.
https://codesandbox.io/s/boardgameio-typescript-player-view-issue-demo-f-hefjq?file=/src/Game.ts

larry801:
Fix #818 
This fix disable strip on client if current client is multiplayer .
#### Checklist

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -261,6 +261,41 @@ describe('multiplayer', () => {
   });
 });
 
+describe('strip secret only on server', () => {
+  let client0;
+  let client1;
+  let spec;
+  let initial = { secret: [1, 2, 3, 4], sum: 0 };
+  beforeAll(() => {
+    spec = {
+      game: {
+        setup: () => initial,
+        playerView: (G, ctx, playerID) => {
+          let r = { ...G };
+          r.sum = r.secret.reduce((prev, curr) => {
+            return prev + curr;
+          });
+          delete r.secret;
+          return r;
+        },
+        moves: { A: (G, ctx) => ({ A: ctx.playerID }) },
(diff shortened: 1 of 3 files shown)


src/client/client.test.ts
```diff
@@ -261,6 +261,41 @@ describe('multiplayer', () => {
...
+      multiplayer: Local(),
+    };
+
+    client0 = Client({ ...spec, playerID: '0' });
+    client1 = Client({ ...spec, playerID: '1' });
+
+    client0.start();
+    client1.start();
+  });
+
```
delucis, line 292: Could we add an equivalent `afterAll` block that calls `stop` for the two clients to complete cleanup?

src/client/client.ts
```diff
@@ -435,7 +435,14 @@ export class _ClientImpl<G extends any = any> {
     // Secrets are normally stripped on the server,
     // but we also strip them here so that game developers
     // can see their effects while prototyping.
-    const G = this.game.playerView(state.G, state.ctx, this.playerID);
+    let G;
```
delucis, line 438: It might be cleaner to use a ternary operator to avoid reassignment here:

```js 
const G = this.multiplayer
  ? state.G
  : this.game.playerView(...)
```

delucis (changes requested):
Here are a couple of small things, but this looks basically good!

delucis (approved):
Thanks @larry801!

PR #826, opened 2020-10-13
refactor(lobby): Convert React components to Typescript

c-w:
Similar to @coyotte508 in https://github.com/boardgameio/boardgame.io/issues/190#issuecomment-617446765, I recently ran into an issue importing `boardgame.io/client` due to the missing types in the Lobby React components. As such, this pull request converts the Lobby React components to Typescript which should fix the import error.

#### Checklist

### DIFF
--- a/src/lobby/react.tsx
+++ b/src/lobby/react.tsx
@@ -9,6 +9,7 @@
 import React from 'react';
 import Cookies from 'react-cookies';
 import PropTypes from 'prop-types';
+import { DebugOpt } from '../client/client';
 import { Client } from '../client/react';
 import { MCTSBot } from '../ai/mcts-bot';
 import { Local } from '../client/transport/local';
@@ -36,7 +37,7 @@ type LobbyProps = {
   gameComponents: GameComponent[];
   lobbyServer?: string;
   gameServer?: string;
-  debug?: boolean;
+  debug?: DebugOpt | boolean;
   clientFactory?: typeof Client;
   refreshInterval?: number;
   renderer?: (args: {
--- a/src/lobby/react.tsx
+++ b/src/lobby/react.tsx
@@ -8,20 +8,64 @@
 
 import React from 'react';
 import Cookies from 'react-cookies';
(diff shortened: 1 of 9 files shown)


src/lobby/react.tsx
```diff
@@ -8,20 +8,64 @@
...
+  matchID: string;
+  playerID: string;
+  credentials?: string;
+};
+
+type LobbyProps = {
+  gameComponents: GameComponent[];
+  lobbyServer?: string;
+  gameServer?: string;
+  debug?: boolean;
```
delucis, line 38: This should probably align with the `debug` option that can be passed to the client functions:

https://github.com/boardgameio/boardgame.io/blob/af435611685d72e3ea6e1a81909c9bb6522a18dd/src/client/client.ts#L104

This allows users to pass the full debug panel implementation if they want to include it in production builds ([docs](https://boardgame.io/documentation/#/debugging?id=using-the-debug-panel-in-production)).

src/lobby/react.tsx
```diff
@@ -42,16 +86,7 @@ const LobbyPhases = {
  *   A React component that provides a UI to create, list, join, leave, play or
  *   spectate matches (game instances).
  */
-class Lobby extends React.Component {
-  static propTypes = {
```
delucis, line 46: Could we keep prop types for this public-facing component? Keeping prop types gives some development warnings for developers who aren’t using Typescript, which might be better than nothing for them.

delucis (changes requested):
Nice work! A couple of small questions, but otherwise this looks great.

src/lobby/react.tsx
c-w: Done in 3bacaef.

src/lobby/react.tsx
c-w: Done in a994a43.

delucis (approved):

delucis:
Thanks @c-w!

PR #832, opened 2020-10-20
feat(simultaneous-moves): add option to long form move to ignore stale stateID (close #828)

### LINKED ISSUE #828: Allow simultaneous moves
Well, I participated in a conversation about this in Gitter and just read an old one. The thing is:

Two or more players make a move "in the same time" but just one is executed, the others throws an "Invalid stateID" error on the server.

Now that my card game, "cards against humanity"-ish, is in beta test we saw this problem occurring a lot of times and it's hard to handle it on the client since the server does not return the error. I think one way to handle this is to return the error, there is an issue open about that. Although, I believe that is not the best way to solve that as the clients will receive errors, that may cause a bad experience, so I believe making the server handle moves with the same stateID is a good approach, of course checking if the move is still allowed. Maybe we could make this optional.

What do you think? I will look the the code and test some implementations.

Follows the talks from Gitter that I found about this issue:

> Chris Swithinbank @delucis out 03 2019 07:47
Nice!
Question: I’ve been wondering about race conditions on the server now that players can be active simultaneously. What happens if two players make a move “at the same time”? Does the server know how to handle that or could this be a source for bugs? Are there good strategies for dealing with this given that bg.io is deterministic and dependent on order of execution.

> Nicolo John Davis @nicolodavis out 03 2019 08:09
There is a stateID that is passed along with each action.
The server will only accept an action from a client that provides it the correct stateId, which it subsequently increments once the move is applied.
If two moves happen simultaneously (i.e. they're both on the wire at the same time without one client seeing the update from the other) then the server will reject the move that arrives second.
This is an exceedingly rare possibility and I don't think it's worth building any intelligent reconciliation logic.
However, I think we should surface the error on the client. I don't think we do currently.

> Chris Swithinbank @delucis out 03 2019 08:42
OK — I wondered whether one approach would be to declare that moves were not order/state dependent and allow just those to be processed “late”. (Similar to declaring if a move can be undone or should be redacted.)
As an example: The project I’m working on includes a messaging interface that needs to be in game state, because other players can play “spy” cards and read their opponents’ discussions. The sendMessage move is enabled most of the time, but it doesn’t really matter if it happens before or after other moves (or other sendMessage calls for that matter), so could safely be made “late” as I describe.
This is probably a really really uncommon use-case, so I wouldn’t propose implementing something like this, but I was curious what the current behaviour was.
(In simple terms: two moves that don’t touch the same state — including the PRNG — can be safely made in any order.)

> Nicolo John Davis @nicolodavis out 03 2019 09:00
Yeah, I'm definitely open to marking moves that can be safely made out of order.
And have the server ignore the stateID for those cases.

> Chris Swithinbank @delucis out 03 2019 09:16
Cool — like I said, definitely not a priority. Surfacing a client error for rejected moves would be helpful (in general). Is there an open issue for that? I thought there was but I can’t seem to find it.

Another talk:
> Ruslan Hydra
@HydraOrc
jul 28 14:09
Guys, how do you ensure that you do not run into race conditions when you allow both players to do moves simultaniously at the same turn or also if one player issues a couple moves at the same time (ERROR: invalid stateID, was=[11], expected=[13])?

> Chris Swithinbank
@delucis
jul 29 06:25
This is an area where the framework still needs some work. Regarding the first issue, the logic for this was written before Stages allowed simultaneous actions and it hasn’t been updated to handle that scenario (it simply blocks the action it sees second as you probably experienced). With the invalid stateID error, I think we could improve that too by queueing player actions on the server or something.

> Ruslan Hydra
@HydraOrc
jul 29 18:03
Maybe make server master sync instead of async? Is that a possible try to solve?

> Chris Swithinbank
@delucis
jul 29 19:01
I would have to dive into the implementation to check the options. Keeping it async would probably be best (to handle multiple matches in parallel), but probably use a per-match queue, so if an action arrives while another is already being processed we can wait and then check if the second action is still valid.

> Evandro Abu Kamel
@evandroabukamel
jul 29 20:19
That's a problem I'm going to have soon. The game we're developing allows players to make moves simultaneously. The idea of a move queue is excellent.

evandroabukamel:
#### Solves issue #828 .

Adding `ignoreStateStateID` option to `LongFormMove`.
Exporting `IsLongFormMove` function from game.ts.
On master.ts, getting the move and checking if it does not have the `ignoreStateStateID` option thuthy before triggering invalid stateID error.

#### Checklist


evandroabukamel:
The file `src/server/index.ts` has a lower coverage and is causing coveralls to fail.

delucis:
> The file `src/server/index.ts` has a lower coverage and is causing coveralls to fail. I didn't work on it.

That’s OK — you don’t need to try to fix the `src/server/index.test.ts` coverage as part of this PR :-)

delucis (changes requested):

evandroabukamel:
I've refactored the code for getMove on master.ts and improved the unit test for simultaneous moves.

src/master/master.test.ts
```diff
@@ -202,6 +202,72 @@ describe('update', () => {
...
+      'A',
+      ['not ignore stale state'],
+      '0'
+    );
+    const actionB = ActionCreators.makeMove('B', ['ignore stale state'], '0');
+
+    // test: ping-pong two moves, then sync and check the log
+    await master.onSync('matchID', '0', 2);
+    await master.onUpdate(actionA, 0, 'matchID', '0');
+    await master.onUpdate(actionB, 0, 'matchID', '0');
```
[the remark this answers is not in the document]
evandroabukamel: Well, I made that through unit tests and it didn't happened.  :smile:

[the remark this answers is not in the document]
delucis: The test is using the synchronous `InMemory` storage, so I’m not sure it fully captures the scenario above. Maybe it would be possible with some wrapper around `InMemory` that adds some artificial delays. Something like:

```ts
class AsyncInMemory extends StorageAPI.Async {
  db: InMemory;

  constructor() {
    super();
    this.db = new InMemory();
  }

  private sleep(interval: number) {
    return new Promise(resolve => void setTimeout(resolve, interval));
  }

  async createMatch(id: string, opts: StorageAPI.CreateMatchOpts) {
    await this.sleep(100);
    return this.db.createMatch(id, opts);
  }

  // etc.
}
```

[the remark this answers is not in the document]
evandroabukamel: Well, I've implemented the AsyncInMemory class and that strange behaviour you predicted happened: one player had it's action not executed or overwritten. I believe there's something regarding the order of execution of the promises. Strange that on my game, using Flatfile, it didn't happened.

```
class InMemoryAsync extends StorageAPI.Async {
  db: InMemory;

  constructor() {
    super();
    this.db = new InMemory();
  }

  async connect() {
    await this.sleep(100);
  }

  private sleep(interval: number): Promise<void> {
    return new Promise(resolve => void setTimeout(resolve, interval));
  }

  /**
   * @param id
   * @param opts
   * @override
   */
  async createMatch(id: string, opts: StorageAPI.CreateMatchOpts) {
    await this.sleep(100);
    this.db.createMatch(id, opts);
  }

  async setMetadata(matchID: string, metadata: Server.MatchData) {
    await this.sleep(100);
    this.db.setMetadata(matchID, metadata);
  }

  async setState(matchID: string, state: State, deltalog?: LogEntry[]) {
    await this.sleep(100);
    this.db.setState(matchID, state, deltalog)
  }

  async fetch<O extends StorageAPI.FetchOpts>(
    matchID: string,
    opts: O
  ): Promise<StorageAPI.FetchResult<O>> {
    await this.sleep(100);
    return this.db.fetch(matchID, opts);
  }

  async wipe(matchID: string) {
    await this.sleep(100);
    this.db.wipe(matchID);
  }

  /**
   * @param opts
   * @override
   */
  async listMatches(opts?: StorageAPI.ListMatchesOpts): Promise<string[]> {
    await this.sleep(100);
    return this.db.listMatches(opts);
  }
}
```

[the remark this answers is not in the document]
evandroabukamel: Maybe we should implement a per-match queue system on the infrastructure level. I was thinking to use the PQueue package (https://www.npmjs.com/package/p-queue). What do you think?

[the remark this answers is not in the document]
delucis: Thanks for testing this @evandroabukamel!

I actually thought about the `FlatFile` storage first (when I was suggesting to write a test with async storage), but remembered that someone added some concurrency management to `FlatFile` (#705). So I expect `FlatFile` would not produce this bug, but other asynchronous storage implementations could, because they won’t generally implement any queuing.

I guess a per-match queue would need to be added to [the server transport layer](https://github.com/boardgameio/boardgame.io/blob/9fcbed76d3fdd196b47e052d4ae2543e7d517a74/src/server/transport/socketio.ts#L121-L130).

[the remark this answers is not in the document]
evandroabukamel: That's what I thought but it will not solve the case on the unit test since it is using the master directly. Wouldn't it be better to create a queue on the master.ts?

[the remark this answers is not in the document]
evandroabukamel: Ok, @delucis, I implemented a per-match queue where you said, taking care to delete the queue when no more players are connected to a given match. Unfortunately I'm not sure how to write unit tests for that. :disappointed: 
I'm gonna try it on my game to see using InMemory database to see if it works.

src/server/transport/socketio.ts
```diff
@@ -126,7 +129,11 @@ export class SocketIO {
             TransportAPI(matchID, socket, this.clientInfo, this.roomInfo),
             this.auth
           );
-          await master.onUpdate(action, stateID, matchID, playerID);
+
+          this.createMatchQueue(matchID);
+          await this.perMatchQueue[matchID].add(() =>
```
delucis, line 134: Is it actually possible to access a map entry like this without using `get`? I wonder if you could implement a `getMatchQueue` method instead of `createMatchQueue` which will make sure the match queue exists and then return it for use.

src/server/transport/socketio.ts
```diff
@@ -162,9 +169,33 @@ export class SocketIO {
...
       });
     }
   }
+
+  /**
+   * Create a PQueue for a given matchID.
+   * @param matchID
+   */
+  createMatchQueue(matchID: string): void {
+    if (!this.perMatchQueue.get(matchID)) {
```
delucis, line 187: You can use `has` instead of `get` to check if the key exists.
```suggestion
    if (!this.perMatchQueue.has(matchID)) {
```

src/server/transport/socketio.ts
```diff
@@ -162,9 +169,33 @@ export class SocketIO {
...
+      this.perMatchQueue.set(matchID, new PQueue({ concurrency: 1 }));
+    }
+  }
+
+  /**
+   * Delete a PQueue for a given matchID.
+   * @param matchID
+   */
+  deleteMatchQueue(matchID: string): void {
+    if (this.perMatchQueue.get(matchID)) {
```
delucis, line 197: As above, you can use `has` here.
```suggestion
    if (this.perMatchQueue.has(matchID)) {
```

delucis (commented):
Hi @evandroabukamel — how’s this going? It looks pretty good to me, we probably just need to include a test using an asynchronous storage (like the `AsyncInMemory`) somehow to make sure the queue protects the master from race conditions as expected.

src/master/master.test.ts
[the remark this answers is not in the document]
delucis: Oh, sorry, I missed this comment. Yes, testing this is pretty tricky. The socket tests are already pretty thin (there are no assertions in [the sync/update tests](https://github.com/boardgameio/boardgame.io/blob/d071e98c0868ea687d85168649e5d6302fb15229/src/server/transport/socketio.test.ts#L180-L194)) and I’m not immediately sure how to do improve them. I guess if we test the queue sufficiently, checking `onUpdate` is called sequentially, that should be enough. I can try and take a look when I have time to see if I can work out how this might be done.

src/server/transport/socketio.ts
evandroabukamel: Sure, I'll work on it.

evandroabukamel: That's the problem on using AsyncInMemory class on Master tests: the tests only work on the Master level, not on the socket level where the queue is. We have to implement a separate test for the socket using master with AsyncInMemory database.

evandroabukamel: @delucis , here are the news. I've made some practical tests today to test my implementation of PQueue for onUpdate messages with InMemory database on my game. It didn't work well, some parts of the state where overwritten and the clients became out of sync. Although, testing it with FlatFile worked perfectly which made me think FlatFile works better for the ignoreStaleStateID options because of that concurrency management you mentioned. The queue system was almost irrelevant, I need to test more because without the queue I saw just a few problems.

evandroabukamel: I've made a few more tests with PQueue and FlatFile and I had no problems. After that I put a 300 milliseconds interval to PQueue for Sync-typed storages, like InMemory, and it worked fine on my app. Also, I've found this article where it shows an example of testing SocketIO with Jest.

> https://medium.com/@tozwierz/testing-socket-io-with-jest-on-backend-node-js-f71f7ec7010f

I'll try it tomorrow, that's the only way I see for testing the update queuing on socket level.

delucis: > It didn't work well, some parts of the state where overwritten and the clients became out of sync.

That’s good to know — looks like we must be missing something important! I don’t have much time in the next few days to look into this, but let me know how you get on with the tests.

evandroabukamel: Thanks, @delucis , no rush. I'm facing some difficulties with the tests, there're some things I don't know how they work. Here is the test I wrote. First, I tried with just the app server but I had a problem where ctx was always undefined. Then I tried to use the API with the lobby client to create a match and authenticate the players but I got that `fetch`, from lobby client, is not defined.

Do you know how can I create and start a match without using lobby API but using the SocketIO server and client?

```javascript
describe('simultaneous moves', () => {
  const game = {
    name: 'test',
    setup: () => {
      const G = {
        players: {
          '0': {
            cards: ['card3'],
          },
          '1': {
            cards: [],
          },
        },
        cards: ['card0', 'card1', 'card2'],
        discardedCards: [],
      };
      return G;
    },
    playerView: PlayerView.STRIP_SECRETS,
    turn: {
      activePlayers: { currentPlayer: { stage: 'A' } },
      stages: {
        A: {
          moves: {
            A: (G, ctx: Ctx) => {
              const card = G.players[ctx.playerID].cards.shift();
              G.discardedCards.push(card);
            },
            B: {
              move: (G, ctx: Ctx) => {
                const card = G.cards.pop();
                G.players[ctx.playerID].cards.push(card);
              },
              ignoreStaleStateID: true,
            },
          },
        },
      },
    },
  };
  let server: ReturnType<typeof Server> | null;
  let runningServer: { appServer: KoaServer; apiServer?: KoaServer } | null;
  let clientConfig;
  let httpAppAddr;
  let client0: _ClientImpl<any>;
  let client1: _ClientImpl<any>;
  let lobbyClient: LobbyClient;

  beforeEach(async () => {
    server = Server({ games: [game] });
    runningServer = await server.run(undefined);
    httpAppAddr = runningServer.appServer.address();

    // lobbyClient = new LobbyClient({ server: `http://[${httpAppAddr.address}]:${httpAppAddr.port}` });

    clientConfig = {
      game: game,
      multiplayer: SocketIOClient({ server: `http://[${httpAppAddr.address}]:${httpAppAddr.port}` }),
      numPlayers: 2,
    };

    // const match = await lobbyClient.createMatch('test', { numPlayers: 2 });

    /*const p0Auth = await lobbyClient.joinMatch('test', match.matchID, {
      playerID: '0',
      playerName: 'Player 0',
    });
    const p1Auth = await lobbyClient.joinMatch('test', match.matchID, {
      playerID: '1',
      playerName: 'Player 1',
    });*/

    client0 = Client({
      ...clientConfig,
      playerID: '0',
      // credentials: p0Auth.playerCredentials,
    });
    client1 = Client({
      ...clientConfig,
      playerID: '1',
      credentials: p1Auth.playerCredentials,
    });

    client0.start();
    client1.start();

    client0.updateMatchID(match.matchID);
    client1.updateMatchID(match.matchID);

    // const state = { G: {}, ctx: { phase: '' }, plugins: {} };
    // const filteredMetadata = [];
    // client0.store.dispatch(sync({ state, filteredMetadata } as SyncInfo));
    // client1.store.dispatch(sync({ state, filteredMetadata } as SyncInfo));

    const state = { restore: true };
    const log = ['0', '1'];
    const action = sync(({ state, log } as unknown) as SyncInfo);
    client0.store.dispatch(action);
    client1.store.dispatch(action);
  });

  afterEach(() => {
    if (server && runningServer) {
      const { apiServer, appServer } = runningServer;
      server.kill({ apiServer, appServer });
    }

    if (client0 && client0.stop) client0.stop();
    if (client1 && client1.stop) client1.stop();
  });

  test('two clients playing', async () => {
    expect(server).not.toBeUndefined();
    expect(runningServer.appServer).not.toBeUndefined();
    expect(runningServer.apiServer).toBeUndefined();

    // test: simultaneous moves
    client0.getInitialState();
    client1.getInitialState();

    client0.moves.A();
    client0.events.setActivePlayers({ all: 'A' });

    client1.moves.B();
    client0.moves.B();

    const stateP0 = client0.getState();
    const stateP1 = client1.getState();

    expect(stateP0.G).toMatchObject({
      players: {
        '0': {
          cards: ['card1'],
        },
      },
      cards: ['card0'],
      discardedCards: ['card3'],
    });
    expect(stateP1.G).toMatchObject({
      players: {
        '1': {
          cards: ['card2'],
        },
      },
      cards: ['card0'],
      discardedCards: ['card3'],
    });
  });
});

```

delucis: I can’t see here where the storage is instantiated, but could you use it directly to create state, e.g.:

```js
const db = new Storage();

const server = Server({ db });

db.createMatch('id', {
  initialState: InitializeGame({ /* ... */ }),
  metadata: createMetadata({ /* ... */ }),
});
```

`InitializeGame` is exported from [`core/initialize`](https://github.com/boardgameio/boardgame.io/blob/master/src/core/initialize.ts#L17) and `createMetadata` from [`server/util`](https://github.com/boardgameio/boardgame.io/blob/master/src/server/util.ts#L6) if you decide to use them, but you could also use manual values.

Otherwise, when `client.start()` is called, connecting the transport, a `SYNC` action should be dispatched to the master, which will create the match automatically.

evandroabukamel: Thanks, @delucis. About sending a SYNC action, how can I do this from the clients? Am I missing something here?

delucis: It should happen automatically. Basically when you call `client.start()`, it calls `transport.connect()`. In the transport implementations, a sync action is dispatched as part of the connect method.

evandroabukamel: That's the problem, the `socket.on('sync', () => { })` is not being called on the client. Looks like the client is not connected to the server or the server didn't respond.

evandroabukamel: From the logs, it looks like the server starts after the client move. How?  :confused: 

```
simultaneous moves › two clients playing

    TypeError: Cannot read property 'phase' of undefined

      167 | 
      168 |   function GetPhase(ctx: { phase: string }): PhaseConfig {
    > 169 |     return ctx.phase ? phaseMap[ctx.phase] : phaseMap[''];
          |                ^
      170 |   }
      171 | 
      172 |   function OnMove(s) {

      at GetPhase (src/core/flow.ts:169:16)
      at Object.GetMove (src/core/flow.ts:632:18)
      at src/core/reducer.ts:190:38
      at dispatch (node_modules/redux/lib/redux.js:218:22)
      at src/client/client.ts:226:22
      at src/client/client.ts:280:22
      at Object.dispatch (src/client/client.ts:295:22)
      at Object.dispatchers.<computed> [as A] (src/client/client.ts:85:13)
      at Object.<anonymous> (src/server/transport/socket-multiplayer.test.ts:143:19)

  console.log src/core/logger.ts:14
    INFO: App serving on 34173...
```

evandroabukamel: Sad news. I've tried a lot of things. That approach is not going to work under a test environment, [see more here](https://stackoverflow.com/questions/15509231/unit-testing-node-js-and-websockets-socket-io). I will try again mocking the server SocketIO and send fake moves.

delucis: Thanks for your persistence! Hopefully I’ll have more time soon to actually read through the changes. Mocking sounds like a good idea.

evandroabukamel: > (By the way, noticed the “Skipping failing test just to push he code.” commit. You can push without running tests by doing `git push --no-verify` if that helps in the future.)

Thanks, I forgot about that.  :smile:

evandroabukamel: Now I see: Master methods onSync and onUpdate are mocked.

evandroabukamel: @delucis Finally!  :smile:  I've managed to write the unit tests for the actions queue on socket level. I've also took liberty to add playerID and action type on Master's error messages, it help me to follow the actions being sent to the server and possible errors.

delucis: Nice! 😄

I’ll try to get a review in soon.

src/server/transport/socketio-simultaneous.test.ts
```diff
@@ -0,0 +1,280 @@
...
+                const card = G.cards.pop();
+                G.players[ctx.playerID].cards.push(card);
+              },
+            },
+          },
+        },
+      },
+    },
+  };
+  let db = new InMemory();
```
delucis, line 147: I might be missing something here, but shouldn’t we use asynchronous storage to test this? Or run tests with both a sync and async storage implementation?

delucis (commented):
Hi @evandroabukamel, I have a couple of questions regarding the implementation, before I look into some more of the details. Thanks!

src/server/transport/socketio.ts
```diff
@@ -126,7 +130,12 @@ export class SocketIO {
             TransportAPI(matchID, socket, this.clientInfo, this.roomInfo),
             this.auth
           );
-          await master.onUpdate(action, stateID, matchID, playerID);
+
+          const queueInterval = IsSynchronous(app.context.db) ? 300 : 0;
```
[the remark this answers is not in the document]
evandroabukamel: I did that because the problem continued to occur with InMemory, even using queue. That interval solved it. Maybe a smaller interval should work, have to test. I'll run more tests for this point.

src/server/transport/socketio-simultaneous.test.ts
evandroabukamel: Yeap, I can do that.

src/server/transport/socketio.ts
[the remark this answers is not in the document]
delucis: > This is complicated because the unit tests will not always reflect the real world.

Yes, that’s my fear — anything interval-based is dependent on the real world staying within that interval, which it might not. It also suggests there’s a more fundamental bug in the queueing mechanism. If the queue is not waiting correctly to process each update with synchronous storage (which shouldn’t happen concurrently anyway 😕?), adding an interval “fixes” it kind of, but I’m guessing only where the update processes fast enough, so we still have a race condition on our hands if the master ever slows down enough.

We probably need to understand why the `InMemory` storage breaks the queuing system, rather than patching it with an interval.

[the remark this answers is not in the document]
evandroabukamel: I understand your point. I believe that writing in memory happens too fast and it does not have a "await" functionality, so an action's result overwrite another action result because they seem to happen at the same time in memory. Right now, I think that adding logs to InMemory will help to investigate that.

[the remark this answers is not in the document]
delucis: Yes, some logging to check which steps are happening before/after each other without the interval could be a good idea. Some strange order things can happen with mixed async and sync stuff like:

```js
const f1 = async () => {
  await somethingThatTakesTime();
  console.log('f1');
};

const f2 = () => {
  f1(); // not awaited
  console.log('f2');
};

f2();
// output:
// 'f2'
// 'f1'
```

In a simple example like this it might be obvious, but it’s easy to get tangled up in this with a more complex codebase like this one.

evandroabukamel:
I'm trying to install this branch on my project and deploy it on Heroku, but I got "run-s not found". Do someone know how can I solve this? As you can see from my last commits I tried a lot o things but nothing solved it. @delucis or @nicolodavis ? Thnaks.

evandroabukamel:
> I'm trying to install this branch on my project and deploy it on Heroku, but I got "run-s not found". Do someone know how can I solve this? As you can see from my last commits I tried a lot o things but nothing solved it. @delucis or @nicolodavis ? Thanks.

I managed to upload a user-scoped package to NPM and it seem to work fine.

src/server/transport/socketio.ts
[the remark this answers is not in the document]
evandroabukamel: Even if you await for 0 seconds on f1, it will be printed out after f2.  :confused:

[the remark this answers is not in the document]
delucis: Yup, exactly. That’s expected, but does need careful management.

src/server/transport/socketio-simultaneous.test.ts
evandroabukamel: Hi, @delucis . I used the last days to fix some bugs in my app and today I made some tests with thoses fixes. Using the PQueue without interval and InMemory storage, I played 4 matches in a row and no errors occurred.  :smile: 
That concludes that the errors I've reported were on the client, not in the engine.

evandroabukamel: So, today I'm gonna push my project to the cloud and test it online. After that I will clean up the code removing unnecessary stuff and it will be ready to mere. Would you like me to test something else?

evandroabukamel: Code pushed. Should I write a documentation for it, too?

delucis (changes requested):
Hi @evandroabukamel — sorry for the wait on going through these and thanks for your patience. Glad to hear the tests with your app went smoothly. Hopefully most of these changes are small details and we can get this merged soon :+1:

src/server/transport/socketio-simultaneous.test.ts
```diff
@@ -0,0 +1,408 @@
...
+      })(),
+    ]);
+
+    expect(spyDeleteMatchQueue).toHaveBeenCalledWith('matchID');
+
+    db.wipe('matchID');
+  });
+
+  test('two clients playing using async storage', async () => {
+    let db = new FlatFile({ dir: './tmp', logging: false });
```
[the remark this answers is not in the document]
evandroabukamel: Yeah, I'm gonna do that.

packages/server.ts
```diff
@@ -7,7 +7,7 @@
  */
 
 import { Server } from '../src/server';
-import { FlatFile } from '../src/server/db';
+import { FlatFile, InMemory } from '../src/server/db';
 import { SocketIO } from '../src/server/transport/socketio';
 
-export { Server, FlatFile, SocketIO };
+export { Server, FlatFile, InMemory, SocketIO };
```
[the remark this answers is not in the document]
evandroabukamel: Because I need it for my real-world tests. InMemory class wasn't exported to be used by the Server just like FlatFile already is.

src/master/master.test.ts
```diff
@@ -286,7 +436,7 @@ describe('update', () => {
     const gameOverArg = 'gameOverArg';
     const event = ActionCreators.gameEvent('endGame', gameOverArg);
     await masterWithMetadata.onUpdate(event, 0, id, '0');
-    const { metadata } = db.fetch(id, { metadata: true });
+    const { metadata } = await db.fetch(id, { metadata: true });
```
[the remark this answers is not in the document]
evandroabukamel: Some of them were using Sync Storage. If you believe it's necessary, I can do that.

packages/server.ts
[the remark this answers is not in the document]
delucis: Ah, you mean to extend like `InMemoryAsync` in your project tests? Could we remove this when we’re ready to merge then?

The server uses `InMemory` by default if no other database is passed though, and I guess most people don’t need to extend `InMemory` in normal situations. Or do you need `InMemory` in the future too?

src/master/master.test.ts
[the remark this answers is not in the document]
delucis: [L432](https://github.com/boardgameio/boardgame.io/blob/fc8cda590009852af1e7bed3a6ab5d7113cdfdeb/src/master/master.test.ts#L432), [L453](https://github.com/boardgameio/boardgame.io/blob/fc8cda590009852af1e7bed3a6ab5d7113cdfdeb/src/master/master.test.ts#L453) and [L490](https://github.com/boardgameio/boardgame.io/blob/fc8cda590009852af1e7bed3a6ab5d7113cdfdeb/src/master/master.test.ts#L490) all seem to be using `InMemoryAsync` as far as I can tell?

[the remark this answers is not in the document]
evandroabukamel: Yes, this test is using InMemory async and those lines you mentioned are being awaited.

packages/server.ts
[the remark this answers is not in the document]
evandroabukamel: Also, I used InMemory inside my game server but it wasn't exposed inside boardgame.io/server.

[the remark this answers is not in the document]
evandroabukamel: Fine, I'm gonna undo that.

src/server/transport/socketio.ts
```diff
@@ -176,4 +187,27 @@ export class SocketIO {
...
+  }
+
+  /**
+   * Delete a PQueue for a given matchID.
+   * @param matchID
+   */
+  deleteMatchQueue(matchID: string): void {
+    if (this.perMatchQueue.has(matchID)) {
+      this.perMatchQueue.delete(matchID);
+    }
```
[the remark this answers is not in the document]
evandroabukamel: Thanks for the review, @delucis . Your suggestions make sense and I've fixed all of them.

delucis (approved):

delucis:
Merged! Thanks for your persistence @evandroabukamel 🎉

PR #709, opened 2020-05-21
0.40.0

delucis:
This PR will merge the 0.40.0 development branch. For now it is mainly intended to provide a space to clearly document upcoming breaking changes as they are made.

# 0.40.0 Breaking Changes

- Matches in the Lobby API (Issue #703, PRs #666, #704)
- Match IDs & metadata in clients (Issue #763, PR #765)
- Passing a custom uuid method to the Lobby API (PR #698)

## Matches in the Lobby API (Issue #703, PRs #666, #704)

Until now, the Lobby API has been slightly inconsistent in how it names an instance of a game, sometimes describing this as a “game” sometimes as a “room”. In release 0.40.0, this is standardised to “match”, making it clearer that you can play a **game**, like Chess or Tic-Tac-Toe, and the specific instance of a game is a **match**.

### Changes

If you are using the provided React Lobby component, you shouldn’t see any breaking changes, but if you are consuming the Lobby REST API directly, you need to be aware of the following changes.

1. In JSON returned by the `games/{name}` endpoint:

    - `rooms` → `matches`

    - `roomID` → `matchID`

    ```diff
    {
    - "rooms": [
    + "matches": [
        {
    -     "roomID": "random-string",
    +     "matchID": "random-string",
          "players": {}        
        }
      ]
    }
    ```

2. In JSON returned by the `games/{name}/create` endpoint:

    - `gameID` → `matchID`

    ```diff
    {
    - "gameID": "random-string"
    + "matchID": "random-string"
    }
    ```

3. In JSON returned by the `games/{name}/{id}` endpoint:

    - `roomID` → `matchID`

    ```diff
    {
    - "roomID": "random-string",
    + "matchID": "random-string",
      "players": {}
    }
    ```

4. In JSON returned by the `games/{name}/{id}/playAgain` endpoint:

    - `nextRoomID` → `nextMatchID`

    ```diff
    {
    - "nextRoomID": "random-string"
    + "nextMatchID": "random-string"
    }
    ```

5. If writing Typescript and using the `Server` types namespace: `Server.GameMetadata` → `Server.MatchData`

### Migration Help

- If you are writing Typescript, you can import types for the data returned by each of the API endpoints so you can know what shape to expect.

  ```ts
  import type { LobbyAPI } from 'boardgame.io';
  
  async function fetchMatches(gameName: string) {
    const response = await fetch(`http://localhost:8000/games/${gameName}`);
    const matches: LobbyAPI.MatchList = await response.json();
  }
  ```

[...]

PR #839, opened 2020-10-28
feat: Track players connection status (#834)

### LINKED ISSUE #834: Keep track of players' connection status
It would be useful to keep track of all players' connections status and broadcast that info to other players. This would allow displaying connection status next to player info and let other players know that they are currently unavailable due to e.g. network issues.

I think the best place to store that info is `matchData` 
```
[
  { id: 0, name: 'Alice', isConnected: true },
  { id: 1, name: 'Bob', isConnected: false }
]
```

It could use server-side SocketIO connect/disconnect events to modify `matchData` and broadcast that info independently of game state.

If you think it's a nice-to-have feature in the framework I'd be happy to work on it and prepare a PR.

pastDexter:
Closes #834 

- Server now keeps track of all players connection status and stores this information in `matchData` alongside `id` and `name`
- Every time player connects or disconnects from the SocketIO server, `onConnectionChange` method is called on `Master` to update players metadata in `DB` and broadcast that info to all the players in the match as `matchData` message

#### Checklist

### DIFF
--- a/docs/documentation/api/Client.md
+++ b/docs/documentation/api/Client.md
@@ -108,7 +108,7 @@ The following properties are available on a client instance:
   ```js
   [
     { id: 0, name: 'Alice' },
-    { id: 1, name: 'Bob' }
+    { id: 1, name: 'Bob', isConnected: true }
   ]
   ```
 
@@ -322,7 +322,7 @@ following as `props`:
     ```js
     [
       { id: 0, name: 'Alice' },
-      { id: 1, name: 'Bob' }
+      { id: 1, name: 'Bob', isConnected: true }
     ]
     ```
 
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -344,6 +344,7 @@ export class _ClientImpl<G extends any = any> {
 
     this.transport.subscribeMatchData(metadata => {
       this.matchData = metadata;
+      this.notifySubscribers();
     });
   }
 
--- a/src/client/transport/socketio.test.js
+++ b/src/client/transport/socketio.test.js
@@ -135,6 +135,16 @@ describe('multiplayer', () => {
     expect(store.getState()).toMatchObject(restored);
   });
 
+  test('receive matchData', () => {
+    let receivedMatchData;
+    m.subscribeMatchData(data => (receivedMatchData = data));
+    const matchData = [{ id: '0', name: 'Alice' }];
+    mockSocket.receive('matchData', 'unknown matchID', matchData);
+    expect(receivedMatchData).toBe(undefined);
+    mockSocket.receive('matchData', 'default', matchData);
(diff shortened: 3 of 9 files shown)


delucis (approved):
Thanks @pastDexter! Added a small refactor to remove some duplicate code, but this looks great 👍

evandroabukamel:
Hey guys, I have a about this: are we able to read that status from inside a move? @delucis , do you know that?

delucis:
@evandroabukamel No, this is only stored in match metadata, which is used by the lobby and can be requested from the lobby server by the client. What’s your use case for using it in a move?

evandroabukamel:
Hi, @delucis . In my game there's a case when currentPlayer may disconnect from the game during their turn and not return back, so their data will remain on the game state. I need to pick a random player who is still connected and begin a new turn with him/her as currentPlayer.

delucis:
How are you planning on triggering the move? The only secure way I can think of is to have a move, which you call from the server on behalf of the disconnected current player. On the server you have access to the full metadata, so you could get the current player’s credentials, pick a next player and dispatch the move with the selected playerID.

You would also have to see how reliable the connection status is (and how to monitor it) — players can probably briefly disconnect then reconnect if they have poor internet (on mobile for example), so you might see disconnections even if a player is still trying to play.

evandroabukamel:
Well, I didn't think about that before. Now it came to me the idea to have a skipTurn move, similar to "Skip voting" from Among Us. Inside that move, if half of players vote to skip within an interval time, the game should pick a random player, but I must assure this player is connected. Maybe, I could choose between those who voted, it's a funny feature.  :smile:

evandroabukamel:
Hi, @delucis . I almost forgot about that. Please, could you clarify that? I don't see how I can achieve it:

> On the server you have access to the full metadata, so you could get the current player’s credentials, pick a next player and dispatch the move with the selected playerID.

Thanks.

delucis:
@evandroabukamel Ah, I assumed you already had some plan to run some custom server-side code for the timeout/next player picking. I meant that on the server you could use the database to fetch the metadata, but you would still need to customise the API or master interface to trigger this somehow (e.g. using a timer, which boardgame.io doesn’t currently have implemented) and I don’t have any immediate ideas about how to best do that.

Here’s an example from the #709 changelog of adding a custom endpoint to the lobby API adapted to demonstrate sharing access to the database:

```js
import { Server, FlatFile } from 'boardgame.io/server';

const db = FlatFile({/* ... */});
const server = Server({ db, /* ... */ });

server.router.get('/custom-endpoint', async (ctx, next) => {
  const { metadata } = await db.fetch('matchID', { metadata: true });
  // use metadata
  ctx.body = 'Custom Response'
});

server.run(8000);
```

PR #842, opened 2020-11-01
ci(Github): Add npm publish workflow

delucis:
This PR adds a GitHub workflow file that will run for new git tags on the master branch that have the form `v0.x.x`.

The workflow will:

  1. Run unit tests as a safety check.

  2. Try to publish to npm, using [this GitHub Action](https://github.com/marketplace/actions/npm-publish?version=v1).

If desirable, we could set `dry-run: true` on the publish step to test out the command before running it for real.

This workflow is only responsible for publishing to npm. Tagging a new release, updating the changelog etc. must still be done locally.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -25,17 +25,17 @@ jobs:
           npm run test
 
       - name: Publish to NPM
-        id: npm_publish
+        id: publish
         uses: JS-DevTools/npm-publish@v1
         with:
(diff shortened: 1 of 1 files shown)


nicolodavis:
Thanks @delucis! I've added the `NPM_TOKEN` secret to the repo.

nicolodavis (approved):

delucis:
Great! Just tried this out and it appears to have published correctly as expected. 🎉

PR #849, opened 2020-11-25
fix(set-stage): setStage event for Stage.NULL makes player active. (closes #848)

### LINKED ISSUE #848: Strange behaviour for setStage and setActivePlayers for Stage.NULL
Hey guys, what's up? I have a question about using the Stage.NULL. In the docs it's said that players can become active but in no stage if we set them to Stage.NULL using `setActivePlayers`.

What I'm trying to achieve is to make all players to use a specific move from no stage but when they make some moves from specific stages they must exit that stage but still become active. So, I'm trying to keep all other players in the same state they are when players calls a move that should put themselves to Stage.NULL. Now, everybody receive a "player not active" error. So I try a few things:

**1. Using `setStage(Stage.NULL)` at the end of a move:** by doing this, the player was removed from `activePlayers` property, thus becoming inactive.

**2. Using `setActivePlayers([ctx.playerID])` at the end of a move:** by doing this, only that player becomes active in Stage.NULL and all other players are removed from `activePlayers` property.

So, am I doing something wrong or is there a bug in one of those functions?
What you suggest me to do? I was thinking of merge the current `ctx.activePlayers` with `setActivePlayers` parameter, but I think this is very risky. Don't know.

Thanks.

evandroabukamel:
#### Checklist

Checking if new stage is different than `undefined` because `Stage.NULL` is a null value and without this fix it falls outside some conditions that make a player active.


src/core/turn-order.test.ts
```diff
@@ -451,6 +451,53 @@ describe('setActivePlayers', () => {
     expect(state.ctx.activePlayers).toBeNull();
   });
 
+  test('setActivePlayers and setStage', () => {
```
delucis, line 454: Let’s use a slightly more descriptive test name:

```suggestion
  test('set stages to Stage.NULL', () => {
```

src/core/turn-order.test.ts
```diff
@@ -451,6 +451,53 @@ describe('setActivePlayers', () => {
...
+          },
+        },
+      },
+    };
+    const reducer = CreateGameReducer({ game });
+    let state = InitializeGame({ game, numPlayers: 3 });
+
+    expect(state.ctx.currentPlayer).toBe('0');
+    expect(Object.keys(state.ctx.activePlayers)).toEqual(['0', '1', '2']);
+    expect(state.ctx.activePlayers['0']).toEqual('start');
```
delucis, line 490: Just for safety, it would be good to also assert the other players’ stages here:

```suggestion
    expect(state.ctx.activePlayers['0']).toEqual('start');
    expect(state.ctx.activePlayers['1']).toEqual(Stage.NULL);
    expect(state.ctx.activePlayers['2']).toEqual(Stage.NULL);
```

delucis (changes requested):
Thanks @evandroabukamel. Added a couple of comments to the test, but the code itself looks good 👍

delucis (approved):

PR #856, opened 2020-12-03
Fix: endStage event crashes debug log panel.

### LINKED ISSUE #855: Debug log panel seems to be broken when changing stage.
## Description
Client crashes when opening the log panel after a stage change (or changing stages while the log panel is open). 
I'm using the React client.

Error in console:
```
ctx[6].join is not a function
```
This error occurs for `setStage("myStage")` or `endStage()` with `next: "myStage"` set.

I blindly tried `setStage(["myStage"])` (i.e. the next stage name in a list). 
This removes the crash and the log appears in the panel, but the new stage is not set. The current player is just removed from the `activePlayers`.
## What I figured out
* Given the context around `ctx[6]` in the webpacked file, I think it corresponds to `action.payload.args ?? []` in `src/client/debug/log/LogEvent.svelte` (line 13 on master as of today).
* `join` will not be available if `args` is a string
* `args` is a string in `endStage` events. Example:
```json
{"action":{"type":"GAME_EVENT","payload":{"type":"endStage","args":"chooseCastingRunes"}},"_stateID":1,"turn":2,"phase":"fight"}
```

### Handle string args in LogEvent
To avoid a breaking changes, I would suggest to handle the string case in `LogEvent.svelte`. Something like:
```javascript
const renderedArgs = typeof args === 'string' ? args : (args || []).join(','); // line 13
// ...
 <div>{actionType}({renderedArgs})</div> // line 135
```

Drestin:
Fixes #855 

As discussed in the issue, handle the case in the debug panel where `args` is a string.
I manually tested it, now `endStage` events appear in log without crash.

#### Checklist

### DIFF
--- a/src/client/debug/log/LogEvent.svelte
+++ b/src/client/debug/log/LogEvent.svelte
@@ -10,7 +10,8 @@
 
   const dispatch = createEventDispatcher();
 
-  const args = action.payload.args || [];
+  const args = action.payload.args;
+  const renderedArgs = typeof args === 'string' ? args : (args || []).join(',');
   const playerID = action.payload.playerID;
   let actionType; 
   switch (action.type) {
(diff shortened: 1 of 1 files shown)


delucis (approved):
Thanks @Drestin!

PR #871, opened 2020-12-15
Chat messages

TomasVidhall:
So I read issue #460 about out of the box chat support and this is WIP for a very basic, non-persistent, chat messages passed between socket-io clients. If you guys think this is a good idea I will polish it and add tests etc. For now I just did something to see if my theory would work in practice. I tested it using the chess example and added some basic UI for it.

@delucis @nicolodavis 

Screenshot 2020-12-15 at 17 34 00

#### Checklist

### DIFF
--- a/examples/react-web/src/chess/board.js
+++ b/examples/react-web/src/chess/board.js
@@ -11,6 +11,7 @@ import PropTypes from 'prop-types';
 import { Chess } from 'chess.js';
 import { Checkerboard, cartesianToAlgebraic } from './checkerboard';
 import { Token } from './token';
+import Chat from './chat';
 import Bishop from './pieces/bishop';
 import King from './pieces/king';
 import Knight from './pieces/knight';
@@ -31,6 +32,8 @@ class Board extends React.Component {
     isActive: PropTypes.bool,
     isMultiplayer: PropTypes.bool,
     isConnected: PropTypes.bool,
+    sendChatMessage: PropTypes.func,
+    chatMessages: PropTypes.array,
   };
 
   constructor(props) {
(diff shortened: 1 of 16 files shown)


nicolodavis:
I do like the idea of an ephemeral chat that utilizes the boardgame.io transport layer. Nice work @TomasVidhall.

Can you add some unit tests for the code that you've added?

TomasVidhall:
Hey @nicolodavis. Added some tests for the code now, not sure if I missed anything important since I don't fully grasp the codebase, so please point out if I did. 

I really wanted to add a test that actually used the client-server connection of SocketIO, but couldn't make it work. If you see a way to achieve that please tell 😄

examples/react-web/src/chess/board.js
```diff
@@ -64,6 +67,12 @@ class Board extends React.Component {
 
         {this._getStatus()}
         {disconnected}
+        {this.props.sendChatMessage && this.props.chatMessages && (
+          <Chat
+            onSend={this.props.sendChatMessage}
```
flamecoals, line 72: Maybe for a future PR: one use case that we might be missing here is the use case that the game itself adds some messages (not by a player)... It would be good if that was available close to the ctx or the G in the game rules, and that if it is a multiplayer game, the server itself could send the message on chat.

src/master/master.ts
```diff
@@ -101,6 +102,10 @@ type TransportData =
   | {
       type: 'matchData';
       args: [string, FilteredMetadata];
+    }
+  | {
+      type: 'chat-message';
```
nicolodavis, line 107: Let's change the type to just 'chat'.

nicolodavis (approved):
> I really wanted to add a test that actually used the client-server connection of SocketIO, but couldn't make it work.

That's fine.  Thanks for the tests that you added.

src/master/master.ts
TomasVidhall: Done!

nicolodavis (approved):

PR #878, opened 2020-12-26
refactor: Fix and update linters

delucis:
I realised that we aren’t actually linting any of our Typescript code, which is now almost everything!

This PR updates the various ESLint plugins and configs and replaces `babel-eslint` with `@typescript-eslint`. Because updating the various presets introduced a bunch of new rules, I’m currently going through them to strike a balance between matching our current code and enabling rules that do seem like a good idea even if it does require refactoring some existing code.


delucis:
@nicolodavis This should be ready. The only warnings now displayed when linting come from the [`jest/expect-expect`][jee] rule for tests which don’t actually assert anything. I don’t feel we should disable this rule, but I think a separate PR would be more appropriate for updating these tests where possible.

[jee]: https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/expect-expect.md

nicolodavis (approved):
Nice work @delucis!

PR #879, opened 2020-12-29
Add authentication to chat messages

### LINKED ISSUE #876: Authenticate chat messages
I didn’t have time to look at #871 before it was merged, but I see now it didn’t include authentication like we now have for all the other game master methods (partly because #853 was only merged after @TomasVidhall started work).

Should we use player credentials for authentication? This would only allow players rather than spectators to chat, but currently an unauthenticated client could post a message claiming to be an authenticated player.

TomasVidhall:
Closes #876 

Follow-up to #871 which implements the new auth logic implemented in #853 for chat messages. I think this code will make it possible for spectators to send chat messages (with no sender id though), but it will be impossible to impersonate another player in the chat.

Please feedback and improve @delucis @nicolodavis 

#### Checklist


src/master/master.ts
```diff
@@ -459,7 +459,30 @@ export class Master {
...
+  ): Promise<void | { error: string }> {
+    const key = matchID;
+    let metadata: Server.MatchData | undefined;
+
+    if (StorageAPI.isSynchronous(this.storageAPI)) {
+      ({ metadata } = this.storageAPI.fetch(key, { metadata: true }));
+    } else {
+      ({ metadata } = await this.storageAPI.fetch(key, { metadata: true }));
+    }
+    if (this.auth && chatMessage.sender !== undefined) {
```
delucis, line 475: @TomasVidhall I’m not 100% sure about allowing spectators to post messages without authentication, although this is an improvement over allowing them to claim to be a player. Is this a functionality you need? In general I would lean towards only permitting players to post, but am open to being more flexible if you definitely need it (or perhaps adding a server option to enable spectator chat).

src/client/transport/socketio.ts
```diff
@@ -97,7 +97,7 @@ export class SocketIOTransport extends Transport {
   }
 
   onChatMessage(matchID, chatMessage) {
-    this.socket.emit('chat', matchID, chatMessage);
+    this.socket.emit('chat', matchID, chatMessage, this.credentials);
```
delucis, line 100: In the other transport methods to help catch differences between server and client code, we use types from `Master` to type the arguments to `emit`. (For example, see [`onAction`](https://github.com/boardgameio/boardgame.io/blob/8f85fb0a4f2951dd66cc3d96485b3f28909b0af7/src/client/transport/socketio.ts#L89-L97).)

It would be good to add that here too:

```suggestion
    const args: Parameters<Master['onChatMessage']> = [
      matchID,
      chatMessage,
      this.credentials,
    ];
    this.socket.emit('chat', ...args);
```

Can we also add types to the method signature itself?

```js
   onChatMessage(matchID: string, chatMessage: ChatMessage) {
```

src/server/transport/socketio.ts
```diff
@@ -185,14 +185,14 @@ export class SocketIO {
             );
           }
         });
-        socket.on('chat', async (matchID, chatMessage) => {
+        socket.on('chat', async (matchID, chatMessage, credentials) => {
```
delucis, line 188: Similarly to in the client transport, let’s type this callback using types from `Master`.

```suggestion
        socket.on('chat', async (...args: Parameters<Master['onChatMessage']>) => {
          const [matchID, chatMessage, credentials] = args;
```

These types don’t change how anything behaves or compiles, but help keep things in sync if we ever refactor these interfaces.

src/client/transport/local.ts
```diff
@@ -161,8 +161,13 @@ export class LocalTransport extends Transport {
...
   }
 
+  /**
+   * Called when any player sends a chat message and the
+   * master broadcasts the update to other clients (including
+   * this one).
+   */
   onChatMessage(matchID, chatMessage) {
-    this.master.onChatMessage(matchID, chatMessage);
+    this.master.onChatMessage(matchID, chatMessage, this.credentials);
```
delucis, line 170: Same comment again here — would be good to type this just like the other transport methods.

delucis (changes requested):
Thanks @TomasVidhall and Happy New Year! This looks good in general. I have one question about leaving the chat open to spectators — let me know what you think.

The other comments recommend some more precise typing that will help with future refactoring. Updating these might require some changes in the tests to fix some TypeScript errors — if anything isn’t clear, I’ll be happy to help.

delucis:
Thanks for getting this rolling @TomasVidhall. I wanted to get this released, so I’ve merged this for now without spectator chat. Please feel free to open a PR adding that back in later if you need it!

Issue #788, opened 2020-08-29
state.deltalog is not iterable


popey456963:
If `endIf` is called prior to the first turn, e.g. with the following code:

```
endIf: (G, ctx) => {
  return 'gameover'
}
```

Then the following error is thrown:

```
(node:5152) UnhandledPromiseRejectionWarning: TypeError: state.deltalog is not iterable
    at EndPhase (C:\Users\Alexander\Documents\Github\robot-fight\node_modules\boardgame.io\dist\cjs\server.js:1480:36)
    at EndGame (C:\Users\Alexander\Documents\Github\robot-fight\node_modules\boardgame.io\dist\cjs\server.js:1443:17)
    at Process (C:\Users\Alexander\Documents\Github\robot-fight\node_modules\boardgame.io\dist\cjs\server.js:1266:21)
    at Object.init (C:\Users\Alexander\Documents\Github\robot-fight\node_modules\boardgame.io\dist\cjs\server.js:1762:20)
    at InitializeGame (C:\Users\Alexander\Documents\Github\robot-fight\node_modules\boardgame.io\dist\cjs\server.js:1901:25)
    at Master.onSync (C:\Users\Alexander\Documents\Github\robot-fight\node_modules\boardgame.io\dist\cjs\server.js:2895:36)
    at Socket.<anonymous> (C:\Users\Alexander\Documents\Github\robot-fight\node_modules\boardgame.io\dist\cjs\server.js:3020:34)
    at Socket.emit (events.js:315:20)
    at Socket.EventEmitter.emit (domain.js:485:12)
    at C:\Users\Alexander\Documents\Github\robot-fight\node_modules\socket.io\lib\socket.js:528:12
```

popey456963:
I realise it's rare for a game to end prior to the first turn, but it's useful for debugging game end state.

delucis:
Closed in #888

PR #888, opened 2021-01-11
Prevent TypeError: state.deltalog is not iterable

mdzw:
#788 "UnhandledPromiseRejectionWarning: TypeError: state.deltalog is not iterable"

Everywhere else that deltalog is updated, it uses this (state.deltalog || []) pattern.

#### Checklist

### DIFF
--- a/src/core/flow.test.ts
+++ b/src/core/flow.test.ts
@@ -713,6 +713,20 @@ describe('endIf', () => {
         .payload.type
     ).toBe('endPhase');
   });
+
+  test('during game initialization with phases', () => {
+    const flow = Flow({
+      phases: {
+        A: {
+          start: true,
+        },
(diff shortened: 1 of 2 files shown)


delucis (commented):
Thanks @mdzw! Do you think we could add a test to `/src/core/flow.test.ts` to cover the case where this fallback is needed? (If I understood the original issue correctly, a scenario with a default `turn.endIf` that returns a truthy value should do it.)

delucis:
@mdzw @popey456963 Just tried writing a test case for this, but couldn’t figure out a way to trigger the original error. Would you be able to share an example game object that triggers this bug?

delucis (approved):

delucis:
Thanks @mdzw!

Issue #608, opened 2020-04-04
onBegin and onEnd for Stages


matthiasseemoo:
I would like to execute code when entering a new stage and when leaving it. To this end, I would like to use onBegin and onEnd in stages. However, those keywords so not seem to have any effect in the stages sections. Would it be possible to add onBegin and onEnd to stages?

Example:
`const MyGame = {
  turn: {
    onBegin: BeginTurn,

    stages: {
      discover: {
        onBegin: BeginDiscoverStage,
        onEnd: EndDiscoverStage,
        moves: { DrawCard, GetCoins },
      },
    },
  },
};`

delucis:
Please see #561. The current recommendation is to call your `BeginDiscoverStage` wherever you are currently doing `setStage('discover')` (or `setActivePlayers`) and call your `EndDiscoverStage` wherever you are currently calling `endStage` if possible.

matthiasseemoo:
Thank you for your answer, next time, I should also search in the closed issues.
I understand that under the circumstances described in #561 onBegin and onEnd can lead to bugs. However, in my case I have multiple stages which also define a "next" element to continue to the next stage when the endStage event is called. Rewriting the endStage function to manually call setStage seems like much more code than simply having the ability to hook onBegin and onEnd of a stage. In my game architecture, I only have a single active player, so there will be no bugs related to multiple players being active at once. At least for those simple use cases, it would be nice to have the two hooks available.

delucis:
I agree, that sounds like a good use case for these hooks. Maybe it’s something that can be added in the future.

senritsu:
Just to mention it, as discussed in #985 it may be also useful for various things to add `endIf` and the hypothetical-and-not-yet-discussed `shouldEnd` to stages as well. This would allow functionality similar to current `moveLimit` (or in the PR `minMoves`/`maxMoves`) for all parts of the flow, without being restricted to just use the number of moves players have made.

PR #891, opened 2021-01-17
feat: Change move and hook signature

### LINKED ISSUE #662: move signature change

delucis:
### 📣 Try the pre-release today 📣

You can start working with the new hook signature by installing the alpha release channel from npm:

```bash
npm i boardgame.io@alpha
```

If you run into any issues, let us know!

---

Closes #662.

This PR implements the change from `(G, ctx)` to `({ G, ctx, playerID, ...plugins })`.

Game code that currently looks like this:

```js
const move = (G, ctx, arg) => {
  const score = arg * ctx.random.D6();
  G.players[ctx.playerID] = score;
  if (score < 12) ctx.events.endTurn();
};
```

Would become:

```js
const move = ({ G, random, playerID, events }, arg) => {
  const score = arg * random.D6();
  G.players[playerID] = score;
  if (score < 12) events.endTurn();
};
```

This applies across moves and hooks.

### To-do

The following parts of game code haven’t been changed yet:

1. [x] `playerView` is currently still `(G, ctx, playerID) => G`. Should this become `({ G, ctx, playerID }) => G`?

2. I haven’t updated the bot interface at this point. Should the following be changed for consistency?
    - `enumerate(G: any, ctx: Ctx, playerID: PlayerID)`
    
    - Various other methods in the MCTS bot implementation

3. [x] The `undoable` method of the long-form move syntax, which can be `(G, ctx) => boolean`. Should this become `({ G, ctx }) => boolean`?

### Notes

- In refactoring tests, I noticed that this change often results in (to my mind) clearer game code and will hopefully help to clear up ongoing confusion about `ctx` being different in moves and on the client. Types are also greatly improved.

- One thing this helped clarify for me, was that `playerID` (currently `ctx.playerID`) is _only_ available in moves, not in any hooks. Given the slightly confusing nature of `playerID` in hooks, this may not be a bad thing. For example, it’s not necessarily easy to understand why `playerID` in `turn.onBegin` is the ID of the player whose action caused the previous turn to end, not the ID of the player whose turn is beginning. That said, if we add `onEnd`/`onBegin` events for stages as proposed in #608, there will be a need to revisit this in the future. 

- While updating the docs, I noticed it wasn’t very clear what to call the whole object that contains `G`, `ctx` etc. I think of this whole lot as “context” for lack of a better word, but that would be confusing as `ctx` is already in use. I wonder if in fact, this refactor might be an opportunity to rename `ctx` itself to something more descriptive. Not sure what that name would be, but `flow` comes to mind.

    ```ts
    (context: { G: any; flow: Flow; ... }) => {
      const { G, flow } = context;
      console.log(flow.turn, flow.currentPlayer);
      if (!flow.activePlayers) { /* ... */ }
    }
    ```
- I’m not sure when will be best to merge and release this as obviously it’s a significant breaking change for everyone, but I wanted to get the ball rolling and we can keep this PR up-to-date as other changes arrive on the main branch.

## Migration Paths

This change is pretty impactful across game definitions. Here are some migration strategies.

### 1. Rewrite your functions

Ideally, you would rewrite your game to use the new syntax. In general this will result in simpler code. For example:

```diff
// Destructure G, ctx, and plugin APIs from the first argument
- function move(G, ctx, arg) {
+ function move({ G, random, playerID }, arg) {

// Access plugin APIs directly instead of from ctx
-   G.diceRoll = ctx.random.D6();
+   G.diceRoll = random.D6();

// Access playerID directly instead of from ctx
-   G.players[ctx.playerID].score += G.diceRoll;
+   G.players[playerID].score += G.diceRoll;

}
```

### 2. Use a stop-gap plugin for compatibility

If you don’t want to rewrite your moves, you can use a custom plugin to wrap those functions so they continue to receive the old style of arguments.

```js
const CompatibilityPlugin = {
  name: 'compat-plugin',

  fnWrap: (fn) => ({ G, ctx, ...api }, ...args) => {
    const ctxWithAPI = { ...ctx, ...api };
    return fn(G, ctxWithAPI, ...args);
  },
};

const game = {
  plugins: [CompatibilityPlugin],

  moves: {
    oldStyle: (G, ctx) => {
      // Use ctx.random etc. as before
    },
  },
};
```

### 3. Gradual migration

Using a plugin is an all-or-nothing approach — it forces you to continue using the old-style of arguments. If you want to gradually migrate your code, you’ll need to wrap only functions you haven’t updated yet. To do this we could adapt the `fnWrap` method from above to selectively wrap certain moves.

```js
// fnWrap adapted to account for long-form move objects, which
// boardgame.io does automatically when applying plugin wrappers
const moveAdapter = (moveFnOrObj) => {
  const isLongFormMove = typeof moveFnOrObj.move === 'function';
  const fn = isLongFormMove ? moveFnOrObj.move : moveFnOrObj;
  const move = ({ G, ctx, ...api }, ...args) => {
    const ctxWithAPI = { ...ctx, ...api };
    return fn(G, ctxWithAPI, ...args);
  };
  return isLongFormMove ? { ...moveFnOrObj, move } : move;
};

const game = {
  moves: {
    oldStyle: moveAdapter((G, ctx, arg) => {
      // Use ctx.random etc. as before
    }),
    newStyle: ({ G, ctx }, arg) => {},
  },
};
```

Depending on how your project is structured, you may be able to simplify applying this adapter rather than doing it manually for each function, for example if you store your moves something like this:

```js
const moves = { rollDice, moveToken, capture, build };
```

You could apply the adapter programmatically:

```js
for (const key in moves) {
  moves[key] = moveAdapter(moves[key]);
}
```

Then you could migrate by splitting your moves into two collections:

```js
const oldstyleMoves = { rollDice, build };
const migratedMoves = { moveToken, capture };

for (const key in oldstyleMoves) {
  oldstyleMoves[key] = moveAdapter(oldstyleMoves[key]);
}

const moves = { ...oldstyleMoves, ...migratedMoves };
```

#### Use Typescript to help find errors

If you’re already using TypeScript, make sure you use boardgame.io’s bundled types for your game to catch changes:

```ts
import type { Game, Move } from 'boardgame.io';

interface G {
  score: number;
}

const game: Game<G> = { /* ... */ };
const rollDice: Move<G> = ({ G, random }) => { /* ... */ };
```

If you’re writing plain JavaScript, you can often still use TypeScript types to get some in-editor hints by using Typescript’s [JSDoc syntax](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html) (VS Code does this well for example).

```js
/**
 * @typedef {{ score: number }} G
 */

/**
 * @type {import('boardgame.io').Game<G>}
 */
const game = { /* ... */ };

/**
 * @type {import('boardgame.io').Move<G>}
 */
const rollDice = ({ G, random }) => { /* ... */ };
```


src/core/flow.ts
```diff
@@ -23,6 +23,7 @@ import type {
   ActionShape,
   State,
   Ctx,
+  FnContext,
```
nicolodavis, line 26: I do agree with the point that it probably makes sense to call the aggregate (G + ctx + ...) as `Context`, which begs the question about what to do to `ctx`.

Do you think `ctx: GameContext` is a suitable name combination?   I'm less convinced by `flow`.

nicolodavis (commented):
We can probably punt on bringing the bot code to be consistent with the new format, but I think having `playerView` and `undoable` adopt the new format right away would prevent bugs in user code when we release this.

src/core/flow.ts
delucis: I’ll try to think further on what it might be named. I suppose I’m trying to avoid `context.ctx` as it seems rather confusing and vague to have a context inside a context. Some other options to brainstorm from (but something terser would be better): `playState`, `flowState`, `engineState`. I personally like how `flow` combines with the various members, e.g. `flow.currentPlayer`, `flow.playOrder`, `flow.phase` etc. while not being too verbose, but perhaps there’s something better.

That said, `G` + `ctx` is a pretty core concept for current users, so I could also definitely understand keeping it as is if we can’t find anything better. (Although previously a move could also have been `(G, foo, ...args) => {}` so `ctx` was only the conventional name for that second argument.)

cwatsonc:
Is this change still proposed or overcome by events?  Context.ctx seems confusing to me but flow somehow less appealing. The stage hooks might afford better outcomes in state modeling which is my real interest -- this is now a year old PR and might be too stale to consider but like that you are considering evolving the state model to help with clarity -- it was the biggest lift in _trying to understand_ bgio when I took first steps.

delucis:
@cwatsonc This branch is up to date and I do intend to merge it. Potentially soon, as I quite like the idea of this bigger change landing in v0.50, which is our next release.

Just to emphasise: there are no functionality changes or features involved here, it’s just a question of how arguments are provided to game functions, i.e. `(G, ctx) => {}` vs a new `{{ G, ctx }) => {}`.

YoavHortman:
Will this be merged?
This is a huge improvement over the current state of things.

vdfdev:
I received feedback that people really want this and are already using the alpha. I vote for finishing reviewing/merging all other outstanding PRs, doing a minor release. After that, we merge this PR and do a major release. What do you think, @delucis ?

delucis:
Major would mean boardgame.io@1.0.0. Not totally opposed but I think we could also release this as a minor (0.50.0 was my target for merging this) — minors below v1 are treated as major semver by npm, so it’s OK to introduce breaking changes from minor to minor.

So if we have stuff we want to get out before this, that could be 0.50.0 and then this as 0.51.0 or 1.0.0 if we’re ready! (One thing I had always thought needed work prior to v1 was the bot/AI architecture, but improving that can be a v2 goal too if we want).

delucis:
Sounds good. Thanks @vdfdev! 🙌

vdfdev (dismissed):

src/core/flow.ts
vdfdev: I dont think this is a major problem, because the Context in this case is only used internally. Most of our docs were updated to use the object destructuring assignments, so it doesnt look confusing for users.

vdfdev:
Uh oh, I think I borked the merge, will revert and fix it.

vdfdev (dismissed):

vdfdev:
Seems like https://github.com/boardgameio/boardgame.io/commit/aa99a9cce28012cb747fa6db8b3f8ad73c28be0a conflicted with this PR. I was able to fix one test, but there is one that still fails. I will keep trying to fix it

vdfdev (approved):

on3iro:
@delucis Hey, I have a pretty huge bgio codebase and after the update got over 500 typescript errors.
I wanted to use the stop-gap-plugin described above, but as far as I can tell this will only cover the issue at runtime. Typescript will still complain about incorrect signatures. Is there a way to fix this or am I stuck with either fully migrating or staying with a version <50?

PR #896, opened 2021-01-24
Adds check to throw error if non-serializable state is used in a move

flamecoals:
This check is meant to be used only in unit-tests to avoid breaking games that might be working (by chance) in production.

#### Checklist

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -471,6 +471,19 @@ describe('move dispatchers', () => {
     expect(store.getState().G).toMatchObject({ moved: '0' });
   });
 
+  test('with undefined playerID - multiplayer mode', () => {
+    const store = createStore(reducer, initialState);
+    const api = createMoveDispatchers(
+      game.moveNames,
+      store,
+      undefined,
+      null,
+      true
+    );
+    api.B();
+    expect(store.getState().G).toMatchObject({ moved: undefined });
+  });
+
   test('with null playerID - singleplayer mode', () => {
     const store = createStore(reducer, initialState);
     const api = createMoveDispatchers(game.moveNames, store, null);
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -471,19 +471,6 @@ describe('move dispatchers', () => {
     expect(store.getState().G).toMatchObject({ moved: '0' });
   });
 
-  test('with undefined playerID - multiplayer mode', () => {
-    const store = createStore(reducer, initialState);
-    const api = createMoveDispatchers(
-      game.moveNames,
-      store,
-      undefined,
-      null,
-      true
-    );
-    api.B();
-    expect(store.getState().G).toMatchObject({ moved: undefined });
-  });
-
   test('with null playerID - singleplayer mode', () => {
     const store = createStore(reducer, initialState);
     const api = createMoveDispatchers(game.moveNames, store, null);
--- a/src/plugins/plugin-serializable.ts
+++ b/src/plugins/plugin-serializable.ts
@@ -39,14 +39,12 @@ const SerializablePlugin: Plugin = {
 
   fnWrap: (move: AnyFn) => (G: unknown, ctx: Ctx, ...args: any[]) => {
     const result = move(G, ctx, ...args);
-    /* istanbul ignore if */
-    if (process.env.NODE_ENV === 'production') {
-      // Do not perform this check in production.
-      return result;
-    }
-    if (!isSerializable(result)) {
-      throw new Error(`Move state is not JSON-serialiazable. See 
-      https://boardgame.io/documentation/#/?id=state for more information.`);
+    // Check state in non-production environments.
+    if (process.env.NODE_ENV !== 'production' && !isSerializable(result)) {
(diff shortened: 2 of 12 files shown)


flamecoals:
Context: 
This is a follow-up on this issue https://github.com/freeboardgames/FreeBoardGames.org/issues/754
caused by a bad bug in our most popular game (checkers): https://github.com/freeboardgames/FreeBoardGames.org/issues/750
caused by a well-meaning refactor that slipped through code review that started using classes which are not serializable, and broke only on online mode (and still worked in local mode)

src/client/client.test.ts
```diff
@@ -471,19 +471,6 @@ describe('move dispatchers', () => {
...
-    const api = createMoveDispatchers(
-      game.moveNames,
-      store,
-      undefined,
-      null,
-      true
-    );
-    api.B();
-    expect(store.getState().G).toMatchObject({ moved: undefined });
-  });
```
flamecoals, line 485: FYI had to remove this because this test started failing. { move: undefined } is not a serializable state, so I think the test was pointless.

delucis (changes requested):
Thanks for this @flamecoals! I think it’s a great idea — I’ve seen a few people trip up on this after prototyping locally and then moving to a server.

A couple of things:

- Not everyone writes tests, but it would be nice to provide this feedback to everyone, so I would suggest moving the check behind `process.env.NODE_ENV !== 'production'`. Bundlers should then strip it out of production builds, but the errors will be surfaced more quickly if someone starts trying to add classes, functions, Maps, Sets or DOM nodes (!!) to state.

- Using parse/stringify/deepStrictEqual seems to throw up a few false positive as you noted with state like `{ moved: undefined }`. (While it’s true that that becomes `{}`, I would still expect that to pass the test as these are equivalent in practice.)

    I would suggest using a more explicit check like this:

    ```ts
    import isPlainObject from "lodash/isPlainObject";

    /**
     * Check if a value can be serialized (e.g. using `JSON.stringify`).
     * Adapted from: https://stackoverflow.com/a/30712764/3829557
     */
    function isSerializable(value: any) {
      // Primitives are OK.
      if (
        value === undefined ||
        value === null ||
        typeof value === 'boolean' ||
        typeof value === 'number' ||
        typeof value === 'string'
      ) {
        return true;
      }

      // A non-primitive value that is neither a POJO or an array cannot be serialized.
      if (!isPlainObject(value) && !Array.isArray(value)) {
        return false;
      }

      // Recurse entries if the value is an object or array.
      for (const key in value) {
        if (!isSerializable(value[key])) return false;
      }

      return true;
    };
    ```

What do you think?

flamecoals:
Great! I agree with both points. I will patch the PR soon

delucis:
@flamecoals Great, thanks!

We could also potentially add this logic as a new default plugin to avoid adding extra stuff to the core game function:

```js
const PluginCheckSerializable = {
  fnWrap: (fn) => (G, ctx, ...args) => {
    G = fn(G, ctx, ...args);
    if (!isSerializable(G)) throw new Error(/* ... */);
    return G;
  };
};
```

flamecoals:
I didn't forget about this PR, life has just been extra busy on the past few days (I was moving houses). I think I will find some time to work on this soon

delucis:
@flamecoals Thanks for letting us know. No rush!

flamecoals:
@delucis @nicolodavis sorry for the delay, I finally finished setting up everything in the new house.

I think I addressed all the points above, PTAL.

src/plugins/plugin-serializable.ts
```diff
@@ -0,0 +1,55 @@
...
+    const result = move(G, ctx, ...args);
+    /* istanbul ignore if */
+    if (process.env.NODE_ENV === 'production') {
+      // Do not perform this check in production.
+      return result;
+    }
+    if (!isSerializable(result)) {
+      throw new Error(`Move state is not JSON-serialiazable. See 
+      https://boardgame.io/documentation/#/?id=state for more information.`);
+    }
```
delucis, line 50: This can be simplified like this I think (which also avoids the need for an istanbul ignore comment). I’ve adjusted the error message as the template literal would include the indentation after the new line.

```suggestion
    // Check state in non-production environments.
    if (process.env.NODE_ENV !== 'production' && !isSerializable(result)) {
      throw new Error(
        'Move state is not JSON-serialiazable.\n' +
          'See https://boardgame.io/documentation/#/?id=state for more information.'
      );
    }
```

delucis (changes requested):
Looks good @flamecoals. I have one small suggestion to refactor the plugin code, but this looks fundamentally solid. Thanks!

src/plugins/plugin-serializable.ts
flamecoals: In general, I think the use of guard statements reduces the nesting of the code and makes it easier to read, but in this specific case either way works as the code is quite simple

flamecoals:
Accepted suggestion

delucis (approved):

delucis:
@flamecoals Thanks for this!

PR #898, opened 2021-01-27
Add details to exceptions raised in LobbyClient

c-w:
This pull request customizes the error type used in `LobbyClient` to enable passing-through to callers additional details about the specific error that the client encountered. Amongst others, this behavior is useful in situations where the errors thrown by the Lobby API may contain remediation hints for the user. For example, when using custom setupData validation (see https://github.com/boardgameio/boardgame.io/pull/831), the errors returned by the API may contain instructions to the user setting up a game on what exactly is incorrect about their configuration.

An alternative implementation could also expose the `response` object directly on `LobbyClientError` which would allow more flexibility to callers but also increase coupling with the client's implementation (e.g. changing away from `fetch` would be a breaking change).

#### Checklist

### DIFF
--- a/packages/client.ts
+++ b/packages/client.ts
@@ -7,6 +7,6 @@
  */
 
 import { Client } from '../src/client/client';
-import { LobbyClient } from '../src/lobby/client';
+import { LobbyClient, LobbyClientError } from '../src/lobby/client';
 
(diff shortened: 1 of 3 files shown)


delucis (approved):
@c-w This looks great — thanks!

PR #912, opened 2021-02-16
Create CODE_OF_CONDUCT.md

nicolodavis:


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,21 @@
...
+
+## What to do if you see a violation of these guidelines
+
+You may do either or both of the following:
+
+1. Politely message the perpetrator to steer them in the right direction.
+
+1. Contact a moderator privately (or email moderators@boardgame.io) to report bad behavior and request intervention.
+
+Moderators will encourage better behavior or issue a warning as appropriate. Further action (like banning a member) may be necessary as a last resort when other measures fail.
```
delucis, line 21: Thanks for putting this together!

Should we specify who the current moderators are here? I'm assuming it would be me and you for now, but it would be nice to clarify who receives emails sent to the moderators address.

nicolodavis: Sure, that makes sense.

delucis (approved):

Issue #592, opened 2020-03-29
Caller needs to know move returned as `INVALID_MOVE`


flamecoals:
It would be useful to also know the reason of the invalid move, so allow for some extra data specific to the game.

delucis:
To continue the thread from Gitter. Couple of options/considerations:

### Moves could throw an error

```js
// define custom error class
class InvalidMoveError extends Error {
  constructor(data) {
    super('INVALID_MOVE');
    this.name = "InvalidMoveError";
    this.data = data;
  }
}
```
```js
// move definition
const { InvalidMoveError } = require('boardgame.io/core');

function move() {
  // move is invalid, throw an error
  throw new InvalidMoveError({ message: 'something went wrong', player: '2' });
}
```
```js
// process move
try {
  moveFn(G, ctx, ...args);
} catch (error) {
  if (!(error instanceof InvalidMoveError)) throw error;
  // return error.data to client
}
```

#### Notes

- It’s idiomatic Javascript — an invalid move is an error.


### Moves could return an error object

```js
// define error creator
function InvalidMove(data) {
  return { error: 'INVALID_MOVE', data }; // for example
}
```
```js
const { InvalidMove } = require('boardgame.io/core');

function move() {
  // move is invalid, return an error object
  return InvalidMove({ message: 'something went wrong', player: '2' });
}
```
```js
// process move
const result = moveFn(G, ctx, ...args);
if (result && result.error === 'INVALID_MOVE') {
  // move is invalid, return result.data the client
}
```

#### Notes

- This is a little closer to the current `return INVALID_MOVE` approach than throwing an error.


### `ctx` could have an error method

On the surface this seems pretty similar to an `InvalidMove` function to me, just that the function is attached to `ctx` instead of exported in core.

The major advantage of this would presumably be that it could hook into `ctx` in deeper ways — perhaps firing other kinds of changes to state or flow. What would those be?

nicolodavis:
One way to work around the asynchronous issue is to return a Promise from the move dispatcher.  We just need to ensure that the client can still be called in a synchronous way when using an in-memory store.

nicolodavis:
Of the approaches that we've discussed so far, I think I'm leaning toward moves returning an error object.

jeetmehta:
👋 Hi there! My friends and I are using boardgame.io and the framework has been great. I'm currently trying to indicate to the user/player _why_ their move might be invalid.

[...]

PR #918, opened 2021-03-07
Updating logging methods to be spied upon after framework is loaded. #917

### LINKED ISSUE #917: Updating logging functions for spying after framework is loaded
Tangentially related to #592, I propose we update how we call the logging methods `console.error` and `console.log` so they can be properly spied upon during tests after the framework has loaded. 

For context i'm using Angular CLI to generate an app. I then pull in boardgame.io and start writing tests. Angular CLI is using Typescript  4.1.2 and jasmine 3.6.0. When writing a test it looks something like this:

```
import { Client } from 'boardgame.io/client';
...


describe('MyGame', () => {
  beforeEach(() => {
    spyOn(console, 'error');
    game = Client({ game: MyGame });
    ...
  });

  it('should not do a thing', () => {
    game.moves.someMove();
    expect(console.error).toHaveBeenCalledWith('ERROR:', `disallowed move: someMove`);
  });
});
```

At the moment the expectation will fail because the boardgame code is being loaded before the spy can be created. Per #592 there is no other good way to determine if an invalid move has been made (without in depth state checking) so spying on the error logs is the only other approach. Although I suspect spying on the logs can provide other test benefits besides just invalid move testing. Since the boardgame code is binding directly to the `console.error` function reference there is no way to spy on it after the fact. 

IF spying on logging functions after the codebase has been loaded is a valid use-case then I can think of two possible approaches:
1. Allowing the client to provide both a log and error function which defaults to `console.XYZ` method, respectively.
2. Update the logger implementation to reference a function that calls the `console.XYZ` methods.

cha55son:
Closes #917 

Allows logging to be spied after the framework has been loaded. Eases testing of invalid moves.

### DIFF
--- a/src/core/logger.test.js
+++ b/src/core/logger.test.js
@@ -51,6 +51,9 @@ describe('logging', () => {
       jest.resetModules();
       logging = require('./logger');
     });
+    afterAll(() => {
+      process.env.NODE_ENV = oldNodeEnv;
+    });
 
     test('info stripped', () => {
       logging.info('msg2');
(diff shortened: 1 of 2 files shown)


delucis (approved):
Thanks @cha55son!

Worth noting that logging is an internal API so we probably can’t guarantee that at some future time your tests might be broken by internal changes, but I appreciate that in the absence of #592 (or #723) it’s the best option available currently.

Issue #190, opened 2018-04-26
TypeScript integration


shirakaba:
I don't see any TypeScript definition files, nor an `@types/boardgame.io` module on `npm`.

1. Would addition of TypeScript definition files be welcomed *(this would be totally non-intrusive, and can be done incrementally, but would have to be maintained every time any JS API updates)*?
2. Would actually migrating all the JS source to TypeScript be welcomed *(the type definitions would be future-proofed, and we could use any level of ES features throughout the code, transpiling down to whatever target necessary; but it may alienate contributors only knowledgeable of JS)*?

chrisheninger:
@shirakaba if you'd like a starting point– @Felizardo has a file in his turnato repo:

https://github.com/Felizardo/turnato/blob/master/%40types/boardgame.io/index.d.ts

There was also a little discussion around TS when discussing the Angular client: https://github.com/google/boardgame.io/issues/131

shirakaba:
@chrisheninger That's invaluable, thank you!

Would a single-file declaration describing the whole codebase via many `declare module "x" {}` statements be preferable to declaration files alongside each JS file, then?

shirakaba:
For now, I've had to do declaration files alongside each JS file because I'm unsure how else to make test files for the typings. The test files are similarly littering the UI folder. I've also been incrementally adding the types into one global declaration file so that we have both options. You can see my progress [on my fork, on the shirakaba/typescript branch](https://github.com/shirakaba/boardgame.io/tree/shirakaba/typescript).

I've so far typed all the src in the UI folder, but still have a few issues to resolve...

The foremost issue is that I don't know how to correctly type Logo and Card (which both use stateless component functions – what's more, they're assigned to const variables, and Logo adds in a default export to increase the headache further). Some help there would be appreciated.

* Generally, I'm surprised that the UI components have so many optional features, and that many of them expect just a single child rather than multiple children.

* In the `ui/token.js`, `defaultProps.template` references the `Square` class but doesn't construct an instance of it; is this a bug?

* What templates may Token accept? Just `Square`s?

There's plenty more work to be done, still. Would welcome any extra pairs of hands.

shirakaba:
> The foremost issue is that I don't know how to correctly type Logo and Card (which both use stateless component functions

I figured it out with a little reference to https://github.com/piotrwitek/react-redux-typescript-guide#component-typing-patterns and some guesswork.

[...]

PR #922, opened 2021-03-16
feat: improve typescript typings for the player plugin and react client component

brebory:
## Overview
This fixes the issue I brought up in #190 where the `PlayerPlugin<T>` interface was not part of the public api.

Additionally, there is a second fix for a related issue I ran into when using the React `Client` component where the `game` property used only `Ctx` in the type signature instead of the generic type constraint `ContextWithPlugins`.

This caused the code setting up the client to not compile when plugins were used.

### Steps to Reproduce

1. Create a game definition that uses `Ctx` extended with plugins:

```ts
interface CustomGameContext implements Ctx {
  // Extend the context with properties and methods
}

const CustomGameDefinition: Game<any, CustomGameContext, any> = {
  // game definition elided for brevity
};
```

2. Use the game definition with the React Client component:

```ts
const App = Client<CustomGameState, BoardProps<CustomGameState>, CustomGameContext>({ game: CustomGameDefinition });
```

Notice the following error on build:

```
Type 'Game<CustomGameState, CustomGameContext, any>' is not assignable to type 'Game<CustomGameState, Ctx, any>'.
  Types of property 'setup' are incompatible.
    Type '((ctx: CustomGameContext, setupData?: any) => any) | undefined' is not assignable to type '((ctx: Ctx, setupData?: any) => any) | undefined'.
      Type '(ctx: CustomGameContext, setupData?: any) => any' is not assignable to type '(ctx: Ctx, setupData?: any) => any'.
        Types of parameters 'ctx' and 'ctx' are incompatible.
          Type 'Ctx' is not assignable to type 'CustomGameContext'.
            Type 'Ctx' is not assignable to type 'PlayerPlugin<PlayerState>'.ts(2322)
client.d.ts(15, 5): The expected type comes from property 'game' which is declared here on type 'ReactClientOpts<CustomGameState, State<CustomGameState, Ctx> & { isActive: boolean; isConnected: boolean; log: LogEntry[]; } & Omit<...> & ExposedClientProps<...> & { ...; }, Ctx>'
```

Note, in my example I extended `Ctx` with the `PlayerPlugin<T>` interface, but all custom plugins that extend `Ctx` would have the same problem.

#### Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -16,3 +16,4 @@ Satana Charuwichitratana
 Pete Nykänen
 Philihp Busby
 Jason Harrison
+Brendon Roberto
--- a/packages/plugins.ts
+++ b/packages/plugins.ts
@@ -7,5 +7,6 @@
  */
 
 import PluginPlayer from '../src/plugins/plugin-player';
+import { PlayerPlugin } from '../src/plugins/plugin-player';
 
-export { PluginPlayer };
+export { PluginPlayer, PlayerPlugin };
--- a/src/client/react.tsx
+++ b/src/client/react.tsx
@@ -10,6 +10,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Client as RawClient } from './client';
 import type { ClientOpts, ClientState, _ClientImpl } from './client';
+import type { Ctx } from '../types';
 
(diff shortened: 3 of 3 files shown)


delucis (approved):
@brebory Thanks for this!

PR #931, opened 2021-04-14
Dissolutio/docs add chat guide

### LINKED ISSUE #911: missing documentation for chat
I think title is self-explaining.

Dissolutio:
Closes #911 

#### Checklist

* [ CHECK] Use a separate branch in your local repo (not `master`).
* [UNTESTED ] Test coverage is 100% (or you have a story for why it's ok).


delucis:
@Dissolutio Thanks for this! I’m going to tweak it slightly, but it looks good to me.

One question. In #911 you mentioned:

> add warning about needing to generate your own unique id's for chats?

My understanding is that the client does this automatically:

https://github.com/boardgameio/boardgame.io/blob/53142fd7f7adaf073635279ef7ce5d037c471386/src/client/client.ts#L338

Did you mean something else?

Dissolutio:
Oh I see. I thought I needed to pass the id and playerID to sendChatMessage, because of its Type.

[BoardProps.sendChatMessage](https://github.com/boardgameio/boardgame.io/blob/53142fd7f7adaf073635279ef7ce5d037c471386/src/client/client.ts#L165) requires the argument [ChatMessage](https://github.com/boardgameio/boardgame.io/blob/53142fd7f7adaf073635279ef7ce5d037c471386/src/types.ts#L437-L441)

delucis:
@Dissolutio Oh, good catch! That’s actually the wrong type on the client there. It should be:

```ts
  sendChatMessage: (message: any) => void;
```

delucis (approved):

delucis:
OK, I was able to confirm that that was the case and I’ve fixed that typing. Will be released in v0.44.1 shortly.

Thanks again for this PR!

Issue #910, opened 2021-02-06
Custom plugins aren't available in onBegin/onEnd


MathieuLoutre:
Hello there! I'm using @delucis' [bgio-effects](https://github.com/delucis/bgio-effects) and I'm trying to access it inside a phase's `onEnd` hook but it comes out as `undefined` while `ctx.random` is there alongside the other default plugins.

@delucis created a [reproduction](https://codesandbox.io/s/bgio-effects-undefined-bug-ob45w) avec a short [chat](https://gitter.im/boardgame-io/General?at=601c472d24cd6b60d81b49ae).

I did a bit more digging writing some failing tests and discovered that there seems to be link with the `flush` of the plugins. Plugins without flush are available and if a plugin has a `flush` it's not. I'm not too familiar with the structure of boardgame.io so I mostly dug around with some console logs to track it down.

When this [part of the code](https://github.com/boardgameio/boardgame.io/blob/master/src/plugins/main.ts#L174) executes, `pluginState` contains all the right things, but only the [data gets written](https://github.com/boardgameio/boardgame.io/blob/master/src/plugins/main.ts#L187) which removes the API that the plugin provides and thereafter the relevant plugin in `state.plugins` is just an empty `data` object.

I'm happy to provide a PR/fix but I'm not too sure what is supposed to happen. Would adding back the rest of the state like [here](https://github.com/boardgameio/boardgame.io/blob/master/src/plugins/main.ts#L157) be a simple fix that doesn't disturb anything else?

MathieuLoutre:
Replacing `[plugin.name]: { data: newData }` on [L187](https://github.com/boardgameio/boardgame.io/blob/master/src/plugins/main.ts#L187) with `[plugin.name]: { ...pluginState, data: newData }` makes the tests pass (and no other test fails). However, as I mentioned, I've no idea of the repercussions of this down the line or why it was "data only" in the first place.

delucis:
Thanks for the detailed report @MathieuLoutre! My understanding is that `flush` intentionally removes plugin APIs, so that they are not persisted and we don’t try to serialize them and send them to the client [1]. The APIs are created by calling `Plugins.Enhance`.

For example, the current flow appears to be:

1. `InitializeGame`:
    - `Plugins.Setup` creates initial data in `state.plugins`.
    - `Plugins.Enhance` creates the APIs in `state.plugins`.
    - `Plugins.EnhanceCtx` applies the APIs to `state.ctx`, for consumption by game code.
[...]

PR #932, opened 2021-04-22
Fix custom plugins not available in events hooks. Add test.

MathieuLoutre:
#### Checklist

This changes the order the plugins are flushed in to fix the problem mentioned in #910. It also adds a test to make sure it works.

This removes the additional reverse as it seemed necessary to separate PluginEvents from the rest (hence CORE_PLUGINS and DEFAULT_PLUGINS) to execute everything in the right order.

### DIFF
--- a/src/plugins/main.ts
+++ b/src/plugins/main.ts
@@ -167,47 +167,50 @@ export const Enhance = (
  * Allows plugins to update their state after a move / event.
  */
 export const Flush = (state: State, opts: PluginOpts): State => {
-  // We flush the events plugin first, then custom plugins and the core plugins
-  // This means custom plugins cannot use the events API but will be available in event hooks
-  [PluginEvents, ...opts.game.plugins, ...CORE_PLUGINS].forEach((plugin) => {
-    const name = plugin.name;
-    const pluginState = state.plugins[name] || { data: {} };
+  // We flush the events plugin first, then custom plugins and the core plugins.
+  // This means custom plugins cannot use the events API but will be available in event hooks.
+  // Note that plugins are flushed in reverse, to allow custom plugins calling each other.
+  [...CORE_PLUGINS, ...opts.game.plugins, PluginEvents]
+    .reverse()
+    .forEach((plugin) => {
+      const name = plugin.name;
+      const pluginState = state.plugins[name] || { data: {} };
 
-    if (plugin.flush) {
-      const newData = plugin.flush({
-        G: state.G,
-        ctx: state.ctx,
-        game: opts.game,
-        api: pluginState.api,
-        data: pluginState.data,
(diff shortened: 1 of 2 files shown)


delucis:
Thanks @MathieuLoutre! One quick thought regarding removing the `reverse`. Currently, you could build plugins on top of each other, something like:

```js
const PluginA = {
  name: 'A',
  api: () => ({
    getN: () => 5,
  }),
};

const PluginB = {
  name: 'B',
  setup: ({ ctx }) => ({
    n: ctx.A.getN(),
  }),
  flush:
};

const game = {
  plugins: [PluginA, PluginB],
};
```

src/plugins/main.ts
```diff
@@ -172,9 +167,9 @@ export const Enhance = (
  * Allows plugins to update their state after a move / event.
  */
 export const Flush = (state: State, opts: PluginOpts): State => {
-  // Note that we flush plugins in reverse order, to make sure that plugins
-  // that come before in the chain are still available.
-  [...DEFAULT_PLUGINS, ...opts.game.plugins].reverse().forEach((plugin) => {
+  // We flush the events plugin first, then custom plugins and the core plugins
+  // This means custom plugins cannot use the events API but will be available in event hooks
+  [PluginEvents, ...opts.game.plugins, ...CORE_PLUGINS].forEach((plugin) => {
```
delucis, line 172: Currently, it is possible to build custom plugins on top of each other, something like:

```js
const PluginA = {
  name: 'A',
  api: () => ({
    getFive: () => 5,
  }),
};

const PluginB = {
  name: 'B',
  setup: ({ ctx }) => ({
    n: ctx.A.getFive(),
  }),
  flush: ({ ctx, data }) => ({
    n: data.n + ctx.A.getFive(),
  }),
};

const game = {
  plugins: [PluginA, PluginB],
};
```

Here `PluginA` provides some API methods, which `PluginB` consumes. Without reversing the user-provided plugins, this breaks during `flush`, because `PluginA` will be removed before `PluginB.flush` is called.

To continue support for this pattern we should probably keep `reverse` here, like this:

```suggestion
  [...CORE_PLUGINS, ...opts.game.plugins, PluginEvents].reverse().forEach((plugin) => {
```

delucis (changes requested):
Thanks @MathieuLoutre! I think we need to keep `reverse` (see comment), but otherwise this is great!

src/plugins/main.ts
MathieuLoutre: Makes sense!

MathieuLoutre:
Great! I've added the `reverse` and a comment to that effect — hope it's clear enough!

delucis (approved):

Issue #892, opened 2021-01-18
Stepping through log outside of debug?


joepinion:
Is there a way, built-in already, to take the initial game state and step through the log to receive a G object back at each step?

I see the ability to do this in debug, but how can we do this in the client regular user interface? The use case is, being able to rewind time to review previous moves, especially after loading the page mid-game.

delucis:
There isn’t, although it might be possible to adapt [the debug panel logic][debug]. You can `import { CreateGameReducer } from 'boardgame.io/internal'`, get `client.initialState`, and then replay each log action as in the debug panel. There’s some Svelte-specific syntax in the linked code, but `let { log } = $client` is basically getting `log` from the `client.subscribe` callback, so it’s equivalent to:

```js
client.subscribe((state) => {
  if (!state) return;
  let { log } = state;
  // ...
});
```

But that requires access to the plain JavaScript client rather than the React client.

For better reusability, it might make sense to move the log stepping logic into the client itself and expose some kind of API like:

```ts
interface LogStepAPI {
  stepForward: () => void;
  stepBack: () => void;
  stepTo: (idx: number) => void;
}
```

[debug]: https://github.com/boardgameio/boardgame.io/blob/dc96b2649d64b1de928c979ad45fd3375356e276/src/client/debug/log/Log.svelte#L14-L73

richardl62:
Currently the 'logs' box on the homepage reads as if logs and time travel are full supported.  As a short term measure would it be worth changing the text to make it clear that these features limited to the debug panel?

on3iro:
Hey folks,

I am currently building something similar. And had a question regarding the example @delucis linked to:
As far as I can tell, the log might have multiple action entries per `_stateID`. The example above accounts for automatic actions by skipping them, but there still maybe for example a `MAKE_MOVE`-action and a `GAME_EVENT` action for the same `_stateID`.
Now whenever the reducer runs it produces an updated `_stateID`. 

My assumption was that by running all actions for `_stateID = 0` I will end up at `_stateID = 1` however the new `_stateID` will be the original id + the amount of stepped through log entries event though there might be other moves with a lower `_stateID`.

For example my resulting `_stateID` will be `2` when I only call the reducer twice applying only the first two actions of the following log:

```ts
[
  {
    action: {
      type: 'MAKE_MOVE',
      payload: {
        type: 'pickCamps',
        args: [[0, 1, 2], '0'],
        playerID: '0',
      },
    },
    _stateID: 0,
[...]

PR #940, opened 2021-05-18
feature(#723): Improve client error handling [Part 1]

shaoster:
This PR sets up the initial reducer/master plumbing for error handling
by introducing a "transients" option bag containing errors.

The goal of this change is to be as backwards-compatible as possible,
with subsequent changes building on top of the API extensions added
here.

Notes:

- Previously, there was an assumption that the state resulting from a
reducer processing an action is suitable for client consumption.
- Now, transient artifacts might be appended onto the state and need to
be stripped before being sent to the client.
- To work around this change, I added some new types to help signal
the "transient" nature of State that enters and exits the reducer.
- I'm assuming the master is the single caller of the core reducer and
leaving it up to the master to do this stripping..

#### Checklist

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -14,7 +14,10 @@ import * as Actions from '../core/action-types';
 import * as ActionCreators from '../core/action-creators';
 import { ProcessGameConfig } from '../core/game';
 import type Debug from './debug/Debug.svelte';
-import { CreateGameReducer } from '../core/reducer';
+import {
+  CreateGameReducer,
+  TransientHandlingMiddleware,
+} from '../core/reducer';
 import { InitializeGame } from '../core/initialize';
 import { PlayerView } from '../plugins/main';
 import type { Transport, TransportOpts } from './transport/transport';
@@ -303,6 +306,7 @@ export class _ClientImpl<G extends any = any> {
     };
 
     const middleware = applyMiddleware(
+      TransientHandlingMiddleware,
       SubscriptionMiddleware,
       TransportMiddleware,
       LogMiddleware
--- a/src/core/reducer.test.ts
+++ b/src/core/reducer.test.ts
@@ -131,11 +131,16 @@ test('valid patch', () => {
 
 test('invalid patch', () => {
   const originalState = { _stateID: 0, G: 'patch' } as State;
-  const state = reducer(
+  const { transients, ...state } = reducer(
     originalState,
     patch(0, 1, [{ op: 'replace', path: '/_stateIDD', value: 1 }], [])
   );
   expect(state).toEqual(originalState);
+  expect(transients.error.type).toEqual('patch_failed');
+  // It's an array.
+  expect(transients.error.payload.length).toEqual(1);
+  // It looks like the standard rfc6902 error language.
+  expect(transients.error.payload[0].toString()).toContain('/_stateIDD');
 });
 
 test('reset', () => {
--- a/src/core/reducer.test.ts
+++ b/src/core/reducer.test.ts
@@ -7,7 +7,8 @@
  */
 
 import { INVALID_MOVE } from './constants';
-import { CreateGameReducer } from './reducer';
+import { applyMiddleware, createStore } from 'redux';
+import { CreateGameReducer, TransientHandlingMiddleware } from './reducer';
 import { InitializeGame } from './initialize';
 import {
   makeMove,
@@ -32,6 +33,7 @@ const game: Game = {
     A: (G) => G,
     B: () => ({ moved: true }),
     C: () => ({ victory: true }),
+    Invalid: () => INVALID_MOVE,
   },
   endIf: (G, ctx) => (G.victory ? ctx.currentPlayer : undefined),
 };
@@ -136,7 +138,7 @@ test('invalid patch', () => {
     patch(0, 1, [{ op: 'replace', path: '/_stateIDD', value: 1 }], [])
   );
   expect(state).toEqual(originalState);
-  expect(transients.error.type).toEqual('patch_failed');
+  expect(transients.error.type).toEqual('update/patch_failed');
   // It's an array.
   expect(transients.error.payload.length).toEqual(1);
   // It looks like the standard rfc6902 error language.
@@ -752,5 +754,35 @@ describe('undo / redo with stages', () => {
       C: false,
     });
     expect(state.ctx.activePlayers['0']).toBe('B');
+  });
+});
+
+describe('TransientHandlingMiddleware', () => {
+  const middleware = applyMiddleware(TransientHandlingMiddleware);
+  let store = null;
+
+  beforeEach(() => {
+    store = createStore(reducer, initialState, middleware);
+  });
+
+  test('regular dispatch result has no transients', () => {
+    const result = store.dispatch(makeMove('A'));
+    expect(result).toEqual(
+      expect.not.objectContaining({ transients: expect.anything() })
+    );
+    expect(result).toEqual(
+      expect.not.objectContaining({ stripTransientsResult: expect.anything() })
+    );
+  });
+
+  test('failing dispatch result contains transients', () => {
+    const result = store.dispatch(makeMove('Invalid'));
+    expect(result).toMatchObject({
+      transients: {
+        error: {
+          type: 'action/invalid_move',
+        },
+      },
+    });
   });
 });
--- a/src/core/reducer.test.ts
+++ b/src/core/reducer.test.ts
@@ -438,7 +438,15 @@ describe('undo / redo', () => {
 
   test('redo only resets deltalog if nothing to redo', () => {
     const state = reducer(initialState, makeMove('move', 'A', '0'));
-    expect(reducer(state, redo())).toEqual({ ...state, deltalog: [] });
+    expect(reducer(state, redo())).toMatchObject({
+      ...state,
+      deltalog: [],
+      transients: {
+        error: {
+          type: 'action/action_invalid',
+        },
+      },
+    });
   });
 });
 
@@ -533,13 +541,29 @@ describe('undo stack', () => {
 
   test('can’t undo at the start of a turn', () => {
     const newState = reducer(state, undo());
-    expect(newState).toEqual({ ...state, deltalog: [] });
+    expect(newState).toMatchObject({
+      ...state,
+      deltalog: [],
+      transients: {
+        error: {
+          type: 'action/action_invalid',
+        },
+      },
+    });
   });
 
   test('can’t undo another player’s move', () => {
     state = reducer(state, makeMove('basic', null, '1'));
     const newState = reducer(state, undo('0'));
-    expect(newState).toEqual({ ...state, deltalog: [] });
+    expect(newState).toMatchObject({
+      ...state,
+      deltalog: [],
+      transients: {
+        error: {
+          type: 'action/action_invalid',
+        },
+      },
+    });
   });
 });
 
@@ -595,7 +619,15 @@ describe('redo stack', () => {
     expect(state._redo).toHaveLength(1);
     const newState = reducer(state, redo('0'));
     expect(state._redo).toHaveLength(1);
-    expect(newState).toEqual({ ...state, deltalog: [] });
+    expect(newState).toMatchObject({
+      ...state,
+      deltalog: [],
+      transients: {
+        error: {
+          type: 'action/action_invalid',
+        },
+      },
+    });
   });
 });
 
--- a/src/core/reducer.ts
+++ b/src/core/reducer.ts
@@ -11,6 +11,7 @@ import * as plugins from '../plugins/main';
 import { ProcessGameConfig } from './game';
 import { error } from './logger';
 import { INVALID_MOVE } from './constants';
+import type { Dispatch } from 'redux';
 import type {
   ActionShape,
   Ctx,
@@ -20,9 +21,12 @@ import type {
   LongFormMove,
   Move,
   State,
+  Store,
+  TransientMetadata,
   TransientState,
   Undo,
 } from '../types';
+import { stripTransients } from './action-creators';
 import { ActionErrorType, UpdateErrorType } from './errors';
 import { applyPatch } from 'rfc6902';
 
@@ -125,6 +129,24 @@ function initializeDeltalog(
 }
 
 /**
+ * ExtractTransientsFromState
+ *
+ * Split out transients from the a TransientState
+ */
+function ExtractTransients(
+  transientState: TransientState | null
+): [State | null, TransientMetadata | null] {
+  if (!transientState) {
+    // We preserve null for the state for legacy callers, but the transient
+    // field should be undefined if not present to be consistent with the
+    // code path below.
+    return [null, undefined];
+  }
+  const { transients, ...state } = transientState;
+  return [state as State, transients as TransientMetadata];
+}
+
+/**
  * WithError
  *
  * Augment a State instance with transient error information.
@@ -145,6 +167,39 @@ function WithError<PT extends any = any>(
     },
   };
 }
+
+/**
+ * Middleware for processing TransientState associated with the reducer
+ * returned by CreateGameReducer.
+ * This should pretty much be used everywhere you want realistic state
+ * transitions and error handling.
+ */
+export const TransientHandlingMiddleware = (store: Store) => (
+  next: Dispatch<ActionShape.Any>
+) => (action: ActionShape.Any) => {
+  const result = next(action);
+  switch (action.type) {
+    case Actions.STRIP_TRANSIENTS: {
+      return result;
+    }
+    default: {
+      const [, transients] = ExtractTransients(store.getState());
+      if (typeof transients !== 'undefined') {
+        store.dispatch(stripTransients());
+        // Dev Note: If parent middleware needs to correlate the spawned
+        // StripTransients action to the triggering action, instrument here.
+        //
+        // This is a bit tricky; for more details, see:
+        //   https://github.com/boardgameio/boardgame.io/pull/940#discussion_r636200648
+        return {
+          ...result,
+          transients,
+        };
+      }
+      return result;
+    }
+  }
+};
 
 /**
  * CreateGameReducer
@@ -168,20 +223,15 @@ export function CreateGameReducer({
    * @param {object} action - A Redux action.
    */
   return (
-    state: TransientState | null = null,
+    stateWithTransients: TransientState | null = null,
     action: ActionShape.Any
   ): TransientState => {
+    let [state /*, transients */] = ExtractTransients(stateWithTransients);
     switch (action.type) {
       case Actions.STRIP_TRANSIENTS: {
         // This action indicates that transient metadata in the state has been
         // consumed and should now be stripped from the state..
-        const {
-          transients,
-          // Revert the game state to before the error so existing consumers do
-          // not see garbage.
-          ...strippedState
-        } = state;
-        return strippedState;
+        return state;
       }
 
       case Actions.GAME_EVENT: {
@@ -211,7 +261,7 @@ export function CreateGameReducer({
         }
 
         // Execute plugins.
-        state = plugins.Enhance(state as State, {
+        state = plugins.Enhance(state, {
           game,
           isClient: false,
           playerID: action.payload.playerID,
@@ -432,9 +482,6 @@ export function CreateGameReducer({
 
       case Actions.PLUGIN: {
         // TODO(#723): Expose error semantics to plugin processing.
-        // DevNote: The upcast to State is reasonable both from a typing
-        // and functionality perspective because plugin actions can only be
-        // invoked after any transient artifacts like errors have been cleaned.
         return plugins.ProcessAction(state, action, { game });
       }
 
--- a/src/core/reducer.ts
+++ b/src/core/reducer.ts
@@ -391,10 +391,8 @@ export function CreateGameReducer({
         const { _undo, _redo } = state;
 
         if (_undo.length < 2) {
-          // TODO(#723): Add an error case here.
-          //  error(`No moves to undo`);
-          //  return WithError(state, ActionErrorType.ActionDisabled);
-          return state;
+          error(`No moves to undo`);
+          return WithError(state, ActionErrorType.ActionInvalid);
         }
 
         const last = _undo[_undo.length - 1];
@@ -405,10 +403,8 @@ export function CreateGameReducer({
           actionHasPlayerID(action) &&
           action.payload.playerID !== last.playerID
         ) {
-          // TODO(#723): Add an error case here.
-          //  error(`Cannot undo other players' moves`);
-          //  return WithError(state, ActionErrorType.ActionDisabled);
-          return state;
+          error(`Cannot undo other players' moves`);
+          return WithError(state, ActionErrorType.ActionInvalid);
         }
 
         // If undoing a move, check it is undoable.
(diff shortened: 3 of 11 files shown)


package-lock.json
```diff
@@ -10967,7 +10967,8 @@
         },
         "yargs-parser": {
           "version": "13.1.1",
-          "resolved": "",
+          "resolved": "https://registry.npmjs.org/yargs-parser/-/yargs-parser-13.1.1.tgz",
```
shaoster, line 10970: Not sure if I should preserve these updates or strip.

src/plugins/main.ts
```diff
@@ -14,6 +14,7 @@ import PluginSerializable from './plugin-serializable';
 import type {
   AnyFn,
   PartialGameState,
+  TransientState,
```
shaoster, line 17: In a subsequent change I want to add more negative cases to this test suite. The added test is the only one so far that actually exercises a reducer negative case. In fact, there's a ton of duplicated error handling in the reducer and master that could just be delegated to the reducer in the next step of this feature.

package-lock.json
delucis: Right, npm seems to vacillate between resolving and not resolving this dependency. Not sure why, but we probably shouldn’t change package-lock in this PR.

delucis (changes requested):
Thanks @shaoster! Good job separating out this chunk of the work. Here are a few initial comments.

Regarding the general architecture, it’s a little more complicated than reducer → master → client. If you search for `CreateGameReducer` in the repo, you’ll find all the places it’s used, but here’s a quick overview:

1. The reducer runs [directly inside the client](https://github.com/boardgameio/boardgame.io/blob/dc6cd451d85b4cb8a857d21af06c66e68a217815/src/client/client.ts#L314), this allows the client to run without a server/local master and allows the client to update optimistically even before receiving an update from the server when running a multiplayer game.

2. When in use, the reducer runs in the master either on the server or inside `Local` (this is the scenario you already considered).

3. AI code [uses the reducer](https://github.com/boardgameio/boardgame.io/blob/dc6cd451d85b4cb8a857d21af06c66e68a217815/src/ai/ai.ts#L61) to run simulations. (The same is true in [the MCTS bot implementation](https://github.com/boardgameio/boardgame.io/blob/dc6cd451d85b4cb8a857d21af06c66e68a217815/src/ai/mcts-bot.ts#L88).)

4. Finally, we also currently run the reducer when stepping through items [in the Log pane of the debug panel](https://github.com/boardgameio/boardgame.io/blob/dc6cd451d85b4cb8a857d21af06c66e68a217815/src/client/debug/log/Log.svelte#L14).

#### Implications

- In the client we need to consider [the middleware added to the store](https://github.com/boardgameio/boardgame.io/blob/dc6cd451d85b4cb8a857d21af06c66e68a217815/src/client/client.ts#L219-L314) to see how exactly to handle the transient state and later surface `transients.error` when produced by the local reducer. As elsewhere, adding `transients` doen’t cause any actual errors, but presumably we’d want to strip it as in the master and use the error somehow.

- 3 & 4 can perhaps wait — adding the `transients` field shouldn’t actually cause problems here. It may later be possible to use `transients.error` to improve the logic in AI simulations and log stepping (this latter could perhaps be done as part of #892)

src/core/reducer.ts
```diff
@@ -142,8 +167,23 @@ export function CreateGameReducer({
    * @param {object} state - The state before the action.
    * @param {object} action - A Redux action.
    */
-  return (state: State | null = null, action: ActionShape.Any): State => {
+  return (
+    state: TransientState | null = null,
+    action: ActionShape.Any
+  ): TransientState => {
```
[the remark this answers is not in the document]
shaoster: At least from tests, the null case gets hit pretty often. Nonetheless I think the benefit of what you're describing is such that adding a little helper function here, to deal with the nulls, is worth it.

shaoster:
> Thanks @shaoster! Good job separating out this chunk of the work. Here are a few initial comments.
> 
> Regarding the general architecture, it’s a little more complicated than reducer → master → client. If you search for `CreateGameReducer` in the repo, you’ll find all the places it’s used, but here’s a quick overview:
> 
> 1. The reducer runs [directly inside the client](https://github.com/boardgameio/boardgame.io/blob/dc6cd451d85b4cb8a857d21af06c66e68a217815/src/client/client.ts#L314), this allows the client to run without a server/local master and allows the client to update optimistically even before receiving an update from the server when running a multiplayer game.
> 2. When in use, the reducer runs in the master either on the server or inside `Local` (this is the scenario you already considered).
> 3. AI code [uses the reducer](https://github.com/boardgameio/boardgame.io/blob/dc6cd451d85b4cb8a857d21af06c66e68a217815/src/ai/ai.ts#L61) to run simulations. (The same is true in [the MCTS bot implementation](https://github.com/boardgameio/boardgame.io/blob/dc6cd451d85b4cb8a857d21af06c66e68a217815/src/ai/mcts-bot.ts#L88).)
> 4. Finally, we also currently run the reducer when stepping through items [in the Log pane of the debug panel](https://github.com/boardgameio/boardgame.io/blob/dc6cd451d85b4cb8a857d21af06c66e68a217815/src/client/debug/log/Log.svelte#L14).
> 
> #### Implications
> * In the client we need to consider [the middleware added to the store](https://github.com/boardgameio/boardgame.io/blob/dc6cd451d85b4cb8a857d21af06c66e68a217815/src/client/client.ts#L219-L314) to see how exactly to handle the transient state and later surface `transients.error` when produced by the local reducer. As elsewhere, adding `transients` doen’t cause any actual errors, but presumably we’d want to strip it as in the master and use the error somehow.
> * 3 & 4 can perhaps wait — adding the `transients` field shouldn’t actually cause problems here. It may later be possible to use `transients.error` to improve the logic in AI simulations and log stepping (this latter could perhaps be done as part of #892)

The latest version has dealt with 1 & 2 as quoted above.

My implementation strategy of creating a new shared middleware between the client and master, however, suggests that a more fundamental unification of those code paths might be beneficial at some point. Having some kind of shared component that deals with all dispatches, and adding client/master specializations in an outer-most middleware might be easier to maintain.

Don't want to bite off more than I can chew though!

src/client/client.ts
```diff
@@ -85,7 +92,7 @@ function createDispatchers(
 ) {
   return innerActionNames.reduce((dispatchers, name) => {
     dispatchers[name] = function (...args: any[]) {
-      store.dispatch(
+      return store.dispatch(
```
delucis, line 95: Could you explain why you’re returning the result of `dispatch` here and elsewhere? `dispatch` [returns the action that was dispatched](https://redux.js.org/api/store#dispatchaction), which I guess is not the intended return value here.

src/core/reducer.test.ts
```diff
@@ -749,3 +756,36 @@ describe('undo / redo with stages', () => {
     expect(state.ctx.activePlayers['0']).toBe('B');
   });
 });
+
+describe('TransientHandlingMiddleware', () => {
```
delucis, line 760: Related to returning the result of `dispatch`, I don’t think these tests are testing what you expect. The `result` here is actually just the action object, so for example:

```js
const result = store.dispatch(makeMove('A'));
// result = {
//   type: 'MAKE_MOVE',
//   payload: {
//     type: 'A',
//     args: undefined,
//     playerID: undefined,
//     credentials: undefined,
//   },
// };
```

Because the tests assert absence of the transient keys, they pass because those keys are also not present in the action object.

Instead, you need to get the state from the store after dispatching:

```js
store.dispatch(makeMove('A'));
const result = store.getState();
```

src/core/reducer.ts
```diff
@@ -121,6 +128,77 @@ function initializeDeltalog(
...
+  };
+}
+
+/**
+ * Middleware for processing TransientState associated with the reducer
+ * returned by CreateGameReducer.
+ * This should pretty much be used everywhere you want realistic state
+ * transitions and error handling.
+ */
+export const TransientHandlingMiddleware = (store: Store) => (
```
delucis, line 177: Could you explain what this middleware is trying to do? As above, you probably don’t want to be using the return of `store.dispatch`, but even assuming `stripTransientsResult` was the new state and not the action, wouldn’t this create a state in the following shape?

```ts
type MiddlewareReturn = State & {
  stripTransientsResult: State;
  transients: TransientMetadata;
};
```

delucis (dismissed):
Hey @shaoster, here’s a new review. I’m a little confused about what the new middleware is trying to do 😅 

---

**Edit:** Did more reading, checked the code again, and I understand a little better. Here’s where my understanding is currently (correct me if I’m wrong):

- `TransientHandlingMiddleware` intercepts any `transients` the reducer produces, strips them from the store and then returns them at the `dispatch` call site alongside the original action object. Example return:

    ```js
    {
      type: 'MAKE_MOVE',
      payload: { ... },
      stripTransientsResult: { type: 'STRIP_TRANSIENTS' }, // invariant so perhaps superfluous?
      transients: { error: { type: 'action/invalid_move', payload: undefined } },
    }
    ```

Automatically stripping the `transients` seems like a good idea. I’m wondering if there’s a better way to surface the error than returning it though.

Would it make sense to instead follow a model more like that of the [`SubscriptionMiddleware` in the client](https://github.com/shaoster/boardgame.io/blob/475ef1096cb83ba9f35f20d62f8b3dd51fbdaeb0/src/client/client.ts#L304-L310)? I mean to say that basically a transient handler could strip transients and then trigger side effects, rather than returning the transients. Something like:

```js
const createErrorHandlingMiddleware =
  (onError: (error: ActionError) => void) =>
  (store: Store) =>
  (next: Dispatch<ActionShape.Any>) =>
  (action: ActionShape.Any) => {
    const result = next(action);
    const [, transients] = ExtractTransients(store.getState());
    if (typeof transients !== 'undefined') {
      store.dispatch(stripTransients());
      if (transients.error) onError(transients.error);
    }
    return result;
  };

// Usage
const middleware = applyMiddleware(
  createErrorHandlingMiddleware((error) => {
    // side effects
  })
);
```

I’m not sure if that’s quite what we want, but basically I think we’re missing one more step to wire up the transients in the master and the client so I can get my brain around where this is going, so I’m curious what you’re planning.

src/core/reducer.ts
delucis: (I’m also not sure it’s permitted to call `dispatch` from middleware. The `dispatch` docs say [it’s an error “to call `dispatch` from inside the reducer”](https://redux.js.org/api/store#dispatchaction), but I’m not clear if there’s a distinction for middleware. Given middleware is sometimes used to handle async actions, I’m guessing maybe dispatching from middleware is supported.)

src/client/client.ts
delucis: Huh, I’m now seeing in the client test that you are getting the `transients.error` out of the move dispatcher — I’ll wait for your explanation 😄

src/client/client.ts
```diff
@@ -42,6 +45,10 @@ type ClientAction =
   | ActionShape.Patch;
 type Action = CredentialedActionShape.Any | ClientAction;
 
+// TODO(#723): Figure out a nice API for the returned ActionResult.
```
shaoster, line 48: @delucis: This is basically the crux to the issues you raised:

1. TransientHandlingMiddleware is meant to be used (eventually) by all invokers of the middleware. I wanted it to have some common interface, which for the sake of discussion, let's call `ActionWithTransients`. It should have exactly the type that you described:
    ```typescript
    type MiddlewareReturn = ActionShape.Any & {
      stripTransientsResult?: ActionShape.StripTransients;
      transients?: TransientMetadata;
    };
    ```

2. At this point the handling on the client and master diverge somewhat but both can use this `ActionWithTransients` to deal with their concerns either with divergent middlewares sitting in front of the `TransientsHandlingMiddleware` or by directly wrapping the return value of the dispatch calls. For example, the master's handling can pretty much just use `createErrorHandlingMiddleware` slightly modified as follows:

    ```typescript
    const createResultHandlingMiddleware =
      (onSuccess: (actionResult: SomeGenericSuccessfulReturnType) => void) =>
      (onError: (error: ActionError) => void) =>
      (store: Store) =>
      (next: Dispatch<ActionShape.Any>) =>
      (action: ActionShape.Any) => {
        // Delegate to the stripping handler.
        const result = next(action);
        if (action.type instanceof VirtualActions /*i.e. Actions.STRIP_TRANSIENTS */) {
           // Trust the underlying middleware to handle these "virtual" actions.
           return result;
        }
        // At this point we should just handle remaining actions as we would.
        assert ! action.type instanceof VirtualActions;
        const { transients } = result;
        if (typeof transients !== 'undefined') {
          // Leave re-dispatch and result multi-plexing to the TransientHandlingMiddleware.
          if (transients.error) onError(transients.error);
        }
        return onSuccess(new SomeGenericSuccessfulReturnType(result));
      };
    ```

3. The only reason why I'm not super sure that pinning the above change and the typing of ActionResult here to the current change is ideal is because I think the most reasonable type would be a breaking change:

    ```typescript
    /* The promise will not resolve until the reducer has actually processed the action!!! */
    export type ActionResult = Promise<NiceGeneralResultBody>;
    export type NiceGeneralResultBody = any /* The actual data model of action results */;
    ```
 
Returning this promise with this semantic, for example requires all callers (even for external users of the API) to _have_ to `await` the result of their `move.Foo()` calls in order to get the correct behavior. I would probably want to work around this by leaving existing synchronous APIs that directly return `NiceGeneralResultBody`.

In my naive view, biting off this next piece requires a significant ballooning of scope in order to restore backwards compatibility. In the current state, the backwards-incompatibility is resolved with the argument that the return value isn't nice and user-consumable, but the previous return value was Void and there's at least a type indicator that we can fix up (and document) once we're ready to export it to users.

src/client/client.ts
delucis, line 95: Until we decide how to surface errors, how about we leave dispatchers unchanged as void functions?

```suggestion
      store.dispatch(
```

In my RFC I wondered about returning promises, but perhaps as a first step we should focus on the existing subscription/update system rather than introducing this new path for state updates.

This could be something like adding an error parameter to subscription callbacks, so where currently you do

```js
client.subscribe((state: State | null) => {});
```

we could implement errors so you do

```js
client.subscribe((state: State | null, error?: ActionError) => {});
```

Existing consumers would be unimpacted, but users could upgrade to get that error object.

src/core/reducer.ts
```diff
@@ -121,6 +128,77 @@ function initializeDeltalog(
...
+    }
+    default: {
+      const [, transients] = ExtractTransients(store.getState());
+      if (typeof transients !== 'undefined') {
+        const stripTransientsResult = store.dispatch(stripTransients());
+        // TODO(#723): Revisit this result payload and consider converting
+        // to a promise.
+        return {
+          ...rawResult,
+          stripTransientsResult,
```
delucis, line 193: Do we need to include `stripTransientsResult` here? What might it be used for?

src/core/reducer.ts
```diff
@@ -121,6 +128,77 @@ function initializeDeltalog(
...
+/**
+ * Middleware for processing TransientState associated with the reducer
+ * returned by CreateGameReducer.
+ * This should pretty much be used everywhere you want realistic state
+ * transitions and error handling.
+ */
+export const TransientHandlingMiddleware = (store: Store) => (
+  next: Dispatch<ActionShape.Any>
+) => (action: ActionShape.Any) => {
+  const rawResult = next(action);
```
delucis, line 180: Could this variable just be `result` at this point?

```suggestion
  const result = next(action);
```

src/core/reducer.ts
```diff
@@ -121,6 +128,77 @@ function initializeDeltalog(
   };
 }
 
+/**
+ * ExtractTransientsFromState
+ *
+ * Split out transients from the a TransientState
+ */
+function ExtractTransients(
+  transientState: TransientState
```
delucis, line 137: We should probably mark that this can be `null` in the signature:

```suggestion
  transientState: TransientState | null
```

(Our tsconfig doesn’t have strict null checking, so `null` and `undefined` are just ignored as possibilities by the compiler, but I try to be specific in newer code in the hope of one day being able to enable stricter typechecking.)

delucis (changes requested):
Thanks for the extra context — the piece I was missing was the eventual intention of how the additional `transients` would be consumed.

I’ve made a few suggestions to tidy up and then I guess the final thing needed for this PR would be to go through implementing the errors that haven’t yet been assigned an error code in the reducer, right?

src/core/reducer.ts
shaoster: I don't have a concrete use case yet, and it seems like this detail isn't very important, but here's my long-winded thinking behind this decision. While not necessarily pertinent to this change, it may provide an opportunity for you to tell me I'm thinking about this totally wrong and save a lot of time going forward:

 1. I'm basically thinking about this like a redux-friendly exception stack pattern where each middleware acts as a stack frame. (I just haven't implemented a stack yet, because there's only one virtual-action-spawning middleware for now, and creating a more general API for adding and consuming from a stack with only one data point probably be a waste of time.)

 2. Other middleware used alongside this one should generally make a decision about whether to specifically handle "virtual" actions (like the auto-generated STRIP_TRANSIENTS) or just pass them through untouched with a case statement on the action type. For these middlewares that don't care, it would be easier for them to delegate the multiplexing of the result values to the the middleware that spawns the "virtual" action (i.e. this one).

 3. For middlewares that might actually care (like loggers or bots), it might be useful, for visualization/debug/tree-pruning purposes, to know the provenance of the virtual action (i.e. which "real" action caused the virtual action to be spawned). In the current setup, there's only a single virtual action so such a logger middleware could simply maintain some state and look back at the previous action received, but this is messy and won't generalize if we expand on this "virtual" action pattern.
    Imagine if we introduce a second action-spawning middleware in our stack (e.g. to automatically spawn master->client synchronization actions that might allow us to subsequently unify the client and master logic).
   Then (absent some stack-based result-multiplexing) the implementation of a provenance-tracking logger would have to depend on the exact stack-order of the virtual-action-spawning middlewares... plausible, but nasty!

 4. Because the spawned "virtual" action is also exposed to the other middle-wares such that it could be modified/filtered on its way to and from the reducer, it is useful to know what the actual result value is at the handleTransients layer. If another middleware is added before this one (as we've discussed), 

 5. As for why I'm adding the "virtual" result relation to the result of the "real" action and not the other way around (i.e. adding the spawning action to the virtual action result), this is just because of the stack-resolution order of the actions. While the incoming dispatch order is real action -> virtual action, the dispatch return order is virtual action -> real action. Thus, if something fails in the middleware handling of the virtual action, useful forensic information could be lost.

src/client/client.ts
shaoster: Seems good to me. Only tactical concern is that, without making some decisions in this PR on how to expose the error parameters to the subscription callbacks, testability becomes a bit tricky.

delucis: I think testing it only where currently relevant is OK:

1. Test the reducer in isolation.
2. Test the store with the transient handling middleware in isolation.
3. Test the client, but only to make sure transients don’t end up in state, something like

    ```js
    client.moves.invalid();
    expect(client.getState().transients).toBeUndefined();
    ```

4. Test the master, but for now only to ensure transients don’t get sent in the update payload.

Once the master is emitting an error event and the client is doing whatever it ends up doing, we can then test those.

src/core/reducer.ts
delucis: Hmm, OK. To the extent that I understand this, I still think this could be a case for [“YAGNI”](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). For now, the action shape will be `{ type, payload }` or `{ type, payload } & { transients, stripTransientsResult }`, i.e. presence of `transients` indicates transients were stripped and the actual value of `stripTransientsResult` is always the same so it doesn’t actually carry any significant information. Another middleware could achieve all the same functionality with just `{ type, payload, transients? }`.

Happy for you to explore stacks etc. at some point, but for now I’d keep this as simple as possible for what we need. Augmenting the situation later should be a simple change.

shaoster: image

delucis: Cool! I guess all I’m saying is let’s avoid adding data that is not actually consumed yet. If a future PR needs the `stripTransientsResult` property, let’s add it then.

shaoster: Removed the payload, updated the master.test, and added a Dev Note with a link to this thread, since it's a bit trickier than it looks and I think this discussion/diagram would be useful to anybody tempted to muck with this code in the future..

src/client/client.ts
shaoster: Sweet. Set the ActionResult type alias back to `void`, reverted the added `return`s, and modified the client test to only check that no transients remained (with actionable TODOs to add the appropriate assertions in the subsequent PRs for #723).

shaoster:
Review feedback addressed. PTAL

src/core/reducer.ts
```diff
@@ -294,12 +385,15 @@ export function CreateGameReducer({
...
-          return state;
+          return WithError(state, ActionErrorType.ActionDisabled);
         }
 
         const { _undo, _redo } = state;
 
         if (_undo.length < 2) {
+          // TODO(#723): Add an error case here.
+          //  error(`No moves to undo`);
+          //  return WithError(state, ActionErrorType.ActionDisabled);
```
delucis, line 396: Do you need a hand deciding on names for the various commented out errors here and below or do you want to have a go at creating error types for these? Or are you reserving this for a separate PR?

src/core/reducer.ts
```diff
@@ -121,6 +128,79 @@ function initializeDeltalog(
...
 }
 
+/**
+ * ExtractTransientsFromState
+ *
+ * Split out transients from the a TransientState
+ */
+function ExtractTransients(
+  transientState: TransientState | null
+): [State | null, TransientMetadata | null] {
```
delucis, line 138: Missed this first time round, but the actual return type here is:

```suggestion
): [State | null, TransientMetadata | undefined] {
```

delucis (commented):
Ok, I think we’re getting there! One small fix and a question.

src/core/reducer.ts
shaoster: I'm fine with either:
1. Taking some help to decide on error types and updating the tests as part of this change.
2. Just leaving this as a TODO to the next PR that starts adding semantics.

delucis: I think it would be nice to include these as part of this PR so we have the reducer code basically done. I’ll take a look and make some suggestions.

delucis: OK, so the cases to cover are:

- No moves to undo
- Cannot undo other players' moves
- Move cannot be undone
- No moves to redo
- Cannot redo other players' moves

Do you think collapsing these all into a single error type would be OK?

- `action/action_invalid` — The requested action is not currently possible

It would make handling each specific case separately, but keeping the total number of error types down also seems desirable.

shaoster: So when trying to update the master tests after raising these errors, I found several tests that don't do what they claim to be doing and, in general, that file looks like it needs to be reworked. There's a ton of dependence between different test cases AND different test suites, and they are extremely order dependent.

It's also super unclear what, if any, coverage the seemingly duplicated tests are adding compared to a parameterized test on the game.deltaState field.

Thus, I'm going to do the minimal thing to get these tests passing and covering the updated behavior.

shaoster: Should I raise a bug for ^ and fix that first?

delucis: Good catch! A fix PR would be ideal (no need to open an issue for something like this).

shaoster: Opened https://github.com/boardgameio/boardgame.io/pull/942 to fix this underlying issue.

delucis:
@shaoster I’ve merged your fix PR. Let me know if you need a hand resolving the conflict with this branch.

shaoster:
De-conflicted against #942 and local change for the additional error cases rebased on top.

delucis (approved):
Fantastic work! Particularly isolating all those test cases instead of having them depend on test order. I think this looks good to go 🎉

PR #946, opened 2021-05-28
Upgrading socket.io packages

### LINKED ISSUE #945: Upgrading socket.io and koa-socket-2
Socket.io is now in major version 4. Any reason to still be using version 2 here? koa-socket-2 also just released version 2.0. If interested I could give it a go at a PR and running the tests etc.

chasen:
Socket.io latest version comes with some quality of life improvement that will help with the game development I am working on. So that we dont have to run a separate socket.io server and we could instead just hook into the main one used by the game it would be nice to be able to use these features.

All tests run and pass

Closes #945 

#### Checklist

- [X] Use a separate branch in your local repo (not `main`).
- [X] Test coverage is 100% (or you have a story for why it's ok).


delucis (changes requested):
Thanks for this @chasen, this was on our to-do list. A few comments:

- We don’t actually import `socket.io` directly anywhere, this is done by `koa-socket-2`, so we can remove `socket.io` as a direct dependency.

- `koa-socket-2` still uses socket.io v3 ([source](https://github.com/ambelovsky/koa-socket-2/blob/4033606da0532dd5dd0a5d3970b1d871a8a994a2/package.json#L34)), so we’d be using a v3 server with a v4 client. According to [the socket.io v3 → v4 migration guide](https://socket.io/docs/v4/migrating-from-3-x-to-4-0/):

    > a v3 client will be able to reach a v4 server and vice-versa

    So this is probably OK, but maybe worth double checking?

- As of socket.io v3, type definitions are bundled with the main packages ([source](https://socket.io/docs/v4/migrating-from-2-x-to-3-0/#The-Socket-IO-codebase-has-been-rewritten-to-TypeScript)), so we should remove the `@types/socket.io` and `@types/socket.io-client` packages. This does trigger some TypeScript errors though, which will need resolving.

- Most of the relevant breaking changes to socket.io ([v3](https://github.com/socketio/socket.io/blob/master/CHANGELOG.md#300-2020-11-05), [v4](https://github.com/socketio/socket.io/blob/master/CHANGELOG.md#400-2021-03-10)) don’t seem to impact us but we do need to make some changes to handle [how socket.io >3 handles CORS](https://socket.io/docs/v3/migrating-from-2-x-to-3-0/#CORS-handling). If I run our examples (you can do this by running `npm start` from the boardgame.io repo), all the multiplayer examples (except Tic-Tac-Toe > Multiplayer, which uses the `Local` multiplayer adapter rather than the socket server) are broken on this branch and print CORS errors to the browser console. Ideally we would provide a default set-up that continues working as currently, or give clear guidance on how to configure CORS and apply that guidance to these examples.

---

(As you can see, the tests can’t be entirely trusted when it comes to socket communications as in general one side or the other gets mocked in tests, so unfortunately a bit of manual testing is necessary until that’s improved.)

chasen:
Sounds good, ill take a stab at updating these over the next few days

vdfdev:
@delucis PTAL... I think I addressed all your comments. Important to note a new "origins" param for the Server

delucis:
Thanks for taking this on @vdfdev! This all looks good and the examples are working as expected. One thought: do you think we should throw an error if the `origins` option is not set? Or log some kind of warning? It is possible to serve the front-end from the same origin as the game server, but I’m guessing the vast majority don’t (as in all our documentation apart from the Heroku deployment guide). Maybe a warning when starting the server would be helpful to people debugging after updating.

delucis (approved):

larry801:
CVE-2020-36048
Please test this vuln resolved or not?
https://github.com/socketio/engine.io/commit/734f9d1268840722c41219e69eb58318e0b2ac6b
https://blog.caller.xyz/socketio-engineio-dos/

delucis:
@larry801 The updates on this branch will bump the used engine.io version to 4.1.1 and that CVE lists the vulnerable versions as <4.0.0, so I guess merging this would resolve that?

delucis:
@vdfdev I noticed that the `origins` can also be regular expressions, so I’ve updated the docs to mention that. (Allows us to cover any localhost ports in the tutorial, which is nice as Parcel serves on `1234` by default and Create React App on `3000`.) I’ve also added the logging I suggested to help point people to the docs if they update and encounter errors.

vdfdev:
Hum, Nice, I didn't know it could be a regex. Maybe we can leave the local host with any port as the default? Thanks for adding the warning!

vdfdev (dismissed):

delucis:
> Maybe we can leave the local host with any port as the default?

I also considered that, but I guess the socket.io CORS default was chosen with security in mind, and requiring setting it gives us the chance to introduce the concept in the tutorial (instead of encountering it for the first time as yet another thing that breaks when deploying to a proper server).

vdfdev:
I agree with this trade-off... sounds good then!

delucis (approved):

delucis:
Thanks for getting the ball rolling @chasen and for getting things working @vdfdev! This is now released as 0.45.0 🎉

PR #955, opened 2021-07-13
feat(server): Use `origins` option to configure Lobby API CORS

delucis:
#946 added an `origins` option to the server, which is used to enable CORS for the socket.io transport.

The Lobby API server currently uses `@koa/cors` with no configuration to reflect the request origin back in the `Access-Control-Allow-Origin` header, enabling requests from any origin:

https://github.com/boardgameio/boardgame.io/blob/2b1d013a03132d9d40c38c0d34b94e91d0a2509a/src/server/api.ts#L451

Given the server now *requires* setting origins, I thought we should probably also lock down the Lobby REST API to those same origins. This PR adopts the same method used by socket.io to validate request origins against the `origins` option for use with the `@koa/cors` middleware.

What do you reckon @vdfdev? Is there a chance the socket server and REST API need to allow access to different origins, which would require a second `apiOrigins` option?

### DIFF
--- a/docs/documentation/api/Server.md
+++ b/docs/documentation/api/Server.md
@@ -56,6 +56,9 @@ A config object with the following options:
 
 7. `authenticateCredentials` (_function_): an optional function that tests if a player’s move is made with the correct credentials when using the default socket.io transport implementation.
 
+8. `apiOrigins` (_array_): a list of allowed origins for requests to the Lobby API. Defaults
+   to the value provided as the `origins` option (which also applies to the socket transport).
+
 #### Returns
 
 An object that contains:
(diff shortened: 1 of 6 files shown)


vdfdev:
Oh, good catch, I forgot about the API server! I think having the default of both being the same is great, but also think we should allow overwriting the API server cors. In the case of FBG, we don't want to allow anybody to access our API server but our own server, so this option would be good to restrict that

delucis:
OK, added that option, setting it to `origins` by default. PTAL.

vdfdev (approved):

vdfdev:
Looks great, thank you!

PR #963, opened 2021-07-19
feat: Allow plugins to declare an action invalid

### LINKED ISSUE #962: Add API allowing plugins to declare the current action invalid
Currently we support returning `INVALID_MOVE` from moves to bail out of a state update. There are circumstances where a plugin might also want to flag that an action is invalid. For example, a plugin that validates state or — as came up in https://github.com/boardgameio/boardgame.io/pull/957#pullrequestreview-709062516 — in the events plugin if there is an error processing the event queue.

This would probably look something like the current `noClient` plugin method. A plugin could declare an `isInvalid` method that could return any truthy value to invalidate the current action.

```js
const plugin = {
  name: 'validator',
  isInvalid: ({ G }) => {
    if (typeof G.value !== number) {
      return { error: 'G.value must be a number' };
    }
  },
};
```

Once we have something like #723, plugin errors could be added to the possible error types and the return value of `isInvalid` returned to the client for debugging.

delucis:
Closes #962

This adds support for an `isInvalid` method in boardgame.io plugins. Similar to the current `noClient` plugin method, this allows the reducer to ask plugins if everything is OK or if the current event/move is invalid from their point of view. If one or more plugins returns a truthy value from its `isInvalid` method, the reducer will fall back to its initial state, analogous to what it does when a move returns `INVALID_MOVE`.

Currently this is not in use but the intention would be to use it for the events plugin, following on from #957. It also allows for other custom use cases, like state validators.

#### Notes

- `isInvalid` should return a string if the action is invalid. This could be less strict if that seems desirable.

- The error returned to the client is of the form
    ```js
    {
      type: 'action/plugin_invalid',
      payload: {
        plugin: 'name of plugin that declared action invalid',
        message: 'string returned by plugin’s isInvalid method'
      },
    }
    ```

- If more than one plugin’s `isInvalid` method returns a string, we just error with the first of them, rather than reporting all of them.

### DIFF
--- a/src/core/reducer.ts
+++ b/src/core/reducer.ts
@@ -129,6 +129,32 @@ function initializeDeltalog(
 }
 
 /**
+ * Update plugin state after move/event & check if plugins consider the action to be valid.
+ * @param newState Latest version of state in the reducer.
+ * @param oldState Initial value of state when reducer started its work.
+ * @param pluginOpts Plugin configuration options.
+ * @returns Tuple of the new state updated after flushing plugins and the old
+ * state augmented with an error if a plugin declared the action invalid.
+ */
+function flushAndValidatePlugins(
+  newState: State,
+  oldState: State,
+  pluginOpts: { game: Game; isClient?: boolean }
+): [State, TransientState?] {
+  newState = plugins.Flush(newState, pluginOpts);
+  const isInvalid = plugins.IsInvalid(newState, pluginOpts);
+  if (isInvalid) {
+    const { plugin, message } = isInvalid;
+    error(`plugin declared action invalid: ${plugin} - ${message}`);
+    return [
+      newState,
+      WithError(oldState, ActionErrorType.PluginActionInvalid, isInvalid),
+    ];
+  }
+  return [newState];
+}
+
+/**
  * ExtractTransientsFromState
  *
  * Split out transients from the a TransientState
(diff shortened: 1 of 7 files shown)


src/core/reducer.ts
```diff
@@ -128,6 +128,32 @@ function initializeDeltalog(
...
+ * state augmented with an error if a plugin declared the action invalid.
+ */
+function flushAndValidatePlugins(
+  newState: State,
+  oldState: State,
+  pluginOpts: { game: Game; isClient?: boolean }
+): [State, TransientState?] {
+  newState = plugins.Flush(newState, pluginOpts);
+  const isInvalid = plugins.IsInvalid(newState, pluginOpts);
+  if (isInvalid) {
```
vdfdev, line 146: NIT to decrease nesting, you can do if (!isInvalid) { return [newState]; } as a guard clause https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

src/plugins/main.test.ts
```diff
@@ -8,6 +8,7 @@
 
 import { Client } from '../client/client';
 import { Local } from '../client/transport/local';
+import type { Ctx, Game } from '../types';
```
vdfdev, line 11: Ctx defined but never used

src/plugins/main.test.ts
```diff
@@ -134,6 +135,68 @@ describe('default values', () => {
...
+  test('basic plugin', () => {
+    const game: Game = {
+      plugins: [
+        {
+          name: 'test',
+          isInvalid: ({ G }) => 'bad' in G && 'not ok',
+        },
+      ],
+      moves: {
+        good: () => ({ good: 'nice' }),
```
vdfdev, line 153: NIT move the {good: 'nice'} to a constant so we you dont need to repeat it when checking later?

vdfdev (dismissed):
10/10 very readable and well tested code, thank you!

delucis:
Thanks for the review @vdfdev!

PR #966, opened 2021-07-26
Refactor: Decouples player view calculation from Server's Master

vdfdev:
In this PR I refactor the interface between Master and Transport, so the Master will only send a single message to the transport, and then the transport will calculate the player view for each player. 

This will allow us to use a pub/sub to propagate the Master message, allowing us to scale the bgio server horizontally.

#### Checklist

### DIFF
--- a/src/master/master.ts
+++ b/src/master/master.ts
@@ -31,7 +31,7 @@ import { createMatch } from '../server/util';
 import type { Auth } from '../server/auth';
 import * as StorageAPI from '../server/db/base';
 import type { Operation } from 'rfc6902';
-import { createPatch } from 'rfc6902';
+import { redactLog } from './filter-player-view';
 
 /**
  * Filter match data to get a player metadata object with credentials stripped.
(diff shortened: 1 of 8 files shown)


src/server/transport/socketio.ts
```diff
@@ -70,9 +72,9 @@ export function TransportAPI(
   /**
    * Send a message to all clients.
    */
-  const sendAll: MasterTransport['sendAll'] = (makePlayerData) => {
```
delucis, line 73: We also need to refactor the `send` method. It is currently only used by the `sync` action, but that is still filtering in the Master: https://github.com/vdfdev/boardgame.io/blob/38a62267ce23508f82596557e3020b78ab4e0c36/src/master/master.ts#L371-L395

(Not sure if the `send` method is such an issue for Pub/Sub implementation, but it’s probably a good idea to refactor both together for consistency nonetheless.)

src/master/master.ts
```diff
@@ -114,9 +78,31 @@ export type TransportData =
       args: [string, ChatMessage];
     };
 
+export type IntermediateTransportData =
```
delucis, line 81: There’s quite a bit of duplication between this and the `TransportData` type. Could we extract the types shared between both?

```ts
type CommonTransportData =  { /* ... */ } | { /* ... */ };
type TransportData = CommonTransportData | { /* ... */ };
type IntermediateTransportData = CommonTransportData | { /* ... */ };
```

We could even break out the related data types if you like:

```ts
import type { List } from 'ts-toolbelt';

// ...

type IntermediateUpdateData = {
  type: 'update';
  args: [string, State];
};

type UpdateData = {
  type: IntermediateUpdateData['type'];
  args: List.Append<IntermediateUpdateData['args'], LogEntry[]>;
};
```

(With Typescript 4 we wouldn’t need the `ts-toolbelt` helper and could write `args: [...IntermediateUpdateData['args'], LogEntry[]]`, but I don’t remember if  updating caused problems with Rollup.)

delucis (changes requested):
This looks like a good way to untangle these — good work!

We also need to account for the sync action, but otherwise this looks ready to go.

src/server/transport/socketio.ts
vdfdev: Done, good catch. I didnt notice this before because I was only concerned with the `sendAll` (`send` should not need pub sub as we dont need to propagate to multiple servers), but It looks much cleaner now that `Master` only cares about the `IntermediateTransportData` and the Transport cares about the `TransportData`

src/master/master.ts
vdfdev: I went with your first suggestion, I am not very familiar with the ts-toolbelt and I felt like it could make things harder to read (at cost of some repetition).

vdfdev (commented):
Comments addressed

delucis (approved):

PR #972, opened 2021-08-03
feat(phases): Support setting next phase with a function

### LINKED ISSUE #954: Feature Request - Allow to specify next phase using a function
**Problem:** Currently the `next` field in the phase configuration only accepts a string.

 I'm trying to write a phase using the `TurnOrder.ONCE` turn order, however the next phase needs to be determined dynamically. The only way to do this as far as I know is to specify an `endIf` function, but this clashes with automatic phase-ending `TurnOrder.ONCE`. I could write such a function, but especially since `endIf` is called before the phase is even begun, in my case the logic to accomplish this gets unnecesarily convoluted. It would be convenient if the `next` field accepted a function to allow to determine the next phase dynamically, without interfering with phase-ending logic.

jorbascrumps:
Closes #954 

I don't typically write Typescript nor have I contributed to this project in many moons so if I've missed something please let me know!

#### Checklist


src/core/flow.ts
```diff
@@ -163,6 +163,10 @@ export function Flow({
       onEnd: HookWrapper(phaseConfig.turn.onEnd),
       endIf: TriggerWrapper(phaseConfig.turn.endIf),
     };
+
+    if (typeof phaseConfig.next === 'function') {
+      phaseConfig.wrapped.next = HookWrapper(phaseConfig.next);
+    }
```
delucis, line 169: Could we always wrap `next` so that later code doesn’t need to check its type?

Something like:

```suggestion
if (typeof phaseConfig.next !== 'function') {
  const { next } = phaseConfig;
  phaseConfig.next = () => next;
}
phaseConfig.wrapped.next = HookWrapper(phaseConfig.next);
```

src/types.ts
```diff
@@ -205,7 +205,7 @@ export interface PhaseConfig<
   CtxWithPlugins extends Ctx = Ctx
 > {
   start?: boolean;
-  next?: string;
+  next?: ((G: G, ctx: CtxWithPlugins) => any) | string;
```
delucis, line 210: This is good, but we can be more precise like so:

```suggestion
  next?: ((G: G, ctx: CtxWithPlugins) => string | void) | string;
```

This says that a `next` method should return a string or nothing (`void` is basically like `undefined`) so if people try `next: () => 5` or whatever, they’ll see errors.

src/types.ts
```diff
@@ -217,6 +217,7 @@ export interface PhaseConfig<
     ) => boolean | void | { next: string };
     onBegin?: (state: State<G, CtxWithPlugins>) => any;
     onEnd?: (state: State<G, CtxWithPlugins>) => any;
+    next?: (state: State<G, CtxWithPlugins>) => any;
```
delucis, line 222: Same here, we can specify the return type of `next`:

```suggestion
    next?: (state: State<G, CtxWithPlugins>) => string | void;
```

delucis (changes requested):
Nice work! Looks good, I’ve just added a couple of comments.

src/types.ts
jorbascrumps: I get the thinking and I agree - however, this causes failures elsewhere because the `Ctx` interface expects `phase` to be of type `string`. I've tried updating the expected type to `string | void` with no luck :(

Any guidance would be appreciated.

src/core/flow.ts
jorbascrumps: Great approach. Wish I'd thought of it!

src/types.ts
delucis: Oh, right, but that’s actually good because it’s telling us there’s a bug. `phase` should be `null` when there is no phase, but we’d be setting it to `undefined`. If we set phase to `phaseConfig.wrapped.next(state) || null` this type change will work.

src/core/flow.ts
```diff
@@ -315,7 +319,10 @@ export function Flow({
         return state;
       }
     } else if (phaseConfig.next !== undefined) {
```
delucis, line 321: Re: setting `phase` to `null`, I think this will always be `true`, so we can probably collapse everything into the else case.

src/types.ts
jorbascrumps: Ah, makes sense. I had to also update default wrapped function to match.

src/core/flow.ts
jorbascrumps: 💯

delucis (approved):

delucis:
@jorbascrumps Nice one! Thanks.

PR #977, opened 2021-08-07
fix(master): Don’t crash on missing chatMessage

nopeless:
This line was throwing unhandled promise rejections when I was testing invalid packets

#### Checklist

- [ this was edited in github] Use a separate branch in your local repo (not `main`).
- [no ] Test coverage is 100% (or you have a story for why it's ok)

### DIFF
--- a/src/master/master.test.ts
+++ b/src/master/master.test.ts
@@ -966,6 +966,19 @@ describe('authentication', () => {
 
     beforeEach(resetTestEnvironment);
 
+    test('auth success', async () => {
+      const authenticateCredentials = () => true;
+      const master = new Master(
+        game,
+        storage,
+        TransportAPI(send, sendAll),
+        new Auth({ authenticateCredentials })
+      );
+      const ret = await master.onChatMessage(matchID, chatMessage, undefined);
+      expect(ret).toBeUndefined();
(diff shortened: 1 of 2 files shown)


delucis (commented):
Looks like there are some linter errors. Try running `npm run prettier` to fix those.

nopeless:
@delucis yeah I figured it out, thanks. I was thinking about this the other night about having a schema validation like validate.js. But then I realized this chat packet is something different from the other events (for example it is not handled by redux). So I guess this simple fix will suffice

nopeless:
wait, let me fix a few other stuff

delucis (changes requested):
Ah, right. You’re using the optional chaining syntax `?.` but our setup doesn’t support that yet, so tests are still throwing errors.

Rather than adding this check to `isAuthentic`, how about we just add some simple validation at the start of the `onChatMessage` method? Something like

```js
if (!chatMessage) {
  return { error: 'missing chatMessage' };
}
```

(I think that’s enough, because the authentication method should account for an undefined or null player ID.)

nopeless:
Actually while fixing some other stuff, I realized that the other stuff should be a separate PR. I pinged you on gitter.

delucis (approved):
Thanks @nopeless!

PR #978, opened 2021-08-07
Adds pub-sub support for horizontally scaling bgio server

vdfdev:
Usage:
```
import { Server, Origins, SocketIO, RedisPubSub } from 'boardgame.io/server';
import TicTacToe from './src/tic-tac-toe/game';
import Chess from './src/chess/game';
import redis from 'redis';
import { PostgresStore } from 'bgio-postgres';
 
const PORT = process.env.PORT || 8000;
const redisConfig = { host: '192.168.8.50' };
const pubClient = redis.createClient(redisConfig);
const subClient = redis.createClient(redisConfig);
const server = Server({
  games: [TicTacToe, Chess],
  db: new PostgresStore('vdf@192.168.8.50/vdf'),
  origins: [Origins.LOCALHOST],
  transport: new SocketIO({
    pubSub: new RedisPubSub(pubClient, subClient),
  }),
});
server.run(PORT, () => {
  console.log(`Serving at: http://localhost:${PORT}`);
});
```

Demo:
https://youtu.be/4BneuOrfs-0 

#### Checklist


src/server/transport/pubsub/generic-pub-sub.ts
```diff
@@ -0,0 +1,23 @@
+import type { Observable } from 'rxjs';
+
+/** Generic interface for pub-subs.  */
+export interface GenericPubSub<T> {
+  // Publish an event for a match.
+  publish(channelId: PubSubChannelId, payload: T);
+
+  // Subscribes to events related to a single match.
+  subscribe(channelId: PubSubChannelId): Observable<T>;
```
delucis, line 9: I’m not 100% sure about using `rxjs` for this. I’ve had a look at their docs and it seems to add quite a bit of conceptual overhead for functionality most of which we don’t need. Insisting on using it here might make contributing additional pubsub adapters more difficult than necessary for people not familiar with `rxjs`

At the very least, can we make the use of `rxjs` an implementation detail here?

We only call `subscribe` in one place and it is used as `subscribe(id).subscribe(callback)`, so it seems we could make this more flexible, by just defining:

```suggestion
  subscribe(channelId: PubSubChannelId, callback: (payload: T) => void): void;
```

A pubsub implementation is then free to use RxJS internally if it likes, but it isn’t a requirement. (I’d lean towards removing at least for the in-memory implementation to avoid adding dependencies, but we can do whatever with the separate redis package.)

src/server/transport/pubsub/redis-pub-sub.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export class RedisPubSub<T> implements GenericPubSub<T> {
+  private pubClient: redis.RedisClient;
+  private subClient: redis.RedisClient;
+  private subscriptions: Map<string, Observable<T>> = new Map();
+
+  constructor(pubClient: redis.redisclient, subClient: redis.redisclient) {
+    this.pubClient = pubClient;
+    this.subClient = subClient;
+  }
```
delucis, line 14: As discussed on Discord, it seems like it would be helpful to allow this to receive a config object and then create clients automatically. Based off your usage example:

```suggestion
  constructor({
    config,
    pubConfig = config,
    subConfig = config,
    pubClient = redis.createClient(pubConfig),
    subClient = redis.createClient(subConfig),
  }) {
    this.pubClient = pubClient;
    this.subClient = subClient;
  }
```

src/server/transport/socketio.ts
```diff
@@ -17,76 +17,52 @@ import type {
 } from '../../master/master';
 import { getFilterPlayerView } from '../../master/filter-player-view';
 import type { Game, Server } from '../../types';
+import type { GenericPubSub } from './pubsub/generic-pub-sub';
+import { PubSubChannelIdNamespace } from './pubsub/generic-pub-sub';
+import type { IntermediateTransportData } from '../../master/master';
+import { InMemoryPubSub } from './pubsub/in-memory-pub-sub';
+import type { PubSubChannelId } from './pubsub/generic-pub-sub';
```
delucis, line 24: Combine type imports from the same file.

```suggestion
import { PubSubChannelIdNamespace } from './pubsub/generic-pub-sub';
import type { IntermediateTransportData } from '../../master/master';
import { InMemoryPubSub } from './pubsub/in-memory-pub-sub';
import type { GenericPubSub, PubSubChannelId } from './pubsub/generic-pub-sub';
```

delucis (commented):
@vdfdev Just added a few comments. The key one is about how we define the generic pubsub interface.

src/server/transport/pubsub/util.ts
```diff
@@ -0,0 +1,5 @@
+import type { PubSubChannelId } from './generic-pub-sub';
+
+export function globalChannelId(channelId: PubSubChannelId): string {
+  return `${channelId.namespace}-${channelId.value}`;
+}
```
delucis, line 5: One more thought regarding the public interface. Both implementations are calling this `globalChannelId` helper everywhere they use the channel ID (often multiple times per method). Could the public interface methods just receive a `string` argument instead of `PubSubChannelId` and boardgame.io can take care of calling `globalChannelId`?

src/server/transport/socketio.ts
```diff
@@ -17,76 +17,52 @@ import type {
...
 const PING_INTERVAL = 10 * 1e3;
 
+const emit = (socket: IOTypes.Socket, { type, args }: TransportData) => {
+  socket.emit(type, ...args);
+};
+
+function getPubSubChannelId(matchID: string): PubSubChannelId {
+  return { namespace: PubSubChannelIdNamespace.MATCH, value: matchID };
+}
+
```
delucis, line 36: In fact, I think this is the only place that creates `PubSubChannelId` objects, right? And then passes them to the pub sub implementation. So this method could just do the string concatenation and we wouldn’t need the `globalChannelId` helper at all.
```suggestion
function getPubSubChannelId(matchID: string): string {
  return PubSubChannelIdNamespace.MATCH + '-' + matchID;
}

```

src/server/transport/pubsub/generic-pub-sub.ts
delucis: (On a side note regarding dependencies, we don’t currently have `rxjs` in our dependencies, so that fact this works is probably just by chance because some other package is pulling it in.)

src/server/transport/pubsub/util.ts
vdfdev: I thought that namespacing the IDs would make the API more extensible in the future, avoiding id collisions and etc. But sure, replaced with string.

src/server/transport/socketio.ts
vdfdev: Right, this had the advantaged of giving type safety and guaranteeing that the id wouldnt collide if they were used for other purposes beyond the match (maybe chat outside of a match in the future...). Generally over-using elementary types like string and etc to represent higher level concepts isnt advisable. But I followed your recommendation here.

src/server/transport/pubsub/generic-pub-sub.ts
vdfdev: It is very easy to get to callback hell using them, and make code that is very hard to understand. We cant use promises here because there will be multiple . RxJS use case is exactly this (streams) and gives powerful functional operations like map, filter, etc... But sure, followed your advice.

src/server/transport/pubsub/redis-pub-sub.ts
vdfdev: This class is going to be removed from this PR, and moved to its own repo.

src/server/transport/socketio.ts
vdfdev: Done

delucis (dismissed):
Thanks @vdfdev — I hope you didn’t mind making those changes. Open to discuss them further if you think we’d be better off keeping the namespace object etc. I don’t have a complete perspective of what happens on the other end of the pubsub mechanism (in Redis, etc.) so maybe it would be better to keep that namespace/matchID object if some implementations would benefit from using the namespace directly rather than already receiving the formatted channel ID.

I think this basically looks good to go. I’ve just added a tiny nitpick, but otherwise feel free to merge!

vdfdev:
Don't worry about it, I didn't mind. Good to keep my over-engineering in check sometimes hahaha

src/server/transport/socketio.ts
```diff
@@ -152,6 +130,23 @@ export class SocketIO {
...
+    this.pubSub.subscribe(
+      getPubSubChannelId(matchID),
+      (payload: IntermediateTransportData) => {
+        this.roomInfo.get(matchID).forEach((clientID) => {
+          const client = this.clientInfo.get(clientID);
+          const data = getFilterPlayerView(game)(client.playerID, payload);
+          emit(client.socket, data);
+        });
+      }
+    );
```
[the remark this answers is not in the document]
vdfdev: Done

vdfdev:
All checks have passed, I lost your approval but because I only did what you asked I will merge anyway

PR #980, opened 2021-08-12
feat: Expose method types to plugin `fnWrap` & improve events errors

### LINKED ISSUE #979: Infinite loop detector not detecting infinite loop
I was trying to upgrade FBG to the latest BGIO version, but the secret dracula bgio unit tests got into an infinite loop on this line:

https://github.com/freeboardgames/FreeBoardGames.org/blob/4ab6e45fb47849be212ce771b08a025a1d31b182/web/src/games/secretDracula/game.test.ts#L30

which is literally the first move the unit test executes. I got into the debugger and I found out that the code was looping through the check to detect infinite loops, but failing to detect it:

image

`endedTurns` would always show up as 1, even though this was looping thousands of times.  Interesting too, I saw the `this.dispatch` to grow unbounded:

image

I started suspecting of this recent PR: https://github.com/boardgameio/boardgame.io/pull/957, so I downgraded to one release before it (v0.45.2) and the same test passes :).

Also, I spent several hours this weekend trying to debug why the mergers unit test was failing. In that case, it was failing in this line:

https://github.com/freeboardgames/FreeBoardGames.org/blob/4ab6e45fb47849be212ce771b08a025a1d31b182/web/src/games/mergers/game.test.ts#L1120

It seemed like this BGIO framework was ignoring this command to move to the correct stage:
https://github.com/freeboardgames/FreeBoardGames.org/blob/4ab6e45fb47849be212ce771b08a025a1d31b182/web/src/games/mergers/game.ts#L523

And interesting enough, at v0.45.2 it also passes! And at v0.46.0 it fails! I am not very familiar with the turn order/flow code to understand what is going on, and it is getting late here and I need to go to bed... But  tomorrow I will try to see if I can nail down the root cause. It seems very likely, however, that the PR #957 broke some code that it shouldn't.

### LINKED ISSUE #981: `setStage` in `turn.onBegin` always uses playerID 0
As [reported on Gitter](https://gitter.im/boardgame-io/General?at=61154f7309a1c2738272d21e), calling `setStage` from `turn.onBegin` always places player `'0'` in the stage, not the player whose turn is starting.

Reproduction: https://codesandbox.io/s/boardgame-io-setstage-bug-reproduction-vcd40?file=/src/Game.ts

Initially I thought this was another bug introduced in the recent work on events, but after testing with a few previous versions, it seems it is a more long-standing bug. The `setActivePlayers` event works as expected.

delucis:
Here’s an attempt to close #979. Feedback very much wanted and if you want to pick this up and iterate on it, you’re welcome to as I likely won’t have much time in the next couple of weeks.

Also closes #981.

#### Approach

When wrapping game hooks (`onBegin`, `onMove`, `onEnd`), provide a unique string to the plugin interface so that it knows which hook is being wrapped. (These unique strings are from a kind of custom enum accessed like `GameMethod.Phase.ON_BEGIN` or `GameMethod.MOVE`.)

A plugin’s `fnWrap` then receives this string in addition to the function to wrap, so it can change behaviour appropriately.

```js
const plugin = {
  name: 'discerning-wrapper',
  fnWrap: (fn, fnType) => (...args) => {
    if (fnType === GameMethod.Turn.ON_MOVE) {
      console.log('Looks like this is an onMove hook.);
    }
    return fn(...args);
  },
};
```

In the events plugin we store this string, like we do with the turn/phase details and use it when processing events. This is actually pretty nice because we can give more informative errors. The main drawback is that this depends on `fnWrap` which is *not* applied to triggers (like `endIf`) and other game methods (like `turn.order.first`), so the best we can do is error if the hook type is `undefined` (which is probably OK, because *only* hooks should be used for events) without knowing exactly what method triggered the error.

#### Notes

I don’t particularly like that this exposes something new to the public plugin API, but I couldn’t really see how to do this otherwise as that is how the events plugin is built. One alternative could be to only expose this new data via some separate plugin field, like `_internal_fnWrap`, which would still be public but marked as private. Not sure exactly what the best approach is, but happy to see alternative approaches.

#### To-do

- [x] If we go with this model, we should expose the `GameMethod` object properly, so that plugin authors can import it. Probably from the `boardgame.io/core` package?
- [x] Are there any other scenarios where it would be good to error with some useful diagnostic information?

---

**CC:** @cristiands7 — Happy to hear your thoughts too!

### DIFF
--- a/docs/documentation/events.md
+++ b/docs/documentation/events.md
@@ -158,3 +158,21 @@ const game = {
 
 !> This doesn't apply to events in moves or hooks, but just the
 ability to call an event directly from a client.
+
+### Calling events from hooks
+
+The events API is available in game hooks like it is inside moves. However,
+because of how hooks and events interact, certain events cannot be called from
+certain hooks. The following table shows which hooks support which events.
+
+|                    | turn<br>`onMove` | turn<br>`onBegin` | turn<br>`onEnd` | phase<br>`onBegin` | phase<br>`onEnd` | game<br>`onEnd` |
+|-------------------:|:----------------:|:-----------------:|:---------------:|:------------------:|:----------------:|:---------------:|
+|         `setStage` |         ✅        |         ❌         |        ❌        |          ❌         |         ❌        |        ❌        |
+|         `endStage` |         ✅        |         ❌         |        ❌        |          ❌         |         ❌        |        ❌        |
+| `setActivePlayers` |         ✅        |         ✅         |        ❌        |          ❌         |         ❌        |        ❌        |
+|          `endTurn` |         ✅        |         ✅         |        ❌        |          ✅         |         ❌        |        ❌        |
+|         `setPhase` |         ✅        |         ✅         |        ✅        |          ✅         |         ❌        |        ❌        |
+|         `endPhase` |         ✅        |         ✅         |        ✅        |          ✅         |         ❌        |        ❌        |
+|          `endGame` |         ✅        |         ✅         |        ✅        |          ✅         |         ✅        |        ❌        |
+
+✅ = supported &nbsp;&nbsp;&nbsp; ❌ = not supported
--- a/packages/core.ts
+++ b/packages/core.ts
@@ -7,7 +7,15 @@
  */
 
 import { INVALID_MOVE } from '../src/core/constants';
+import { GameMethod } from '../src/core/game-methods';
 import { ActivePlayers, TurnOrder, Stage } from '../src/core/turn-order';
 import { PlayerView } from '../src/core/player-view';
 
-export { ActivePlayers, Stage, TurnOrder, PlayerView, INVALID_MOVE };
+export {
+  ActivePlayers,
+  GameMethod,
+  Stage,
+  TurnOrder,
+  PlayerView,
+  INVALID_MOVE,
+};
--- a/src/client/transport/local.test.ts
+++ b/src/client/transport/local.test.ts
@@ -19,28 +19,6 @@ import type { ChatMessage, Game, State, Store, SyncInfo } from '../../types';
 
 const sleep = (ms = 500) => new Promise((resolve) => setTimeout(resolve, ms));
 
-/**
- * Run a callback on an interval and resolve when it returns `true`.
- * @param condition Callback that should eventually return `true`.
- * @param interval Interval on which to test the condition callback (in ms).
- * @param timeout How long to wait for the condition to be `true` before rejecting (in ms).
- */
-const waitFor = (condition: () => boolean, interval = 50, timeout = 1000) =>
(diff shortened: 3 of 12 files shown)


cristiands7:
Another option could be:
```javascript
const plugin = {
  name: 'discerning-wrapper',
  fnWrap: (fn) => (...args) => {
    if (fn._fnType === GameMethod.Turn.ON_MOVE) {
      console.log('Looks like this is an onMove hook.);
    }
    return fn(...args);
  },
};
```

vdfdev:
Can we change from a string to an enum, in order to gain type safety?

vdfdev (dismissed):

delucis:
> Can we change from a string to an enum, in order to gain type safety?

Sure. I liked the expressiveness of the enum-like object (`GameMethod.Phase.ON_BEGIN` vs `GameMethod.PHASE_ON_BEGIN`), but I guess using a genuine enum would enforce importing the `GameMethod` for Typescript users.

What do you think about @cristiands7’s suggestion? I like that it’s a more “private” approach, but am a little wary of this kind of `Fn & { _fnType: GameMethod }` type.

vdfdev (dismissed):

delucis:
Here’s a summary of the scenarios that now produce errors. Some of these were previously harmless/ignored silently, so may produce breaks in some games, but hopefully this extra diagnostic information will steer people to better understanding the game flow.

- #### Event called outside hook
    Events must be called from moves or the `onBegin`, `onEnd`, and `onMove` hooks. This error probably means you called an event from other game code, like an `endIf` trigger or one of the `turn.order` methods.

- #### `endTurn` called in `onEnd`
    `endTurn` is disallowed in `onEnd` hooks — the turn is already ending. (Previously ignored silently.)

- #### Maximum number of turn endings exceeded
    This likely means game code is triggering an infinite loop.

- #### Phase event in `onEnd`
    `setPhase` & `endPhase` are disallowed in a phase’s `onEnd` hook — the phase is already ending. If you’re trying to dynamically choose the next phase when a phase ends, use the phase’s `next` trigger. (Previously ignored silently.)

- #### Stage event in `onEnd`
    `setStage`, `endStage` & `setActivePlayers` are disallowed in `onEnd` hooks. (Previously ignored silently.)

- #### Stage event in `phase.onBegin`
    `setStage`, `endStage` & `setActivePlayers` are disallowed in a phase’s `onBegin` hook. Use `setActivePlayers` in a `turn.onBegin` hook or declare stages with `turn.activePlayers` instead. (Previously ignored silently.)

- #### Stage event in `turn.onBegin`
    `setStage` & `endStage` are disallowed in `turn.onBegin`. Use `setActivePlayers` or declare stages with `turn.activePlayers` instead. (Previously permitted but buggy, closes #981.)

Issue #973, opened 2021-08-03
Feature Request: Stages should have their own move limit


jorbascrumps:
It should be possible to define individual move limits per stage. As an example:

1. In a "draw" phase the player draws 3 cards;
2. In a "play" phase the player then plays a single card from their hand

I know this is possible to handle in userland but I feel this could be a good addition. Thoughts?

delucis:
I think you mean something like:

```js
stages: {
  draw: {
    moveLimit: 3,
    moves: { draw },
  },
  play: {
    moveLimit: 1,
    moves: { play },
  },
},
```

Sounds like a good idea! Probably the value in the stage configuration should be overridden if an alternative `moveLimit` is passed in `activePlayers` or `setActivePlayers` (also taking account of [the “advanced” move limit options](https://boardgame.io/documentation/#/stages?id=advanced-move-limits)).

jorbascrumps:
That is the exact API I was thinking. I looked briefly into the implementation and it doesn't look like there are any existing systems in place for handling stage "state" (ie, no hooks for determining if the stage should end). Am I correct in thinking that the only way to currently end a stage is to call `setStage` or if `turn.moveLimit` is reached?

delucis:
Yes, stages currently can end/change via the `setStage`, `endStage` or `setActivePlayers` events (or when a turn/phase ends as that always wipes stages) and when the move limit is reached. There’s no `endIf` for stages currently.

delucis:
Ah, wait, not `turn.moveLimit` — we count moves in the `ctx._activePlayersNumMoves` map and each update compare that to `ctx._activePlayersMoveLimit`, which maps playerIDs to the limit (each of these looks something like `{ '0': 2, '1': 0 }`)

delucis:
Here’s the logic run each move to see if the move limit is reached:

https://github.com/boardgameio/boardgame.io/blob/bff1d29492858842a3c38c8c4dca72e2888dca73/src/core/flow.ts#L668-L673

jorbascrumps:
Ok, so we'll need a way to track number of moves within a stage in addition to the entire turn. We could add additional keys to `ctx` (`ctx._activePlayersStageNumMoves` and `ctx._activePlayersStageMoveLimit` - terrible example names), or we could convert the existing keys into objects (`{ turn: 1, stage: 1 }`). The former obviously has a smaller footprint but doesn't scale as nicely.

delucis:
[...]

PR #985, opened 2021-08-22
feat: deprecate `moveLimit` in favour of `minMoves`/`maxMoves`

### LINKED ISSUE #934: Can't force `endTurn` on early moves, with `moveLimit` > 1
To reproduce, create a game with this in its declaration:

```javascript
turn: 
{
  moveLimit: 2,
},
```

And then in a move, fire the endTurn event with the `force` argument:

```javascript
ctx.events.endTurn( { force: true } );
```

If you take this move as your first action, it just prints a debug message saying eg `INFO: cannot end turn before making 2 moves`. I'm not sure if `force` isn't properly propagating through the event, as in `flows.ts` it does seem to have logic to allow forcing the turn to end early.

For context, I'm trying to build a game where players have two actions by default, but one of those actions should immediately end the turn if taken first.

senritsu:
- `EndTurn` now checks `minMoves` instead of `moveLimit` like before to determine if ending the turn is possible
- `EndStage` now checks `minMoves` similar to how `EndTurn` does, previously it seemed to have no such check; in contrast to `EndTurn` it does not use the phase config, but instead checks the resolved values directly to support usage of `minMoves` in the args of `setActivePlayers`
- `moveLimits` is kept as-is, to prevent breaking changes; a rename to `maxMoves` or introducing `maxMoves` as an alias while deprecating `moveLimits` may be done at a later stage

In most places, I tried to keep the usage as much in-line with `moveLimits` as possible. Notable exceptions:
- `moveLimits` is used for checking if a turn should be ended **automatically** (where `minMoves` is not)
- `minMoves` is used to determine if **manually** ending a turn or stage is possible (where `moveLimits` now has no relevance anymore).

As this is my first time doing a deep-dive into the codebase, please give a thorough sanity check to everything. I am especially unsure about the integration in `EndStage`, and possible repercussions of checking or not checking for `minMoves` there. Additionally I am not sure if there are more places where a check would be needed. 

I also noticed that `Process` in `flow.ts` does not check if it should auto-`EndStage` based on `moveLimits`, in contrast to the similar check for `EndTurn`. Is this on purpose?

#### Checklist

- [x] Test coverage is 100%

---

Closes #934


delucis:
@senritsu Thanks for this! I’ll try to get a review to you soon.

senritsu:
Thinking about this some more, I am not sure if the whole minMoves/maxMoves split in it's current incarnation is worth the complexity it adds for its fairly constrained usability.

From how I understand it, `moveLimit` is really just a convenience option for

```js
endIf: (G, ctx) => ctx.somewhere.numMoves >= MOVE_LIMIT
```

On the other hand, `minMoves` is really only useful for one thing, blocking a manual `endTurn` (or `endStage`). This would probably be better served by a similar predicate to `endIf`, something like a hypothetical `canEnd`.

```js
canEnd: (G, ctx) => ctx.somewhere.numMoves >= MIN_MOVES
```

This would have the additional benefit of being way more flexible and useful, since it can be used to check for arbitrarily complex constraints and not just the number of moves. Additionally it would also be very useful for the UI, since a manual "End turn" button could be disabled properly when the turn (or phase/stage) can't be ended.

This seems like a better tradeoff between complexity budget and functionality, and it also raises the question if `moveLimit` itself is really necessary. Seeing as the game config is not serializable anyways, it doesn't seem to offer much benefit over a lambda in `endIf` and does add a fair bit of really specialized code.

senritsu:
To expand on the previous point, supposing the mentioned predicates both exist, it would also be fairly trivial to replace the convenience of a simple `moveLimit` property by a collection of higher-order functions. This would also expand the out-of-the-box functionality for some more obscure scenarios like all players having a shared move-pool.

```js
endIf: afterMoves(4) // equivalent to old `moveLimit`
endIf: madeMoves(4) // expands to (G, ctx) => ctx.something.numMoves >= 4
endIf: moveLimit({currentPlayer: 2, others: 3}) // different limits per player
endIf: moveCount({allCombined: 5}) // shared limit for all players
// functions should work identical for both predicates, naming should be carefully chosen to fit both
canEnd: afterMoves(4)
canEnd: madeMoves(4)
canEnd: moveLimit({currentPlayer: 2, others: 3})
canEnd: moveCount({allCombined: 5})
```

As an additional bonus, the relation between the somewhat unclear `moveLimit` and `endTurn` would be clarified slightly since it is explicitely assigned to `endIf`.

delucis:
Thanks for the thoughts — lot of good ideas there.

This proposal would also impact stages — currently stages have a `moveLimit` option that can be passed (see also [the stage move limit docs](https://boardgame.io/documentation/#/stages?id=advanced-move-limits)):

```js
setActivePlayers({ others: 'A', moveLimit: 2 });
```

Because this is dynamic, the move limit has to be serializable. (In general we’ve also steered clear of hooks for stages to avoid developers needing to think about who is in which stage too much — i.e. unlike `turn.endIf`, `stage.endIf` would be called *per* player rather than just once. Not impossible to understand obviously, but a declarative `moveLimit` has less moving parts.) There was a proposal (#973) to add a declarative `moveLimit` to stage configs themselves too, which seems a good idea, but `endIf`/`canEnd` here wouldn’t allow us to remove any of the `moveLimit` complexity because they couldn’t provide the dynamic `moveLimit` functionality.

Given that, for consistency across turns and stages, I think we can’t throw out support for the declarative approach. That said, I don’t see anything against a `canEnd` method for more flexibility.

In theory, we could refactor support for the `moveLimit` or `maxMoves` option to wrap `endIf`. [At the top of `flow.ts`](https://github.com/boardgameio/boardgame.io/blob/1116ab082d8ef3c1402537fdc66acbe541c8cfdf/src/core/flow.ts#L117-L146) there’s a bunch of code that normalizes a game definition for use. We could convert things there something like:

```js
    // Deprecate `moveLimit` without a hard break for now.
    if (phaseConfig.turn.moveLimit !== undefined) {
      logging.info(
        '`turn.moveLimit` is deprecated. Please use `turn.maxMoves` & `turn.minMoves` instead.'
      );
      phaseConfig.turn.maxMoves = phaseConfig.turn.moveLimit;
    }
    // Wrap `endIf` to check the move limit.
    if (phaseConfig.turn.maxMoves !== undefined) {
      { endIf } = phaseConfig.turn;
      phaseConfig.turn.endIf = (G, ctx) => (
        endIf(G, ctx) ||
        ctx.numMoves >= phaseConfig.turn.maxMoves
      );
    }
```

Then code in the rest of the flow reducer only needs to worry about calling `endIf`, not specifically checking `moveLimit`/`maxMoves`. The same approach could be used with `canEnd` and a `minMoves` option.

What do you think?

senritsu:
## Stages/Dynamic

I did actually wonder about how it would integrate with Stages, and I do agree that less cognitive overhead there is definitely a worthwhile goal.

Supporting the dynamic case also makes a lot of sense, generally I am a big fan of data-driven/declarative approaches. Professionally I currently use [mapbox](https://docs.mapbox.com/mapbox-gl-js/style-spec/), which does a lot of cool things in that direction, but having full serializable expression support may be a bit overkill at the moment. Worth a look for inspiration though, if the declarative approach is something that should be a focus. 

Intuitively it would kind of feel bad if the dynamic case were severely limited compared to the hook-based equivalent, but maybe in actual use it wouldn't matter too much. Having incomplete parallels between Turns and Stages feels similar. 

## Wrapping functions

Regarding implementing the serializable options with a hook wrapper, the main argument in favor would probably be extensibility and simplicity in terms of how much stuff needs to be carted around or inherited during the whole flow/context bookkeeping. There are roughly 35 usages of the internal lookup where the per-player move limits are stored, double now with `minMoves`. 

If internally everything would be compiled down to hook wrappers (which would probably require at least internal support for Stage hooks, in which case you might as well make them public for advanced use cases), it might cut down quite a bit of code, while simultaneously opening the path to add any number of similar declarative options (and of course supporting completely arbitrary behavior via the hooks themselves). The full ramifications are hard to theory-craft though, I would probably have to try some things hands-on to get a clearer picture. 

## Conclusions

I do think that the whole topic of refactoring the declarative options to use hooks, as well as providing a new `canEnd` (name pending) hook would be worthy of a separate issue/RFC, and seeing as the declarative/dynamic approach should stay in anyways I guess it would still make sense for the PR to move forward. 

The normalization function is great, I was hoping for something like that but haven't had the time to look for it yet 😁 That was the main reason I left `moveLimit` instead of renaming it to `maxMoves`, because I was missing the convenient entry point for backwards compatibility. Given it exists, making the naming consistent and just providing backwards compatibility (and a deprecation notice) like outlined in your snippet seems the best choice.

PS: on mobile, sorry for potential autocorrect screwups 🙃

senritsu:
Completely unrelated, docs update for `minMoves` (and `maxMoves` if changed) is also still missing from the PR.

src/core/turn-order.ts
```diff
@@ -24,6 +24,7 @@ export function SetActivePlayers(
   let activePlayers: typeof ctx.activePlayers = {};
   let _prevActivePlayers: typeof ctx._prevActivePlayers = [];
   let _nextActivePlayers: ActivePlayersArg | null = null;
+  let _activePlayersMinMoves = {};
   let _activePlayersMoveLimit = {};
```
delucis, line 28: For consistency, let’s rename this `_activePlayersMaxMoves` here and everywhere else.

src/types.ts
```diff
@@ -59,13 +59,16 @@ export type PartialGameState = Pick<State, 'G' | 'ctx' | 'plugins'>;
...
+  | StageName
+  | { stage?: StageName; minMoves?: number; moveLimit?: number };
 
 export interface ActivePlayersArg {
   currentPlayer?: StageArg;
   others?: StageArg;
   all?: StageArg;
   value?: Record<PlayerID, StageArg>;
+  minMoves?: number;
   moveLimit?: number;
```
delucis, line 72: We’ll need to update the types something like this. Using a `@deprecated` JSDoc tag for `moveLimit` adds some editor hinting where supported to further help highlight the change.

```suggestion
  minMoves?: number;
  maxMoves?: number;
  /** @deprecated Use `minMoves` and `maxMoves` instead. */
  moveLimit?: number;
```

Same would go for the `StageArg` type.

src/core/flow.test.ts
```diff
@@ -241,30 +241,51 @@ describe('turn', () => {
...
-  describe('moveLimit', () => {
+  describe('minMoves', () => {
     test('without phases', () => {
       const flow = Flow({
         turn: {
-          moveLimit: 2,
+          minMoves: 2,
         },
       });
+
```
delucis, line 251: “Stretch” goal here, if you have the time: would be good to wrap each block that ends with an `expect` assertion (or two) in its own `test`. Basically this formalises the comments you’ve added, so it’s more like:

```js
test('player 0 can’t end their turn when minMoves hasn’t been reached', () => {
  // ...
  expect(...);
});
```

There’ll also be a couple of tests that need updating, now that we are keeping `moveLimit` also setting `minMoves`.

delucis (changes requested):
OK, so mostly as discussed, here’s what I think this PR still needs.

- [ ] Maintain the current behaviour but deprecate `moveLimit`, by adding something like the following around L150 in `flow.ts`:

    ```js
        // Deprecate `moveLimit` without a hard break for now.
        if (phaseConfig.turn.moveLimit !== undefined) {
          logging.info(
            '`turn.moveLimit` is deprecated. Please use `turn.maxMoves` & `turn.minMoves` instead.'
          );
          phaseConfig.turn.minMoves = phaseConfig.turn.moveLimit;
          phaseConfig.turn.maxMoves = phaseConfig.turn.moveLimit;
        }
    ```

- [ ] We then should make sure we replace any places that check `moveLimit` to check `maxMoves` instead.

- [ ] Update docs to show `maxMoves`/`minMoves` instead of `moveLimit`.

- [ ] Add tests if required to make sure we’re testing the mapping of `moveLimit` to `minMoves`/`maxMoves`.

src/core/turn-order.ts
```diff
@@ -87,13 +92,22 @@ export function SetActivePlayers(
...
 
+    if (arg.minMoves) {
+      for (const id in activePlayers) {
+        if (_activePlayersMinMoves[id] === undefined) {
+          _activePlayersMinMoves[id] = arg.minMoves;
+        }
+      }
+    }
+
     if (arg.moveLimit) {
```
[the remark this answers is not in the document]
delucis: Ah, and we’ll need to do the same for `ApplyActivePlayerArgument`. Not sure if there’s any refactoring here that can centralise more of this logic.

senritsu:
After a long-ish break (thanks to surgery) I finally got around to update this PR. 

Most open tasks should be addressed, but I am having some trouble writing a test for backwards compatibility of `setActivePlayers`. The test is commented out atm, I am confused why it does not work and must be missing something fairly obvious. When debugging it, there is no indication of `setActivePlayers` actually having had any effect.

src/core/flow.test.ts
```diff
@@ -1783,6 +1959,74 @@ test('stage events should not be processed out of turn', () => {
...
+  //   });
+  //   let state = flow.init({ ctx: flow.ctx(2) } as State);
+
+  //   expect(state.ctx.turn).toBe(1);
+  //   expect(state.ctx.currentPlayer).toBe('0');
+
+  //   state = flow.processMove(
+  //     state,
+  //     makeMove('setActivePlayers', null, '0').payload
+  //   );
```
delucis, line 2009: The issue here is that testing flow like this doesn’t actually run the moves, it just does the framework stuff around that (incrementing counts etc.) It would be better to use a `processEvent` call here to dispatch the `setActivePlayers` event:

```js
    state = flow.processEvent(
      state,
      gameEvent('setActivePlayers', { all: 'A', moveLimit: 1 })
    );
```

Then you can assert that `_activePlayersMaxMoves` is set as expected and that `_activePlayersMinMoves` remains `null` for `moveLimit`.

delucis (changes requested):
Hey @senritsu! Hope you’re recovering well — sorry to hear that.

This looks good. Here are a few details and a suggested fix for your commented test.

src/core/turn-order.test.ts
```diff
@@ -779,7 +810,13 @@ describe('setActivePlayers', () => {
       const reducer = CreateGameReducer({ game });
       let state = InitializeGame({ game, numPlayers: 3 });
 
-      expect(state.ctx._activePlayersMoveLimit).toEqual({
+      expect(state.ctx._activePlayersMinMoves).toEqual({
+        '0': 1,
+        '1': undefined,
+        '2': undefined,
+      });
```
delucis, line 817: We should use [`toStrictEqual`](https://jestjs.io/docs/expect#tostrictequalvalue) to check the *absence* of the `undefined` keys here:

```suggestion
      expect(state.ctx._activePlayersMinMoves).toStrictEqual({ '0': 1 });
```

src/core/turn-order.test.ts
```diff
@@ -845,7 +888,13 @@ describe('setActivePlayers', () => {
       const reducer = CreateGameReducer({ game });
       let state = InitializeGame({ game, numPlayers: 3 });
 
-      expect(state.ctx._activePlayersMoveLimit).toEqual({
+      expect(state.ctx._activePlayersMinMoves).toEqual({
+        '0': undefined,
+        '1': 1,
+        '2': 2,
+      });
```
delucis, line 895: Same here:

```suggestion
      expect(state.ctx._activePlayersMinMoves).toStrictEqual({
        '1': 1,
        '2': 2,
      });
```

delucis (changes requested):
Looks good to go — just spotted one tiny testing improvement.

delucis (approved):

delucis:
@senritsu Thanks for this!

PR #987, opened 2021-08-23
fix(types): Expect Game.setup method to return G

alexjurkiewicz:
I'm fresh to this framework, so maybe I'm missing something. But this looks like a very small, simple improvement to the types 😊 

#### Checklist


delucis (approved):
Yeah, nice! I actually had this queued up in another branch along with a bunch of breaking changes, but should be safe to merge this on its own. Thanks!

PR #997, opened 2021-08-30
Create a client-specific Ctx type

alexjurkiewicz:
On the client side, context does not contain "random" methods, since
they rely on secret state. Derive a client-specific version of Ctx which
omits these methods.

There are probably other Ctx items which are not included on the
client-side, they could also be omitted in a future commit.

#### Checklist


src/types.ts
```diff
@@ -97,13 +97,13 @@ export interface Ctx {
...
     seed: string | number;
   };
+  random: RandomAPI;
 
   // TODO public api should have these as non-optional
   // internally there are two contexts, one is a serialized POJO and another
   // "enhanced" context that has plugin api methods attached
   events?: EventsAPI;
   log?: LogAPI;
-  random?: RandomAPI;
```
delucis, line 106: This should also apply to the other default plugins `log` and `events` as they suffer from the same problem and will also be always present in moves etc. but not on the client.

src/client/react.tsx
```diff
@@ -39,10 +39,12 @@ export type BoardProps<G extends any = any> = ClientState<G> &
     isMultiplayer: boolean;
   };
 
+export type CtxClient = Omit<Ctx, 'random' | '_random'>
```
delucis, line 42: There’s a missing `;` here. Try running `npm run prettier` on your changes to fix any style errors like this.

src/client/react.tsx
```diff
@@ -39,10 +39,12 @@ export type BoardProps<G extends any = any> = ClientState<G> &
...
+
 type ReactClientOpts<
   G extends any = any,
   P extends BoardProps<G> = BoardProps<G>,
-  CtxWithPlugins extends Ctx = Ctx
+  CtxWithPlugins extends CtxClient = CtxClient
```
delucis, line 47: I’m not sure this is the correct place for this. Firstly, [the plain JS client](https://github.com/boardgameio/boardgame.io/blob/main/src/client/client.ts) is the main source of truth for client types — the React client wraps it so they should be in sync. Secondly and slightly counter-intuitively, this type here is not actually used for `ctx` on the client I don’t think — it’s just here for compatibility with games that use custom plugins, so this type actually *should* include plugin APIs! (Confusing, I know.)

Probably `ctx` on the client wants updating here: https://github.com/boardgameio/boardgame.io/blob/b406e5a3ca1efa041de5c2b3d88cff6a773e0a0c/src/client/client.ts#L128-L134

Although, I’ll be interested to see if that’s enough, because internally we add and remove the plugin APIs at various points, so having `random` etc. no longer optional may cause errors there.

delucis (changes requested):

alexjurkiewicz:
Thanks for the feedback. Sadly, you are right -- this is much too complex for me to try and address for now :( I need some more familiarity with the codebase before trying such a radical change.

delucis:
Thanks for trying anyway! Once #891 is merged, this will all become much cleaner and until then, you’ll probably have to keep asserting the APIs are there like `ctx.random!.D6()`.

PR #1002, opened 2021-09-02
refactor(transport): Consolidate transport interface

delucis:
This PR refactors the transport interface to simplify the task of implementing new transports. (I’ve developed it out of the attempt at a peer-to-peer transport — I think eventually I’ll publish the P2P transport as a separate package.) This refactor moves various logic common to all transport implementations to the base `Transport` class and moves responsibility for interacting with the client’s Redux store to the client itself, making the transport “thinner”, primarily responsible for passing the master’s output to the client. It also renames some of the class methods to hopefully clarify their purpose.

#### Breaking Changes

If anyone had figured out how to write a custom transport, it will definitely break. I don’t know how likely that is as we don’t export many of the things you’d ideally need access to to create a custom transport.

#### Checklist


nicolodavis (dismissed):
Thanks for fixing my terrible naming conventions :)

Looks good to me.  I don't think we should be too concerned about this being a breaking change.

delucis:
Haha, I’m not sure if they were terrible 😄

My guess was just that the original class organically grew beyond its initial scope. I guess I should maybe also rename `subscribe` to `subscribeToConnectionStatus` while I’m at it.

Issue #1004, opened 2021-09-07
LobbyClient request fetch response


RGBKnights:
The code in question
```
    async request(route, init) {
        const response = await fetch(this.server + route, init);
        if (!response.ok) {
            let details;
            try {
                details = await response.json();
            }
            catch {
                try {
                    details = await response.text();
                }
                catch (error) {
                    details = error.message;
                }
            }
            throw new LobbyClientError(`HTTP status ${response.status}`, details);
        }
        return response.json();
    }
```

Based on this code you will never be able to get any errors from the server as text this is because fetch's response when accessed can not be accessed again and if you do you get an error about the response body has already been used. You can [solve this issue](https://stackoverflow.com/questions/40497859/reread-a-response-body-from-javascripts-fetch) by cloning the response before access.

delucis:
Oh, interesting. Do I understand correctly that `response.text()` fails because `response.json()` was already called?

Fancy submitting a PR applying the proposed fix?

RGBKnights:
> Oh, interesting. Do I understand correctly that `response.text()` fails because `response.json()` was already called?
Yes, this is because even if it fails you still accessed the underlying reader.

I will see if I can fine the time to submit a PR but it could be a while...

delucis:
OK, thanks for flagging this!

delucis:
Closed in #1005

PR #1005, opened 2021-09-08
LobbyClientError

RGBKnights:
Added the clone method to the first access of the response so when access again it dose not fail.
A fix for issue #1004.

#### Checklist

- [ ] Use a separate branch in your local repo (not `main`).


delucis (approved):
Thanks!

PR #1006, opened 2021-09-10
chore: improve changelog isVersionBump & commitID replacement regex

alvaromartmart:
Minor tweaks to the regex used to detect commit id & version bump commits

#### Checklist

### DIFF
--- a/scripts/changelog.js
+++ b/scripts/changelog.js
@@ -30,7 +30,7 @@ const isFeature = (s) => /feat(\(\w+\))?:/.test(s);
 const isUninformative = (s) =>
   /(test|style|chore|docs|ci)(\([\w-]+\))?:/.test(s);
 /** Tests if this commit just bumped the version (via `npm version`). */
-const isVersionBump = (s) => /^\w{8} \d+\.\d+\.\d+$/.test(s);
+const isVersionBump = (s) => /^\w+\s\d+\.\d+\.\d+$/.test(s);
 
 const changes = shell
   .exec(`git log --oneline "${PREVIOUS_TAG}"..`)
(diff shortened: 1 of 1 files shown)


scripts/changelog.js
```diff
@@ -51,14 +51,15 @@ const formatChanges = (changes) =>
         .replace(/[a-z]+\((\w+)\)/, '$1')
         // Linkify commit refs.
         .replace(
-          /^(\w{8})/,
+          /^(\w+)\s/,
```
delucis, line 54: I don’t think we want to replace the whitespace here.
```suggestion
          /^(\w+)/,
```

delucis (changes requested):

scripts/changelog.js
alvaromartmart: You're right, in fact I was re-introducing that space a couple of lines later. Fixed it, thanks!

scripts/changelog.js
```diff
@@ -51,14 +51,15 @@ const formatChanges = (changes) =>
...
-          /^(\w{8})/,
+          /^(\w+)/,
           '* [[$1](https://github.com/boardgameio/boardgame.io/commit/$1)]'
         )
         // Linkify PR references.
         .replace(
           /\(#(\d{3,})\)/,
           '([#$1](https://github.com/boardgameio/boardgame.io/pull/$1))'
         )
+        .replace(/\)](\w*)/, ')]$1')
```
delucis, line 62: It's not always super easy to understand what a regex is intended to do. What's the intention of this addition?

alvaromartmart: Oops, you're right 😅  that line is no longer necessary as we no longer need to re-introduce the space between `[commidID]` and `commit message`

delucis: Ahhh, ok! Gotcha 😄

delucis (approved):

PR #1035, opened 2021-11-29
fix: Add `ctx` type parameter to client

jonaswre:
#### Checklist

- [ x ] Use a separate branch in your local repo (not `main`).
- [ x ] Test coverage is 100% (or you have a story for why it's ok).

Hi, i realised there is an issue with the type implementation. 
When passing a Game that has a diffrent ctx. It expects the default Ctx in the javascript client.


delucis (approved):
Thanks @jonaswre!

Issue #1039, opened 2021-12-17
Option to collapse the debug panel on load


MathieuLoutre:
Is there a way to hide the debug panel by default and if not, would you take a PR for an option to do that? Additionally, would you accept a PR for an option to hide the new button for the debug panel?

My use case is that I bring up the panel to switch between players when testing but I don't need it shown by default. The new button also interferes with parts of my UI.

delucis:
Happy to see a PR for an option to have the panel collapsed on load. And if it’s helpful, I’d also merge a PR to opt-in to hiding the new show/hide button. Or we could even do this via CSS custom properties to allow control over the button position.

```css
.debug-panel {
  --bgio-debug-collapser-display: none;
  --bgio-debug-collapser-top: auto;
  --bgio-debug-collapser-bottom: .5em;
}
```

MathieuLoutre:
Thanks @delucis! I looked at the possibility of adding control via CSS Variables (which is a nice idea) but it seemed more convoluted. That said, the approach I've taken doesn't seem incompatible with adding further options for positioning down the line.

delucis:
Implemented in #1040, released in https://github.com/boardgameio/boardgame.io/tree/v0.49.8 — thanks @MathieuLoutre  🎉

PR #1040, opened 2021-12-20
feat(debug): Add collapse on load & hide toggle button options

MathieuLoutre:
Add more options to debug panel as per #1039

#### Checklist

### DIFF
--- a/docs/documentation/debugging.md
+++ b/docs/documentation/debugging.md
@@ -21,15 +21,15 @@ const client = Client({
 
 ### Debug Panel options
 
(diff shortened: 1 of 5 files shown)


delucis (commented):
Thanks for this @MathieuLoutre! Nice code and works well.

Maybe I’d just suggest slightly clearer naming for the options API?

Instead of `hidePanel` and `hideButton`, how about `collapseOnLoad` and `hideToggleButton`?

MathieuLoutre:
No problem, the new names are more explicit which is good. I updated across but also wrote better tests (hopefully). Let me know if anything needs to change.

delucis (approved):

delucis:
Perfect — those tests are a big improvement :+1:

PR #1060, opened 2022-02-20
fix(lobby): block creation of matches with invalid player counts

RevolutionTech:
#### Checklist

## Description

Fixes https://github.com/boardgameio/boardgame.io/issues/1055. Prevents a user from creating a match with an invalid number of players (only applicable if `minPlayers` and/or `maxPlayers` are defined).

## Testing

I tried to imitate the style and verbosity of the existing tests in `api.test.ts` with the tests I introduced, ~but technically some edge cases are not addressed by tests (eg. player count too low). If verbosity is preferred I'd be happy to add a few more test cases.~ but feel free to let me know if there's anywhere I can make improvements.

### DIFF
--- a/src/server/api.ts
+++ b/src/server/api.ts
@@ -122,9 +122,10 @@ export const configureRouter = ({
     if (!game) ctx.throw(404, 'Game ' + gameName + ' not found');
 
     if (
-      numPlayers != null &&
-      ((game.minPlayers != null && numPlayers < game.minPlayers) ||
-        (game.maxPlayers != null && numPlayers > game.maxPlayers))
+      ctx.request.body.numPlayers !== undefined &&
+      (Number.isNaN(numPlayers) ||
+        (game.minPlayers && numPlayers < game.minPlayers) ||
+        (game.maxPlayers && numPlayers > game.maxPlayers))
     ) {
       ctx.throw(400, 'Invalid numPlayers');
     }
--- a/src/server/api.ts
+++ b/src/server/api.ts
@@ -121,6 +121,14 @@ export const configureRouter = ({
     const game = games.find((g) => g.name === gameName);
     if (!game) ctx.throw(404, 'Game ' + gameName + ' not found');
 
+    if (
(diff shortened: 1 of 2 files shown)


src/server/api.ts
```diff
@@ -121,6 +121,14 @@ export const configureRouter = ({
     const game = games.find((g) => g.name === gameName);
     if (!game) ctx.throw(404, 'Game ' + gameName + ' not found');
 
+    if (
+      numPlayers != null &&
+      ((game.minPlayers != null && numPlayers < game.minPlayers) ||
+        (game.maxPlayers != null && numPlayers > game.maxPlayers))
```
delucis, line 127: Could we avoid the loose inequality operator here? I find it can be hard to reason about all the different ways operands are compared.

I'd propose treating `numPlayers` more carefully because it's user input. We don't need to be so careful about the game config so I'd just check `minPlayers`/`maxPlayers` is truthy. Perhaps something like:

```js
    if (
      numPlayers !== undefined &&
        (typeof numPlayers !== 'number' ||
          (game.minPlayers && numPlayers < game.minPlayers) ||
          (game.maxPlayers && numPlayers > game.maxPlayers))
    ) {
```

That way we also throw an error if `numPlayers` is not actually a number.

delucis (changes requested):
Thanks for this @RevolutionTech!

I've left a comment regarding the logic checking the validity of the input. More tests are always helpful especially for catching regressions later, but I'll leave it up to you how important it is to cover all edge cases.

src/server/api.ts
RevolutionTech: Sounds good. Since `Number.parseInt()` is used in the definition of `numPlayers` above and will always be a `number` (but could be `NaN`) I ended up using:
```javascript
ctx.request.body.numPlayers !== undefined && (Number.isNaN(numPlayers || ...
```

delucis (approved):
Nice catch with the `isNan`. Looks good to me — thanks!

Issue #1059, opened 2022-02-17
Conditional log redacting in long form move


mbrinkl:
It would be nice to have an option to pass G and ctx into the redact option in a long form move. The undoable property already has this option, so it should be fairly easy to implement. So it would looke like...

```typescript
B: {
  move: (G, ctx, ...args) => {},
  redact: true/false or (G, ctx) => {},
},
```

For my particular use case, I have moves in my game that should normally show up in the log, but I want the move args to be secret depending on G. I can strip the data from the player view, but I also need to redact the arg in the logs.

PR #1089, opened 2022-08-23
feat: Conditional log redacting in long form move

hughluo:
#### Checklist

Implemented as requested in #1059

### DIFF
--- a/src/core/reducer.ts
+++ b/src/core/reducer.ts
@@ -118,18 +118,10 @@ function initializeDeltalog(
     logEntry.metadata = pluginLogMetadata;
   }
 
-  function IsRedactFunction(
-    redact: boolean | ((...args: any[]) => any)
-  ): redact is (...args: any[]) => any {
-    return redact instanceof Function;
-  }
-
-  if (typeof move === 'object') {
-    if (IsRedactFunction(move.redact)) {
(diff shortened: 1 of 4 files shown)


src/core/reducer.ts
```diff
@@ -118,8 +118,18 @@ function initializeDeltalog(
...
-  if (typeof move === 'object' && move.redact === true) {
-    logEntry.redact = true;
+  function IsRedactFunction(
+    redact: boolean | ((...args: any[]) => any)
+  ): redact is (...args: any[]) => any {
+    return redact instanceof Function;
+  }
+
+  if (typeof move === 'object') {
+    if (IsRedactFunction(move.redact)) {
```
vdfdev, line 128: I think this can be easier to read with less nesting:
```
  if (typeof move === 'object' && move.redact === true) {
    logEntry.redact = true;
  else if (typeof move === 'object' && move.redact instanceof Function) {
   logEntry.redact = move.redact(state.G, state.ctx);
  }

vdfdev (changes requested):
LGTM, just a small improvement and we can merge

src/core/reducer.ts
vdfdev: I went ahead and applied this change to enable us to merge.

vdfdev (approved):

Issue #1096, opened 2022-09-21
BUG [v0.49.12] React Client: broken reactivity for React 18 (works fine with React 17)


pwoloszun:
package.json
```
    "boardgame.io": "~0.49.12",
    "react": "~18.2.0",
    "react-dom": "~18.2.0",
    "@types/react": "~18.0.20",
    "@types/react-dom": "~18.0.6",
    "typescript": "~4.8.3",

```
It modifies redux state, BUT do not rerender react component.
Works fine with React 17.0.2

Code:
```
import type { Game } from "boardgame.io";
import { Client, BoardProps } from 'boardgame.io/react';

// game state
interface MyGameState {
  someValue: number;
}

const MyGame: Game<MyGameState> = {
  setup: () => {
    return { someValue: 1 };
  },

  moves: {
    someMove: (G, ctx) => {
      G.someValue += 1;
    },
  },
};

// board component
interface MyGameProps extends BoardProps<MyGameState> { }

function MyGameBoard(props: MyGameProps) {
  const { G, moves } = props;
  console.log('RERENDER:', props);

  return (
    <div>
      <p>Val: {G.someValue}</p>
      <button onClick={() => moves.someMove()}>Test</button>
    </div>
  );
}

// game client
const App = Client({
  game: MyGame,
  board: MyGameBoard,
});

// React 18: app bootstrap
import React from 'react';
import ReactDOM from 'react-dom/client';

ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
  <React.StrictMode>
    <App />
  </React.StrictMode>
);


```

nsaritzky:
I've been trying to figure out why my board wasn't re-rendering. Reverting to React 17 fixed it.

pwoloszun:
> I've been trying to figure out why my board wasn't re-rendering. Reverting to React 17 fixed it.

Yes it works with React 17.
Still its a bug in React 18.

nsaritzky:
Right, I meant that only as a confirming anecdote.

Issue #1073, opened 2022-05-27
React board component doesn’t re-render with React 18


eurubkov:
Good afternoon!

Thanks for the great library that makes it easier to make games!
I was experimenting with recreating a card game called `The Game` and I am running into the issue where I call `moves.SomeMove(args)` from the UI, which does update the state of G but the UI doesn't seem to recognize it. The reason I know it gets updated is because if I place the same card on the same pile twice, it gives me an invalid move error in the console (as expected).

I don't do much of the front-end development, so it's likely I am just doing something wrong. 
Could anyone help me take a look at what it is?

This is my repo: https://github.com/eurubkov/the-game

To repro, just drag a card from the bottom row and drop it into any of the boxes at the top row. The top row should get updated with this new card and the card should disappear from the bottom row but neither of these things is happening. 

The move: `./src/Game.js PlayCard`
The Drop event in the UI: `./src/Board.js onDragDrop`

Appreciate the time and help!

delucis:
Thanks for the bug report! Our React client is currently incompatible with React 18 unfortunately, we need to update it.

For now you should be able to downgrade to React 17:

```shell
npm i react@^17 react-dom@^17 @testing-library/react@^12
```

And update `src/index.js` to use the React 17 API for setting up the app by doing the inverse of what is described in the 17 to 18 migration guide: <https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis>

Let’s keep this open to track React 18 support :-)

eurubkov:
> Thanks for the bug report! Our React client is currently incompatible with React 18 unfortunately, we need to update it.
> 
> For now you should be able to downgrade to React 17:
> 
> ```shell
> npm i react@^17 react-dom@^17 @testing-library/react@^12
> ```
> 
> And update `src/index.js` to use the React 17 API for setting up the app by doing the inverse of what is described in the 17 to 18 migration guide: https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis
> 
> Let’s keep this open to track React 18 support :-)

Appreciate the quick response! I will try it later and report back. How big of a job do you think adding React 18 support is going to be? I could try to tackle it

delucis:
I haven’t had a chance to dive into React 18 enough to be sure, but I wonder if there’s some deprecated method used in the `WrappedBoard` class we return from the `Client` factory:

[...]

PR #1104, opened 2022-10-26
Fix React 18 compatibility

### LINKED ISSUE #1068: Wrapping the app in React.StrictMode causes it not to update
Hi! This framework is awesome and I'm excited to start using it.

I was following along with the [tutorial](https://boardgame.io/documentation/#/tutorial) which is great, and used `create-react-app` to create the app (as recommended). It worked fine for the most part, BUT with one major problem: the React components did not re-render when the game state updated. I did some investigation and it appears to be a problem with the default create-react-app template. 

You can see the issue here: https://codesandbox.io/s/boardgame-io-react-typescript-demo-forked-undhfv
This is a fork of the provided Typescript-React codesandbox that demonstrates the problem.

Changes from the original codesandbox:
- update react, react-dom, @types/react, and @types/react-dom from version 17 to 18
- change index.tsx to look like this:

```tsx
import React from 'react';
import ReactDOM from 'react-dom/client';

import App from './App';

const root = ReactDOM.createRoot(document.getElementById('root')!);
root.render(
  <React.StrictMode>
    <App />
  </React.StrictMode>
);
```

^^^ Note that this is the default index.tsx that gets generated by the latest `create-react-app`.

FIX: if you comment out the `<React.StrictMode>` wrapping `<App />`, the DOM starts rendering again as the game state updates.

Note: the issue doesn't happen unless using `React.StrictMode` _and_ `ReactDOM.createRoot() / root.render` rather than just plain `render`.

Obviously it's not a huge deal since the fix is pretty straightforward; but probably still worth fixing so it doesn't trip up newbies like me.

mbrinkl:
Fixes broken client in react 18 strict mode.

Closes  #1068, #1073, #1096

The issue:  WrappedBoard unmounts and remounts in strict mode, and the debug panel did not unsubscribe the clients when it was destroyed. 

#### Checklist


delucis (approved):
Wow, amazing work tracking that down — thank you @mbrinkl!

I’ve tested this out locally, including wrapping the example app in `<StrictMode>` as that was reported as causing the problem in #1068 and everything appears to be working as expected. 🥳

mbrinkl:
Whoops, I had strict mode when debugging but then I recreated in a new branch and forgot to include it, thanks for catching that.

PR #1217, opened 2025-03-22
feat: Server handles games with custom plugin apis

fladrif:
Enable the server to start games with custom plugin api's by passing in parameters for the Game object. 

#### Checklist

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -57,8 +57,11 @@ export const getPortFromServer = (
   return address.port;
 };
 
-interface ServerOpts {
-  games: Game[];
+interface ServerOpts<
+  G extends any = any,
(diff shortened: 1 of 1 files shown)


fladrif:
@benbot If you wouldn't mind taking a look

devill:
Thanks for this, and sorry it sat here for well over a year with no reply — the project was dormant for a long stretch, and that's on us, not on you.

Your diagnosis was right: `ServerOpts.games: Game[]` was pinning the plugin API type at the server boundary, which is exactly why a game with custom plugin APIs got rejected there. That's now fixed on main in 65ca73b (via #1269), so I'm closing this as superseded. Thanks for taking the time to track it down and send a fix.

PR #1219, opened 2025-05-08
docs(typescript): update code examples

hidde-vb:
#### Checklist

- [x] Test coverage is 100% 

### Context

https://github.com/boardgameio/boardgame.io/issues/1207

Updated the basic usage example and the codesandbox with the new syntax.

### DIFF
--- a/docs/documentation/typescript.md
+++ b/docs/documentation/typescript.md
@@ -2,26 +2,29 @@
 
 boardgame.io includes type definitions for TypeScript.
 
-### Basic usage
+## Basic usage
 
 ```typescript
 // Game.ts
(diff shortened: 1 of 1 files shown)


devill:
Thanks for this, and sorry it sat here for over a year without a reply — the project was dormant for a long stretch, and that's on us rather than on you.

Closing this one as stale. Looking at it now, the `###` headings are actually the house style across the docs rather than a mistake, so the remaining change is smaller than it looks. The docs do need a proper pass — stale Playground links and inconsistent snippets among other things — and that's better done as one sweep than piecemeal. Thanks for taking the time to send this.

Issue #1204, opened 2024-12-23
yarn example fails with semantic error TS2339


dlee:
```
❯ yarn examples
yarn run v1.22.22
$ npm run start

> boardgame.io@0.50.2 prestart
> run-p examples:install build


> boardgame.io@0.50.2 examples:install
> node scripts/install-examples.js


> boardgame.io@0.50.2 build
> cross-env BABEL_ENV=rollup rollup --config rollup.config.js  --silent

Starting the application...
[!] (plugin rpt2) Error: /Users/dlee/code/boardgame.io/src/server/transport/socketio.ts(167,17): semantic error TS2339: Property 'io' does not exist on type 'BaseContext & DefaultContext & { db: Async | Sync; auth: Auth; }'.
src/server/transport/socketio.ts
Error: /Users/dlee/code/boardgame.io/src/server/transport/socketio.ts(167,17): semantic error TS2339: Property 'io' does not exist on type 'BaseContext & DefaultContext & { db: Async | Sync; auth: Auth; }'.
    at error (/Users/dlee/code/boardgame.io/node_modules/rollup/dist/shared/node-entry.js:5400:30)
    at throwPluginError (/Users/dlee/code/boardgame.io/node_modules/rollup/dist/shared/node-entry.js:11878:12)
    at Object.error (/Users/dlee/code/boardgame.io/node_modules/rollup/dist/shared/node-entry.js:12912:24)
    at Object.error (/Users/dlee/code/boardgame.io/node_modules/rollup/dist/shared/node-entry.js:12081:38)
    at RollupContext.error (/Users/dlee/code/boardgame.io/node_modules/rollup-plugin-typescript2/src/rollupcontext.ts:37:18)
    at /Users/dlee/code/boardgame.io/node_modules/rollup-plugin-typescript2/src/print-diagnostics.ts:41:11
    at arrayEach (/Users/dlee/code/boardgame.io/node_modules/rollup-plugin-typescript2/node_modules/lodash/lodash.js:516:11)
    at _.each (/Users/dlee/code/boardgame.io/node_modules/rollup-plugin-typescript2/node_modules/lodash/lodash.js:9342:14)
    at printDiagnostics (/Users/dlee/code/boardgame.io/node_modules/rollup-plugin-typescript2/src/print-diagnostics.ts:9:2)
    at Object.transform (/Users/dlee/code/boardgame.io/node_modules/rollup-plugin-typescript2/src/index.ts:225:5)
[...]

PR #1224, opened 2025-06-05
fix: fix example type errors

coreyar:
#### Checklist

Fixes issue #1204 

*src/server/transport/socketio.ts*
The assignment of `io` to `app.context` is not correct and since it is not correct and not used. I removed it. 

*src/server/api.ts*
Types params with an id property to be optional. `updatePlayerMetadata` is already checking that exists and if not it throws so this type makes sense

### DIFF
--- a/src/server/api.ts
+++ b/src/server/api.ts
@@ -392,7 +392,9 @@ export const configureRouter = ({
     ctx.body = body;
   });
 
-  const updatePlayerMetadata = async (ctx: Koa.Context) => {
+  const updatePlayerMetadata = async (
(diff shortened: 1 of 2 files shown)


Rupesh-ark:
Closing: the TS2339 build error this fixed (#1204) no longer reproduces after the toolchain modernization — `tsc --noEmit` is clean on main. On the socketio change specifically: `app.context.io` is intentional API rather than dead code (middleware uses it to reach the socket.io server, and the transport tests assert it), so we want to keep it. Thanks for the report and the PR.

PR #1241, opened 2026-07-03
ci: modernize workflow (Node 24/26, SHA-pinned actions) + dependabot

devill:
- Replace `test.yml` with `ci.yml`: build + test on Node 24 and 26, with all third-party actions SHA-pinned.
- Add `dependabot.yml` plus an auto-merge workflow for security/patch updates.

Independent of the dependency-upgrade series (branches off `main`).


devill:
Superseded by #1243, which combines the build, CI, and examples changes with the rest of the modernization into one CI-green PR. The atomic per-theme commits are preserved there. (These three couldn't pass upstream CI independently — the coverage and integration jobs need the toolchain, CI, and example-app changes together.)

Issue #31, opened 2017-12-19
Moves documentation linked but not written


pedromdias:
The [Client](https://google.github.io/boardgame.io/#/api/Client) links to the [Moves](https://google.github.io/boardgame.io/#/api/Moves) documentation but that page seems to not be implement as it simply states 'not found'

nicolodavis:
Thanks for the report. Will fix.

nicolodavis:
This is now fixed.

Issue #30, opened 2017-12-19
npm package problem - node v6.10.1


ahmetoz:
I'm not sure if this is an error but npm package does not works on node v6.10.1. (windows 10)

And thank you for this awesome framework. 👍 

```
11986 error Windows_NT 10.0.15063
11987 error argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "install" "--save" "boardgame.io"
11988 error node v6.10.1
11989 error npm  v3.10.10
11990 error code ELIFECYCLE
11991 error boardgame.io@0.11.4 postinstall: `ln -s ../packages/node_modules/boardgame.io`
11991 error Exit status 1
11992 error Failed at the boardgame.io@0.11.4 postinstall script 'ln -s ../packages/node_modules/boardgame.io'.
11992 error Make sure you have the latest version of node.js and npm installed.
11992 error If you do, this is most likely a problem with the boardgame.io package,
11992 error not with npm itself.
11992 error Tell the author that this fails on your system:
11992 error     ln -s ../packages/node_modules/boardgame.io
11992 error You can get information on how to open an issue for this project with:
11992 error     npm bugs boardgame.io
11992 error Or if that isn't available, you can get their info via:
11992 error     npm owner ls boardgame.io
11992 error There is likely additional logging output above.
11993 verbose exit [ 1, true ]
```

ChargerIIC:
I seem to be receiving the same error. I got this on Windows 10, executed via PowerShell on NPM 5.6.0:

> boardgame.io@0.11.4 postinstall C:\users\vburt\documents\github\boardgame\node_modules\boardgame.io
> ln -s ../packages/node_modules/boardgame.io

'ln' is not recognized as an internal or external command,
operable program or batch file.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.1.2 (node_modules\fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.1.2: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"x64"})

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! boardgame.io@0.11.4 postinstall: `ln -s ../packages/node_modules/boardgame.io`
npm ERR! Exit status 1
npm ERR!
[...]

Issue #29, opened 2017-12-19
hook for endTurn


kkoch986:
would be cool to be able to define a hook for when endTurn is called to clean up some state in between turns.

maybe im not approaching it correctly so feel free to tell if thats the case. A simple example would be having dice. if i store the current value of the dice in the game state, when the turn is over id like to be able to null out the current dice numbers so the UI knows its time to re-roll.

for now a workaround is to define my own move called endturn which updates the state and in the UI code calling both simultaneously

Felizardo:
I think this is already implemented here:
https://github.com/google/boardgame.io/blob/master/src/both/reducer.js#L54

Which means that if one of your moves is 'END_TURN', this move will be executed before doing the rest of the end turn calculations...

kkoch986:
yea sorry i should have picked through the code a bit more.

Looking at it more though its already provided and they dont expose a way to easily set it (https://github.com/google/boardgame.io/blob/master/src/both/game.js#L43-L58) 

what i did realize though after looking at that code is implementing the following move would accomplish what i want:

```
END_TURN(G, ctx, id) {
  		return {
	  		...G,
	  		currentDice: null
	  	}	
  	}
```

its not super pretty because it still shows up as an extra move. but it does get called automatically when the normal endTurn is called so i no longer need to call both.

happy to close this though since that does work better than what i started with. thanks for the quick response!

PR #1257, opened 2026-07-04
Port server + client fixes from the lean-poker fork

### LINKED ISSUE #1100: Infinite loading screen with simultaneous match joins
Hi ! i have a custom lobby mechanism on my project which implement basic matchmaking between players.
When sufficient players have joined the lobby i create the match from my api and send an event to all the players to redirect them to the match page.
When the match page load, an http call occurs to log the user to the game server.

It seems that the boardgame implementation cannot handle simultaneous connection changes properly.
When someone logs into the match (https://github.com/boardgameio/boardgame.io/blob/main/src/master/master.ts#L400)
the match metadata is fetched from my database, modified to include the new metadata, sent to all clients and then saved to the database.

When 2 players joins the match simultaneously, this logic is called 4 times simultaneously and it looks like that some players are overriding the metadata changes of previous users.

For example:

Player 1 is redirected to match page
Player 2 is redirected to match page

Player 1 onConnectionChange method start and fetch metadata
Player 2 onConnectionChange method start and fetch metadata

Player 1 onConnectionChange method modify metadata and save it to the database
Player 2 onConnectionChange method modify metadata and save it to the database

as Player 2 fetched the metadata before Player 1 finished modifying and saving it to the database, when Player 2 modify and save the metadata into the database, the data is outdated and does not contain Player 1 isConnected and credentials values. Player 2 is correctly logged into the match but Player 1 stay on the "Connecting..." screen indefinitely and the match get stuck when his turn start...

Sometimes everything plays nicely in the correct order but most of the time, for example in a 4 players match, a player stay on the loading screen and the match is blocked, this is really critical for my project, how can this be fixed ?

devill:
Ports three functional changes that were developed and reviewed on the lean-poker fork but never landed upstream. Each is a separate commit; all fork-identity churn (the `@lean-poker` rename, fork CHANGELOG/breaking-changes docs, version bumps) has been stripped — these are clean, upstream-appropriate diffs.

### Changes
1. **`fix(server)`: configurable API body-parser limit (`apiBodyLimit`)** — replaces the per-route `koaBody()` parsers with one shared parser whose size limit is configurable (default `'5mb'`), fixing 413 errors on large `setupData`. _(lean-poker #29)_
2. **`feat(client)`: add `loadState`, rename `overrideGameState` → `previewState`** — `previewState` only previews; new `loadState()` makes a loaded state the actual playable state (dispatches update, clears log). _(lean-poker #30)_
3. **`fix(server)`: serialize `sync`/`disconnect` via the per-match queue** — routes both handlers through the existing per-match `PQueue` (as `update` already is) to fix a race in `Master.onConnectionChange` that left clients stuck "Connecting…". Fixes #1100. _(lean-poker #31)_

> [!WARNING]
> **Breaking change** (commit 2): `Client.overrideGameState()` is renamed to `Client.previewState()`. Kept in this PR per request since it was already reviewed on the fork, but happy to split it out if you'd prefer to version it separately.

### Verification
Clean checkout on the pnpm/Node 24 stack: `eslint` clean, `tsc --noEmit` clean, **42 suites / 811 tests pass**, `build` succeeds.

### DIFF
--- a/docs/documentation/api/Client.md
+++ b/docs/documentation/api/Client.md
@@ -192,6 +192,18 @@ The following methods are available on a client instance:
 - `redo()`: Function that redoes the previously undone move.
 
 
+- `previewState(state)`: Visual override. Replaces the state returned by
+  `getState()` (and passed to subscribers) without changing the internal
+  store. Useful for time-travel debugging or AI preview. Call with no
+  arguments (or `null`) to clear the override.
+
+
+- `loadState(state)`: Replaces the entire internal game state with a saved
+  state object, allowing play to continue from that point. Use this to
+  implement save / load or to reconstruct a game from an external source.
+  The state must contain `G`, `ctx` and `_stateID` keys.
+
+
 - `sendChatMessage(message)`: Function that sends a chat message to other
   players. The `message` argument can be a string or you can send objects
   to include more metadata.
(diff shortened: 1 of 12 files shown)


devill:
I now notice that Claude reauthored the commits under my name. @Rupesh-ark, if you prefer, you can also create the same PR, and I'll approve it. That way your author ship is kept

Rupesh-ark (approved):
Since these are my own changes from the fork which is already reviewed, imma just approve and merge this.

PR #1259, opened 2026-07-08
fix(client): stop subscriber ID collisions after unsubscribe

### LINKED ISSUE #1137: Client.subscribe breaks after unsubscribing
The Client subscribe method has an bug in how it internally stores the handles to registered subscribers. The code starts like this:

```typescript
 subscribe(fn: (state: ClientState<G>) => void) {
    const id = Object.keys(this.subscribers).length;
    this.subscribers[id] = fn;
    ...
```

It is using the count of subscribes as a property index to register a new subscriber. That will then break future subscribers if anyone but the most recent subscriber unsubscribes.

As an example scenario:
1. Subscriber A subscribes to the client and gets assigned ID 0
2. Subscriber B subscribes to the client and gets assigned ID 1
3. Subscriber A unsubscribes. The count of subscribers is now 1
4. Subscriber C subscribes to the client and gets assigned ID 1... again. Subscriber B no longer gets notifications.
5. Subscriber D subscribes to the client and gets assigned ID 1... again... again. Subscriber C no longer gets notifications.
6. etc...

The client implementation could keep the public facing API it has but internally use a [CustomEvent](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent) object to store the subscribed handlers to avoid managing the subscriptions itself.

Rupesh-ark:
## Problem

`Client.subscribe` minted new subscriber IDs from the *current subscriber count*:

```ts
const id = Object.keys(this.subscribers).length;
```

Count and next-free-ID only coincide while occupied IDs are contiguous from 0. As soon as any subscriber **other than the most recent one** unsubscribes, the count points at an ID that is still live, and the next `subscribe` silently overwrites that subscriber. The victim keeps thinking it is subscribed but never receives another state update.

Reproducing the scenario from #1137: A subscribes (ID 0), B subscribes (ID 1), A unsubscribes → count is 1 → C subscribes and is assigned ID 1, evicting B.

This survived unnoticed because typical consumers (the React binding, the debug panel) subscribe once at startup and never unsubscribe mid-session — the existing `multiple subscriptions` test also only exercises the safe unsubscribe-the-newest sequence.

## Fix

Mint IDs from a monotonic per-client counter (`nextSubscriberID++`) so an ID is never reused. Two-line change; storage, iteration, and the unsubscribe closure are untouched, and the public API is unchanged.

Went with this rather than the `CustomEvent` rewrite suggested in the report: it fixes the actual defect (ID minting) at minimal diff, and avoids taking a dependency on a DOM-flavored event system in code that also runs in Node (bots, tests).

## Testing

- New regression test reproducing the exact issue scenario (fails on `main`, asserts the evicted subscriber still receives updates, the new subscriber works, and the unsubscribed one stays silent).
- Full suite green: 880 tests / 42 suites, lint, `tsc --noEmit`.

Fixes #1137

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -141,8 +141,7 @@ export class _ClientImpl<
   readonly multiplayer: (opts: TransportOpts) => Transport;
   private reducer: Reducer;
   private _running: boolean;
-  private subscribers: Record<string, (state: State<G> | null) => void>;
-  private nextSubscriberID: number;
+  private subscribers: Map<symbol, (state: State<G> | null) => void>;
   private transport: Transport;
   private manager: ClientManager;
   readonly debugOpt?: DebugOpt | boolean;
@@ -188,8 +187,7 @@ export class _ClientImpl<
     this.debugOpt = debug;
     this.manager = GlobalClientManager;
     this.previewStateData = null;
-    this.subscribers = {};
-    this.nextSubscriberID = 0;
+    this.subscribers = new Map();
     this._running = false;
 
     this.reducer = CreateGameReducer({
@@ -409,7 +407,7 @@ export class _ClientImpl<
   }
 
   private notifySubscribers() {
-    Object.values(this.subscribers).forEach((fn) => fn(this.getState()));
+    this.subscribers.forEach((fn) => fn(this.getState()));
   }
 
   previewState(state: any) {
@@ -447,10 +445,8 @@ export class _ClientImpl<
   }
 
   subscribe(fn: (state: ClientState<G>) => void) {
-    // IDs must never be reused, so derive them from a monotonic counter
-    // rather than the current subscriber count (issue #1137).
-    const id = this.nextSubscriberID++;
-    this.subscribers[id] = fn;
+    const id = Symbol();
+    this.subscribers.set(id, fn);
     this.transport.subscribeToConnectionStatus(() => this.notifySubscribers());
 
     if (this._running || !this.multiplayer) {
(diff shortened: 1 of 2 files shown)


src/client/client.ts
```diff
@@ -445,7 +447,9 @@ export class _ClientImpl<
   }
 
   subscribe(fn: (state: ClientState<G>) => void) {
-    const id = Object.keys(this.subscribers).length;
+    // IDs must never be reused, so derive them from a monotonic counter
+    // rather than the current subscriber count (issue #1137).
+    const id = this.nextSubscriberID++;
```
devill: That is a valid solution, but I might consider using some sort of universally unique identifier. Since boardgame.io might be used under heavy parallelism relying on a counter can create race conditions.

devill (changes requested):
Tentative RFC. I'm not 100% this is necessary, but my spidy senses tell me that incrementing counters in concurency heavy systems is a mysterious bug waiting to happen.

src/client/client.ts
Rupesh-ark: Updated this to use Symbol() identifiers stored in a Map. This guarantees unique local subscription identities without relying on a counter or introducing UUID generation, and matches the pattern already used by ClientManager.

Also, since subscriber is synchronous, it'll not have a race condition so we are all good xD

devill: Well... you never know when something will become async 😅

devill (approved):

Issue #1136, opened 2023-02-17
Tutorial with React doesn't work


wjl:
I tried following the tutorial here, using the React version: https://boardgame.io/documentation/#/tutorial

The first part which has you run these commands works fine:
```
npx create-react-app bgio-tutorial
cd bgio-tutorial
npm install boardgame.io
touch src/Game.js
```

If at this point you do an "npm start" you get a normal React app placeholder launched in the browser and everything seems okay.

However, the next step in the tutorial replaces src/App.js . After doing this and running "npm start", I get this error:

```
Error in ./~/boardgame.io/dist/cjs/react.js
Module parse failed: /home/wjl/work/boardgame.io-tutorial/bgio-tutorial/node_modules/boardgame.io/dist/cjs/react.js Unexpected token (118:24)
You may need an appropriate loader to handle this file type.
SyntaxError: Unexpected token (118:24)
 @ ./src/App.js 7:13-42
```

I've used both Nodejs and React for a few small projects, but I'm by no means fluent with them. But the "cjs" caught my eye and after some searching on the web I played around with some advice there, installed the esm module, changed index.js to force esm use, and eventually kind of got it to launch, but at that point it had other problems (the boardgame.io debug interface ran but wouldn't update game state) and still couldn't finish the tutorial. 

For background this was done on an Ubuntu 22.04.01 LTS system using the nodejs snap. (The  nodejs deb package is too old to work with React.) Node version is:
```
node --version
v18.13.0
```

Anyway, I'm going to try the tutorial again without React, but I wanted to let you know that the tutorial, at least for me, appears to be broken. My best guess is that something changed with React or with some other dependency, and it's probably an easy fix for someone who knows what they are doing. =)

wjl:
Just to follow up, I did the tutorial with plain js and it worked fine.

wjl:
If it helps, here are the package.json and package-lock.json generated by running the tutorial commands (the only difference in this case was I used a different folder name, "tictactoe-react" so I didn't clobber the plainjs version I just tried). I didn't manually edit these files at all.

[package.json.zip](https://github.com/boardgameio/boardgame.io/files/10771179/package.json.zip)
[package-lock.json.zip](https://github.com/boardgameio/boardgame.io/files/10771182/package-lock.json.zip)

Binteimran853:
[...]

PR #1262, opened 2026-07-11
chore(server): replace koa-body with @koa/bodyparser

Rupesh-ark:
## Motivation

The lobby API only parses JSON and form bodies, but `koa-body` drags in formidable and its multipart-upload machinery — 14 transitive packages the server never touches. `@koa/bodyparser` is the official koajs body parser, handles exactly the json/form cases the lobby needs, and ships a modern dual ESM/CJS build (relevant groundwork for the upcoming ESM packaging work, see #902 / #1136).

## Changes

- `koa-body` → `@koa/bodyparser` in `dependencies` (net −14 packages in the lockfile)
- `src/server/api.ts`: swapped the import and the middleware factory call; per-route mounting and all options unchanged

## Behavior

Unchanged. `jsonLimit`/`formLimit` still honor the `apiBodyLimit` server option (`string | number`, default `'5mb'`). The existing `>1MB setupData` test proves the limit flows through: co-body's built-in default is 1mb, so that test only passes if our option reaches the parser.

## Testing

- `npx tsc --noEmit` ✓
- `pnpm exec eslint src/server/` ✓
- Full suite: 43 suites, 880 tests, all passing

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,6 +161,7 @@
     "typescript": "^5.9.3"
   },
   "dependencies": {
+    "@koa/bodyparser": "^6.1.0",
     "@koa/cors": "^5.0.0",
     "@koa/router": "^15.5.0",
     "@types/koa": "^3.0.2",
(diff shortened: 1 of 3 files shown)


devill (approved):
LGTM

PR #1266, opened 2026-07-15
Return an error for turn events in phase onBegin

### LINKED ISSUE #1237: events.endTurn() is silently discarded when called inside a phase's onBegin hook
Calling `events.endTurn({ next: playerID })` inside a phase's `onBegin` hook has **no effect**. The turn is not redirected. The call is silently discarded with no error or warning.

## Proof

```typescript
const game = {
  phases: {
    main: {
      start: true,
      turn: { order: TurnOrder.DEFAULT },
      onBegin: ({ events }) => {
        events.endTurn({ next: "1" });  // <-- silently discarded
      },
      moves: { ... },
    },
  },
};

// After phase starts:
// ctx.currentPlayer === "0"   (should be "1")
// ctx.turn === 1
```

The same call in `turn.onBegin` works correctly:

```typescript
turn: {
  onBegin: ({ ctx, events }) => {
    if (ctx.turn === 1) {
      events.endTurn({ next: "1" });  // <-- works
    }
  },
},

// After phase starts:
// ctx.currentPlayer === "1"   (correct)
// ctx.turn === 2
```

Tested with boardgame.io v0.50.2, Local() multiplayer, Node.js v22.22.0.

### Reproduction Script

Run from `packages/game/` with `node`:

```javascript
const { Client } = require('boardgame.io/client');
const { Local } = require('boardgame.io/multiplayer');
const { TurnOrder } = require('boardgame.io/core');

// Test 1: endTurn({ next }) in phase onBegin — FAILS
const game1 = {
  setup: () => ({ target: '1' }),
  phases: {
    main: {
      start: true,
      turn: { order: TurnOrder.DEFAULT },
      onBegin: ({ events }) => {
        events.endTurn({ next: '1' });
      },
      moves: {
        doSomething: ({ G }) => { G.done = true; },
      },
    },
  },
};

const clients1 = [];
for (let i = 0; i < 3; i++) {
  const c = Client({ game: game1, numPlayers: 3, multiplayer: Local(), playerID: String(i) });
  c.start();
  clients1.push(c);
}

setTimeout(() => {
  const state1 = clients1[0].getState();
  console.log('TEST 1 — endTurn in phase.onBegin:');
  console.log('  currentPlayer:', state1.ctx.currentPlayer, '(expected: 1, got:', state1.ctx.currentPlayer + ')');
  console.log('  PASS:', state1.ctx.currentPlayer === '1' ? 'YES' : 'NO — endTurn was discarded');
  clients1.forEach(c => c.stop());

  // Test 2: endTurn({ next }) in turn.onBegin — WORKS
  const game2 = {
    setup: () => ({ target: '1' }),
    phases: {
      main: {
        start: true,
        turn: {
          order: TurnOrder.DEFAULT,
          onBegin: ({ ctx, events }) => {
            if (ctx.turn === 1) {
              events.endTurn({ next: '1' });
            }
          },
        },
        moves: {
          doSomething: ({ G }) => { G.done = true; },
        },
      },
    },
  };

  const clients2 = [];
  for (let i = 0; i < 3; i++) {
    const c = Client({ game: game2, numPlayers: 3, multiplayer: Local(), playerID: String(i) });
    c.start();
    clients2.push(c);
  }

  setTimeout(() => {
    const state2 = clients2[0].getState();
    console.log('\nTEST 2 — endTurn in turn.onBegin:');
    console.log('  currentPlayer:', state2.ctx.currentPlayer, '(expected: 1, got:', state2.ctx.currentPlayer + ')');
    console.log('  PASS:', state2.ctx.currentPlayer === '1' ? 'YES' : 'NO');
    clients2.forEach(c => c.stop());
  }, 100);
}, 100);
```

Expected output:

```
TEST 1 — endTurn in phase.onBegin:
  currentPlayer: 0 (expected: 1, got: 0)
  PASS: NO — endTurn was discarded

TEST 2 — endTurn in turn.onBegin:
  currentPlayer: 1 (expected: 1, got: 1)
  PASS: YES
```

## Why It Happens

boardgame.io processes a phase start in two sequential steps inside `flow.ts`:

```
Step 1: StartPhase() — line 301
  ├── Run phase.onBegin(state)
  │     Our code calls events.endTurn({ next: "1" })
  │     This QUEUES the event: { type: "endTurn", turn: 0 }
  │
  └── Queue StartTurn for next step

Step 2: StartTurn() — line 313
  ├── InitTurnOrderState() sets ctx.currentPlayer = playOrder[0]
  ├── ctx.turn = ctx.turn + 1    (turn becomes 1)
  └── Run turn.onBegin(state)
```

After both steps, the Events plugin processes the dispatch queue (`Events.update()`, line 124). It finds our queued endTurn event:

```typescript
// events.ts line 139
const turnHasEnded = event.turn !== state.ctx.turn;
// event.turn = 0 (when it was dispatched)
// state.ctx.turn = 1 (incremented by StartTurn)
// turnHasEnded = true

// events.ts line 187
if (turnHasEnded) continue EventQueue;  // <-- SKIPPED
```

The event was dispatched during turn 0, but StartTurn already advanced to turn 1. boardgame.io's stale-event protection sees the turn has changed and skips the event. This is intentional behavior to prevent processing events from previous turns — but it incorrectly catches events dispatched in phase onBegin because the turn counter advances between dispatch and processing.

janmarsino98:
## Summary

- Add a TurnEventInPhaseBegin error for endTurn and pass calls from phase.onBegin.
- Preserve stale-event protection and return an explicit error instead of silently discarding the event.
- Point game authors to turn.order.first for choosing the starting player.
- Add regression coverage for both endTurn and pass.

## Why

A phase onBegin hook runs before the phase's first turn starts. A queued turn-ending event therefore becomes stale after StartTurn advances the turn counter and was previously discarded without feedback. Processing it would create a phantom turn, so the existing stale-event protection is correct; the missing behavior is an actionable error.

Fixes #1237

## Validation

- Focused phase.onBegin tests: 8 passed.
- ESLint on the two modified files: passed.
- Build: passed.
- git diff --check: passed.
- Full Jest run: 41 of 42 suites passed and all 873 executed tests passed. The debug suite could not load @testing-library/svelte-core because of an unrelated Windows ESM transform/path issue.

### DIFF
--- a/src/core/flow.test.ts
+++ b/src/core/flow.test.ts
@@ -1733,9 +1733,45 @@ describe('events in hooks', () => {
 
       client.events.setPhase('A');
       state = client.getState();
-      expect(state.ctx.turn).toBe(2);
-      expect(state.ctx.currentPlayer).toBe('1');
+      expect(state.ctx.turn).toBe(1);
+      expect(state.ctx.currentPlayer).toBe('0');
+      expect(state.ctx.phase).toBeNull();
+      expect(error).toHaveBeenCalled();
+      const errorMessage = (error as jest.Mock).mock.calls[0][0];
+      expect(errorMessage).toMatch(/events plugin declared action invalid/);
+      expect(errorMessage).toMatch(
+        /`endTurn` is disallowed in a phase’s `onBegin` hook/,
+      );
+      expect(errorMessage).toMatch(
+        /Use `turn.order.first` to choose the starting player/,
+      );
+    });
+
+    test('cannot pass from phase.onBegin', () => {
+      const client = Client({
+        game: {
+          phases: {
+            A: {
+              start: true,
+              onBegin: ({ events }) => events.pass(),
+            },
+          },
+        },
+      });
+
+      const state = client.getState();
+      expect(state.ctx.turn).toBe(1);
+      expect(state.ctx.currentPlayer).toBe('0');
       expect(state.ctx.phase).toBe('A');
+      expect(error).toHaveBeenCalled();
+      const errorMessage = (error as jest.Mock).mock.calls[0][0];
+      expect(errorMessage).toMatch(/events plugin declared action invalid/);
+      expect(errorMessage).toMatch(
+        /`endTurn` is disallowed in a phase’s `onBegin` hook/,
+      );
+      expect(errorMessage).toMatch(
+        /Use `turn.order.first` to choose the starting player/,
+      );
     });
 
     test('can end turn from turn.onBegin at start of phase', () => {
--- a/src/core/flow.test.ts
+++ b/src/core/flow.test.ts
@@ -1697,6 +1697,12 @@ describe('events in hooks', () => {
       events.endTurn();
     };
 
+    const conditionalPass = ({ G, events }) => {
+      if (!G.shouldEnd) return;
+      G.shouldEnd = false;
+      events.pass();
+    };
+
     test('can end turn from turn.onBegin', () => {
       const client = Client({
         game: { moves, turn: { onBegin: conditionalEndTurn } },
@@ -1740,7 +1746,7 @@ describe('events in hooks', () => {
       const errorMessage = (error as jest.Mock).mock.calls[0][0];
       expect(errorMessage).toMatch(/events plugin declared action invalid/);
       expect(errorMessage).toMatch(
-        /`endTurn` is disallowed in a phase’s `onBegin` hook/,
+        /`endTurn` & `pass` are disallowed in a phase’s `onBegin` hook/,
       );
       expect(errorMessage).toMatch(
         /Use `turn.order.first` to choose the starting player/,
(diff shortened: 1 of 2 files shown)


Rupesh-ark (changes requested):
1. Fix the error messages to name pass (both new firing paths):
- TurnEventInPhaseBegin: `endTurn` & `pass` are disallowed in a phase’s `onBegin` hook, no turn has started yet.
- EndTurnInOnEnd now also fires for pass, so it should become `endTurn` & `pass` are disallowed in `onEnd` hooks, the turn is already ending. (matches the existing StageEventInOnEnd convention of listing every covered event).
- Update the test assertions accordingly, the new pass test currently asserts the misleading endTurn-only text, which would lock it in.

2. Add tests for pass in onEnd hooks. The case 'pass' merge silently changes that behavior from "loop until MaxTurnEndings trips (~numPlayers × 100 iterations) then fail with a cryptic error" to "fail fast with a targeted error". That's a good change but has zero coverage, one test for turn.onEnd and one for phase.onEnd, mirroring the existing endTurn onEnd tests at flow.test.ts:1767 and 1790.

janmarsino98:
Thanks for the detailed review — addressed in 587a601.

- Both hook errors now name endTurn and pass, and the existing assertions cover the corrected wording.
- Added regression coverage for pass() in both turn.onEnd and phase.onEnd; each verifies the targeted error is returned immediately without advancing the game state.

Validation passed: the hook block (13 tests), flow.test.ts (133 tests), TypeScript, and the build. The full Jest run completed with 41/42 suites and 875 tests passing; the remaining debug suite still has the pre-existing ESM transform failure in @testing-library/svelte-core.

Could you please take another look?

Rupesh-ark (approved):
LGTM!

Issue #1228, opened 2025-07-11
Log plugin does not work on hooks, only on moves


gabrielecastellano:
Hello everyone!

I use the ctx.log object to show a game log client-side. It works fine for moves, as I can add metadata before the move ends.

However, what if I want to log something outside of a move? Let's say, on some event? For instance, I have a 'onBegin' method in one of my phases, and I want to log something from that method. The ctx.log object is visible there, but If I try to append any metadata there, nothing arrives client-side as that happens outside of a move.

However, the event itself (without any metadata) has its own entry in the log client-side.

Am I doing anything wrong or this is a known limitation? Any way to work around this?

Thanks

fladrif:
Looking at the documentation and code, plugins should be executing in both game events and moves. Do you have an example of it not working?

gabrielecastellano:
My codebase is quite big. I can try to reproduce it on the tic-tac-toe example. 
Meanwhile, I assume you are able to log during events handlers?

Gabriele

gabrielecastellano:
Please check out [this example](https://codesandbox.io/p/sandbox/boardgame-io-forked-x9yz56).

I have modified the tic-tac-toe app introducing two phases. In the first phase, you can only click a button that ends the phase, while the second phase has the usual clickCell move. 

I have tried to `log.setMetadata()` both at `onEnd` of the first phase, and at `onBegin` of the second phase. You can verify from the debug sidebar that those log entries are not reported (there is only the entry for the `endPhase` event, and it doesn't have any metadata). On the other end, all the metadata set during moves are correctly attached to their log entries.

<img width="711" height="350" alt="Image" src="https://github.com/user-attachments/assets/0310a7ee-eade-4285-b07a-fe966bed5d51" />

fladrif:
Thanks, yea I see the issue, it's with hooks and plugins. Plugins work correctly for moves and events, but hooks don't persist their data in execution to state here https://github.com/boardgameio/boardgame.io/blob/main/src/core/flow.ts#L332. All the wrapped hooks will probably need to return all state and not just `{ G }`. I can create an MR, but it doesn't look like @benbot has been active to merge. @nicolodavis @delucis Can I be an interim maintainer? 

note: hooks are the onEnd or onBegin functions you're writing. Events in the context of bgio are specifically the ones called from { events } = FnContext.

fladrif:
[...]

PR #1267, opened 2026-07-15
fix: preserve log metadata for game events

janmarsino98:
## Summary

- Preserve metadata set through `log.setMetadata()` while processing directly dispatched game events.
- Attach metadata to the canonical internal event entry when one exists.
- If no canonical entry is emitted, add a sanitized entry for the directly dispatched event.
- Cover `setPhase` from no active phase and `endGame`, without attributing metadata to adjacent internal events.
- Preserve rollback behavior so metadata from plugin-rejected actions cannot leak.
- Document the supported scope and retain a TODO for hooks triggered by moves.

## Root cause

`GAME_EVENT` processing generated its deltalog entries before plugins were flushed, but never copied the log plugin metadata into those entries. Hook metadata was therefore discarded during the flush.

Some directly dispatched events can produce several internal entries or no canonical entry at all. The final attribution rule is deterministic: use the canonical internal entry when present; otherwise create one sanitized entry for the dispatched event using its pre-action turn and phase.

Part of #1228.

## Remaining limitation

Metadata set by hooks triggered from a move, or by automatic turn triggers such as `turn.maxMoves` and `endIf`, is not yet attached to the resulting event log entry. This PR intentionally does not address that attribution path.

## Validation

Validated at `fe905a1` against `upstream/main@65ca73b`:

- Targeted log metadata tests: 7 passed, 1 TODO.
- Reducer, flow, and log metadata tests: 213 passed, 1 TODO.
- Negative control: reverting the no-entry fallback makes both new regression tests fail; restoring it makes them pass.
- `pnpm run ts` — passed.
- ESLint and Prettier on all modified files — passed.
- `pnpm run build` — passed.
- `pnpm run test:integration` — passed, including consumer typecheck, Vitest, build, ESM/CJS smoke tests, and publint.
- Full Jest coverage run on Windows: 41/42 suites passed and all 882 executed tests passed. The sole failure is the existing Jest ESM transform error for `@testing-library/svelte-core`.
- GitHub Actions build/test and integration — passed on Node 24.x and 26.x.
- Coveralls — passed.

The Windows checkout inherits CRLF line endings, so full-repository lint reports formatting errors in untouched files; the modified files pass ESLint and Prettier.

AI assistance: OpenAI Codex assisted with investigation, implementation, and validation.

### DIFF
--- a/src/core/reducer.ts
+++ b/src/core/reducer.ts
@@ -173,24 +173,73 @@ function initializeDeltalog(
 }
 
 /**
- * Add metadata set by the log plugin to the final log entry for an event.
+ * Get the event type used in the log for a directly dispatched game event.
  */
-function addLogMetadata(state: State): State {
+function getLogEventType(eventType: string): string | undefined {
+  switch (eventType) {
+    case 'endTurn':
+    case 'pass': {
+      return 'endTurn';
+    }
+    case 'endPhase':
+    case 'setPhase': {
+      return 'endPhase';
+    }
+    case 'endStage':
+    case 'setStage': {
+      return 'endStage';
+    }
+    case 'endGame': {
+      return 'endGame';
+    }
+  }
+}
+
+/**
+ * Add metadata set by the log plugin to the log entry for an event.
+ */
+function addLogMetadata(state: State, action: ActionShape.GameEvent): State {
   const metadata = state.plugins.log?.data.metadata;
   const { deltalog } = state;
+  const eventType = getLogEventType(action.payload.type);
 
-  if (metadata === undefined || !deltalog?.length) {
+  if (metadata === undefined || !deltalog?.length || eventType === undefined) {
     return state;
   }
 
-  const lastEntry = deltalog.length - 1;
+  const eventEntry = deltalog.findIndex(
+    (entry) =>
+      entry.action.type === Actions.GAME_EVENT &&
+      entry.action.payload.type === eventType,
+  );
+  if (eventEntry === -1) return state;
+
   const updatedDeltalog = [...deltalog];
-  updatedDeltalog[lastEntry] = {
-    ...updatedDeltalog[lastEntry],
+  updatedDeltalog[eventEntry] = {
+    ...updatedDeltalog[eventEntry],
     metadata,
   };
 
   return { ...state, deltalog: updatedDeltalog };
+}
+
+/**
+ * Remove metadata set during an action that was rejected.
+ */
+function clearLogMetadata(state: State): State {
+  const logPlugin = state.plugins.log;
+  if (logPlugin?.data.metadata === undefined) return state;
+
+  const data = { ...logPlugin.data };
+  delete data.metadata;
+
+  return {
+    ...state,
+    plugins: {
+      ...state.plugins,
+      log: { ...logPlugin, data },
+    },
+  };
 }
 
 /**
@@ -210,7 +259,11 @@ function flushAndValidatePlugins(
   if (!isInvalid) return [newState];
   return [
     newState,
-    WithError(oldState, ActionErrorType.PluginActionInvalid, isInvalid),
+    WithError(
+      clearLogMetadata(oldState),
+      ActionErrorType.PluginActionInvalid,
+      isInvalid,
+    ),
   ];
 }
 
@@ -322,7 +375,7 @@ export function CreateGameReducer({
       }
 
       case Actions.GAME_EVENT: {
-        state = { ...state, deltalog: [] };
+        const oldState = (state = { ...state, deltalog: [] });
 
         // Process game events only on the server.
         // These events like `endTurn` typically
@@ -356,14 +409,18 @@ export function CreateGameReducer({
 
         // Process event.
         let newState = game.flow.processEvent(state, action);
-        newState = addLogMetadata(newState);
+        newState = addLogMetadata(newState, action);
 
         // Execute plugins.
         let stateWithError: TransientState | undefined;
-        [newState, stateWithError] = flushAndValidatePlugins(newState, state, {
-          game,
-          isClient: false,
-        });
+        [newState, stateWithError] = flushAndValidatePlugins(
+          newState,
+          oldState,
+          {
+            game,
+            isClient: false,
+          },
+        );
         if (stateWithError) return stateWithError;
 
         // Update undo / redo state.
--- a/src/core/reducer.ts
+++ b/src/core/reducer.ts
@@ -173,6 +173,27 @@ function initializeDeltalog(
 }
 
 /**
+ * Add metadata set by the log plugin to the final log entry for an event.
+ */
+function addLogMetadata(state: State): State {
+  const metadata = state.plugins.log?.data.metadata;
+  const { deltalog } = state;
+
+  if (metadata === undefined || !deltalog?.length) {
+    return state;
+  }
+
+  const lastEntry = deltalog.length - 1;
+  const updatedDeltalog = [...deltalog];
+  updatedDeltalog[lastEntry] = {
+    ...updatedDeltalog[lastEntry],
+    metadata,
+  };
+
+  return { ...state, deltalog: updatedDeltalog };
+}
+
+/**
  * Update plugin state after move/event & check if plugins consider the action to be valid.
  * @param state Current version of state in the reducer.
  * @param oldState State to revert to in case of error.
@@ -335,6 +356,7 @@ export function CreateGameReducer({
 
         // Process event.
         let newState = game.flow.processEvent(state, action);
+        newState = addLogMetadata(newState);
 
         // Execute plugins.
         let stateWithError: TransientState | undefined;
--- a/src/core/reducer.ts
+++ b/src/core/reducer.ts
@@ -26,7 +26,7 @@ import type {
   TransientState,
   Undo,
 } from '../types';
-import { stripTransients } from './action-creators';
+import { gameEvent, stripTransients } from './action-creators';
 import { ActionErrorType, UpdateErrorType } from './errors';
 import { applyPatch } from 'rfc6902';
 import { RemovePlayer } from './turn-order';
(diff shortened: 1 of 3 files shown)


Rupesh-ark (changes requested):
Thanks for this, but before we merge this, 

This fixes only part of #1228. The new addLogMetadata runs in the GAME_EVENT reducer case, so it covers hooks reached via a directly dispatched event (client.events.endTurn()). But #1228 is about hooks generally, and the most common way hooks fire is via moves. Metadata is still dropped when the turn ends through a move-queued event or an automatic trigger:

const game = {
  moves: {
    finish: ({ events }) => events.endTurn(),
  },
  turn: {
    onEnd: ({ log }) => log.setMetadata({ message: 'turn ended' }),
  },
};
// After client.moves.finish():
// log[1] is the endTurn entry — metadata is undefined

The same happens with turn.maxMoves / endIf auevents are dispatched during the events pluginflush, and the log plugin flush wipes data.met() pass ([...CORE_PLUGINS, custom,PluginEvents].reverse() — events first, log afit to the deltalog. Fixing that path properlyhas attribution subtleties (move-entry metadatn't want to expand this PR's scope — instead:- Please change "Fixes #1228" to "Part of #122close while the move-triggered path stillfails, and add a line to the PR body stating t- Please scope the docstring down accordingly t, you can set metadata… and it will beavailable on that action's log entry" currentlich still doesn't work. Something like:"metadata set during a move, or during hooks rame event, is attached to that action's logentry."
- A test.todo('preserves metadata from hooks ts') would keep the gap visible in the suiteuntil the follow-up lands.

janmarsino98:
Addressed the requested scope updates in commit 74ee90e:

- Changed Fixes #1228 to Part of #1228 and documented the remaining move-triggered and automatic-event limitation.
- Narrowed the LogPlugin docstring to the supported behavior.
- Added a test.todo reminder for metadata from hooks triggered by moves.

Validation passed: targeted Jest, pnpm run ts, and pnpm run build. Targeted ESLint is blocked by the checkout's pre-existing CRLF formatting errors.

Could you please take another look?

Rupesh-ark (changes requested):
Two things to fix:

1. Attach metadata to the correct event entry
Don’t always update the last deltalog entry. A direct endTurn can automatically generate an endPhase, causing the turn metadata to be attached to the phase entry instead.
2. Prevent metadata leaking from rejected actions
Clear or roll back log metadata when plugin validation rejects an action, so it cannot be attached to the next valid event.

Add a regression test for each case. Everything else, including the scope clarification and existing tests, looks good.

janmarsino98:
Addressed both requested issues in commit 553ac94:

- Log metadata is now attached to the canonical entry for the directly dispatched event instead of always updating the last deltalog entry. This keeps endTurn metadata on the turn entry when the same action also ends the phase.
- Plugin-rejected actions now roll back with log metadata cleared, preventing it from leaking into the next valid event.

Added a regression test for each case.

Validation passed: targeted Jest, TypeScript, targeted ESLint, production build, and all GitHub CI / integration jobs on Node 24 and 26.

Could you please take another look?

Rupesh-ark (changes requested):
You need to bump up the test coverage, it's failing rn.

There is a question that we need to answer to merge this,

Where should hook metadata go when one dispatched event creates zero or several internal log entries? @devill will need your input here.

src/core/reducer.ts
```diff
@@ -167,6 +167,76 @@ function initializeDeltalog(
...
+ * Get the event type used in the log for a directly dispatched game event.
+ */
+function getLogEventType(eventType: string): string | undefined {
+  switch (eventType) {
+    case 'endTurn':
+    case 'pass': {
+      return 'endTurn';
+    }
+    case 'endPhase':
+    case 'setPhase': {
```
Rupesh-ark, line 185: setPhase is mapped to an endPhase entry. When ctx.phase === null, no endPhase entry is created, even though the new phase’s onBegin hook runs. Metadata set by that hook reaches eventEntry === -1 and is flushed.

src/core/reducer.ts
```diff
@@ -167,6 +167,76 @@ function initializeDeltalog(
...
+      return 'endPhase';
+    }
+    case 'endStage':
+    case 'setStage': {
+      return 'endStage';
+    }
+    case 'endGame': {
```
Rupesh-ark, line 192: getLogEventType('endGame') returns endGame, but EndGame never creates an endGame log entry. Metadata set in game.onEnd therefore cannot be attached and is discarded.

Either define and add an endGame log entry, or remove this mapping and narrow the documented scope. Add,a regression test for client.events.endGame().

Rupesh-ark (changes requested):

src/core/reducer.ts
janmarsino98: Addressed in `fe905a1`. Metadata now goes to the canonical internal event entry when one exists. If no canonical entry is produced, the reducer appends a sanitized entry for the directly dispatched event. For `setPhase` with `ctx.phase === null`, this creates a `setPhase` entry using the pre-action turn/phase and leaves the adjacent automatic `endTurn` entry untouched. Added a regression test for this exact case; reverting the fallback makes it fail.

src/core/reducer.ts
janmarsino98: Addressed in `fe905a1`. A directly dispatched `endGame` that produces no canonical `endGame` entry now gets one explicit sanitized log entry, so metadata from `game.onEnd` is preserved without being attributed to another internal event. The regression test calls `client.events.endGame()`, verifies the metadata, and verifies that exactly one entry is emitted; reverting the fallback makes the test fail.

janmarsino98:
Addressed the final review feedback in `fe905a1`.

- Defined a deterministic attribution rule: canonical internal event entry when present; otherwise one sanitized entry for the directly dispatched event.
- Added regressions for `setPhase` with no active phase and for `client.events.endGame()`.
- Confirmed both tests fail when the fallback is reverted and pass when restored.
- Recovered reducer coverage; GitHub Actions build/test and integration pass on Node 24 and 26, and Coveralls is green.
- Replied to and resolved both inline review threads.

@Rupesh-ark @devill, could you please take another look?

Rupesh-ark:
Reviewed `fe905a1` — implementation looks correct. Verified locally: targeted suites pass, the negative control reproduces (reverting the fallback fails both new regression tests), and the event→entry mapping checks out against the flow internals.

One design point still open: with the fallback rule, a directly dispatched `endGame` (or `setPhase` with no active phase) gets a log entry only when a hook set metadata, so the log shape becomes metadata-dependent. We'll wait for @devill's input on that before merging.

devill:
On the open design point: create the entry whether or not metadata was set. A log whose shape
depends on whether some hook happened to call `setMetadata` is harder to reason about than one
extra entry.

The missing `endGame` entry was never a deliberate choice. No commit in this repo's history ever
gave `EndGame` one — `EndPhase`, `EndTurn` and `EndStage` each append an entry, and `EndGame`
delegates to `EndPhase` without ever getting the same treatment. The `setPhase` case is the same
accident: `EndPhase` returns early when there's no active phase, and that return sits above the
log-entry block.

**`disableLog` is bypassed — worth fixing either way**

The fallback doesn't check `game.disableLog`, so with `disableLog: true` and metadata set,
`endGame` and `setPhase` still write entries. Gating on metadata hides this today; making the
entry unconditional turns it into a failure in the existing `disableLog` test.

Both changes:

```diff
-  if (metadata === undefined || eventType === undefined) {
+  if (game.disableLog || eventType === undefined) {
     return state;
   }
```

with `game` passed in, `metadata` attached to the fallback entry only when defined, and an early
return before the canonical-entry update when it isn't.

I tried this locally: full suite green (42 suites, 896 tests), `tsc` clean, and the entries
propagate through the master and a late-joining client's sync unchanged.

Rupesh-ark:
Pushed `76e720b` to this branch — @devill's change, plus a narrowing.

Applied as proposed: the fallback entry is now created whether or not metadata was set, and it honours `game.disableLog`. The bypass was real — with `disableLog: true` and metadata set, `endGame` was still writing an entry.

Narrowed the fallback to `endGame` and `setPhase`. Creating an entry for *every* dispatched event with no canonical entry also logs events the flow **refused**. Verified against the unconditional version:

- `endTurn` blocked by `minMoves` → `endTurn` entry, `ctx.turn` unchanged
- `endStage` blocked by `minMoves` → `endStage` entry, stage unchanged
- `endPhase` with no active phase (pure no-op) → `endPhase` entry
- `setStage` from no stage → new `setStage` entry

The debug panel re-dispatches every non-automatic entry through the reducer to rewind (`Log.svelte`), so a refused `endTurn` becomes false history. `endGame` and `setPhase` from `ctx.phase === null` are the only two events that run hooks without ever producing an entry of their own — everything else either has a canonical entry or runs no hooks, so no metadata is lost by leaving them alone.

Regressions added: entry present with no metadata set (`setPhase` and `endGame`), `disableLog` suppressing both, and a flow-refused `endTurn` logging nothing. Full suite green (42 suites, 899 tests), `tsc` clean, `reducer.ts` back to 100% coverage.

Worth noting in the release notes: `client.events.endGame()` now produces a log entry where it previously produced none.

Two follow-ups, deliberately not in this PR:

1. An `endIf`-triggered game end still produces no entry, so log shape depends on *how* the game ended. The durable fix is appending the entry inside `EndGame` in `flow.ts` via `appendLogEntry` (marked `automatic` when triggered), which gets `disableLog` for free and makes the reducer fallback unnecessary for `endGame`.
2. #1228 stays open for metadata from hooks triggered by moves and automatic turn triggers — the `test.todo` keeps it visible.

Thanks @janmarsino98 — merging once CI is green.

Rupesh-ark (approved):
Approved. Implementation, tests, and coverage verified locally at 76e720b: 42 suites / 899 tests pass, tsc clean, reducer.ts at 100%. Follow-ups tracked in the comment above.

devill (approved):

gabrielecastellano:
Hello, thank you for this contribution!
If I correctly understand, this preserves metadata set during hooks (`onBegin`, `onEnd`) only if the event leading to the hook was not triggered from within a move (e.g., calling `ctx.events.endTurn()`)? Does this mean that if a hook is triggered because of an `endIf` condition, then the metadata will be preserved? If not, could you reference an example where this fix applies? I could not think of any based on my understanding of the framework and the way I have used it so far.

devill:
Looking into it, it seems like you are right about that. To answer your question directly: no — metadata from hooks triggered by `endIf` (or from within moves) is still not attached; this change only covers events dispatched directly from the client. I didn't really consider if this part of the problem is worth fixing on its own. I still think that this is good ground work for a second part that actually fixes your issue. In case you have a PR for that we are happy to review it, otherwise it will need to wait for someone to pick it up.

PR #1268, opened 2026-07-16
fix(debug): hide state controls in multiplayer

### LINKED ISSUE #1034: Hide/warn debug controls in multiplayer scenarios
Currently we expose some controls in the debug panel to allow for quick state resets as well as state save/restore:

Top part of the boardgame.io debug panel featuring reset, save, and restore controls

These make use of the local client reducer to manipulate the client state and if used in a multiplayer set up will break the client’s connection.

It would be better to either hide these controls or disable them and show a warning for multiplayer clients, e.g. (quick ugly mock-up)

Top part of the boardgame.io debug panel with the controls covered in a warning about them being disabled in multiplayer contexts

janmarsino98:
Closes #1034

## Summary
- Hide reset, save, and restore controls for multiplayer clients.
- Keep the visibility toggle available.
- Explain why the state controls are unavailable.
- Add coverage for multiplayer debug controls.

## Validation
- pnpm install --frozen-lockfile
- pnpm run ts
- pnpm run build

The targeted Jest command currently fails before test execution because the existing Jest configuration does not transform @testing-library/svelte-core; the same error reproduces on upstream/main.

### DIFF
--- a/src/client/debug/main/Controls.svelte
+++ b/src/client/debug/main/Controls.svelte
@@ -43,6 +43,9 @@
 </style>
 
 <ul id="debug-controls" class="controls">
+  {#if client.multiplayer}
+  <li>State controls are unavailable in multiplayer games.</li>
+  {:else}
   <li>
     <Hotkey value="1" onPress={client.reset} label="reset" />
   </li>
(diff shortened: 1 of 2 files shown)


devill (approved):
Nice :) I wonder if the end point handling this should also be gated, but just hiding the buttons is a nice UX improvement.

janmarsino98:
Thanks, devill! Do you think this is good to merge as is? And would you prefer tracking the endpoint gating in a separate issue?

devill:
Looked more into the code, and actually it's not needed. So that closes this thread. Thanks again for contributing.

PR #1269, opened 2026-07-16
fix(server): accept games with custom plugin APIs

### LINKED ISSUE #1050: Plugin setup ctx issue when using typescript.
I am attempting to use a plugin for the 1st time (having seen a recent reference to the [bgio-effects](https://www.npmjs.com/package/bgio-effects) plugin) ... it looks just like something I really want to use.

All was going fine until I got to the point of adding my game to the game server, and then the typescript compiler is throwing an error.

I have reduce this down to a small example:

    import { EffectsCtxMixin } from 'bgio-effects';
    import { EffectsPlugin } from 'bgio-effects/plugin';
    import { Ctx, Game } from 'boardgame.io';
    import { Server } from 'boardgame.io/server';

    const EffectsConfig = {
        effects: {
            mediocreCards: {},
        },
    } as const;

    interface GameState {
        dummy: number;
    }

    export const MediocrityGame: Game<
        GameState,
        Ctx & EffectsCtxMixin<typeof EffectsConfig>
    > = {
        name: 'Mediocrity',

        plugins: [EffectsPlugin(EffectsConfig)],
    };

    export const server = Server({
        games: [MediocrityGame],
    });

This causes the following error to be generated on the `games: [MediocrityGame],` line:

    Type 'Game<GameState, Ctx & EffectsCtxMixin<{ readonly effects: { readonly mediocreCards: {}; }; }>, any>' is not assignable to type 'Game<any, Ctx, any>'.
      Types of property 'setup' are incompatible.
        Type '((ctx: Ctx & EffectsCtxMixin<{ readonly effects: { readonly mediocreCards: {}; }; }>, setupData?: any) => GameState) | undefined' is not assignable to type '((ctx: Ctx, setupData?: any) => any) | undefined'.
          Type '(ctx: Ctx & EffectsCtxMixin<{ readonly effects: { readonly mediocreCards: {}; }; }>, setupData?: any) => GameState' is not assignable to type '(ctx: Ctx, setupData?: any) => any'.
            Types of parameters 'ctx' and 'ctx' are incompatible.
            
Am I doing something wrong?

As I was looking at how plugins worked, I was wondering if the supplied `PlayerPlugin` should also provide a `PlayerCtxMixin` to allow it to be added to the default `Ctx` (otherwise I cannot see how to use the PlayerPlugin when using typescript?

janmarsino98:
## Summary

- Allow `Server` to accept games with custom plugin API types without a type assertion.
- Preserve support for heterogeneous game lists.
- Keep the Jest scenario as runtime coverage for processing a custom-plugin game and a standard game together through `transport.init`.
- Add a strict consumer type fixture to the packed-package integration test so the type regression is enforceable.

## Root cause

`ServerOpts.games` was typed as `Game[]`, which fixes the plugin API parameter to the default type. A game whose `setup` requires a custom plugin API was therefore rejected at the server boundary even though the server only processes the configuration and does not consume that API.

The server input now accepts `Game<any, any, any>[]`; processed games continue through the existing runtime path unchanged.

Fixes #1050.

## Regression guard

The integration harness installs the packed tarball and runs `tsc --noEmit` with `strict: true` against a consumer fixture importing only public APIs from `boardgame.io` and `boardgame.io/server`.

- Positive control: the strict fixture passes with `ServerOpts.games: Game<any, any, any>[]`.
- Negative control: in an isolated worktree, reverting only that declaration to `Game[]` makes the fixture fail at the `Server({ games: [...] })` call with `TS2322`; the `setup` context is incompatible because `custom` is missing.

## Validation

Validated at `0a4f196` against `upstream/main@3dc2af0`, using Node 24.14.0 and pnpm 10.16.0:

- `pnpm install --frozen-lockfile` — passed.
- `pnpm run lint` — passed from an LF snapshot of the published commit.
- `pnpm exec jest src/server/index.test.ts --runInBand` — passed (18 tests).
- `pnpm run ts` — passed.
- `pnpm run build` — passed.
- `pnpm run test:integration` — passed, including the new strict typecheck, Vitest fixture, build, ESM/CJS smoke tests, and publint.
- `git diff --check` — passed.
- `pnpm run test:coverage` — 41/42 suites passed and all 876 executed tests passed. The sole failure is the Windows Jest ESM transform error for `@testing-library/svelte-core/src/index.js` (`Unexpected token 'export'`). The exact command reproduces on a clean LF snapshot of `upstream/main@3dc2af0`: 41/42 suites passed and all 875 executed tests passed.
- GitHub Actions — build/test and integration passed on Node 24.x and 26.x.
- [Coveralls](https://coveralls.io/builds/80697228) — passed; coverage remained at 100.0%.

The normal Windows checkout inherits CRLF line endings, which makes Prettier report errors across untouched files. Validation used Git-archive LF snapshots so lint ran against the repository's actual blob contents.

AI assistance: OpenAI Codex assisted with investigation, implementation, and validation.

### DIFF
--- a/integration/package.json
+++ b/integration/package.json
@@ -11,12 +11,14 @@
     "@testing-library/react": "^16.0.1",
     "@vitejs/plugin-react": "^4.3.4",
     "jsdom": "^25.0.1",
+    "typescript": "^5.9.3",
     "vite": "^5.4.10",
     "vitest": "^2.1.4"
   },
   "scripts": {
     "start": "vite",
     "build": "vite build",
-    "test": "vitest run"
+    "test": "vitest run",
+    "typecheck": "tsc --project type-tests/tsconfig.json"
   }
 }
--- a/integration/type-tests/server-custom-plugin.ts
+++ b/integration/type-tests/server-custom-plugin.ts
@@ -1,0 +1,21 @@
+import type { Game, Plugin } from 'boardgame.io';
+import { Server } from 'boardgame.io/server';
+
+interface CustomPluginAPIs extends Record<string, unknown> {
+  custom: { enabled: boolean };
+}
+
+const customPlugin: Plugin<CustomPluginAPIs['custom']> = {
+  name: 'custom',
+  api: () => ({ enabled: true }),
+};
+
+const customGame: Game<{ enabled: boolean }, CustomPluginAPIs> = {
+  name: 'custom-plugin-game',
+  plugins: [customPlugin],
+  setup: ({ custom }) => ({ enabled: custom.enabled }),
+};
+
+Server({
+  games: [customGame, { name: 'standard-game' }],
+});
--- a/integration/type-tests/tsconfig.json
+++ b/integration/type-tests/tsconfig.json
@@ -1,0 +1,11 @@
+{
+  "compilerOptions": {
+    "target": "ES2022",
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
+    "strict": true,
+    "noEmit": true,
+    "skipLibCheck": true
+  },
+  "include": ["server-custom-plugin.ts"]
+}
--- a/scripts/integration.js
+++ b/scripts/integration.js
@@ -26,6 +26,7 @@ shell.exec(`pnpm add --config.minimum-release-age=0 ./${packed}`);
 shell.set('-e');
 
 // Test
+shell.exec('pnpm run typecheck');
 shell.exec('pnpm test');
 shell.exec('pnpm run build');
 shell.exec('node node-smoke/esm-test.mjs');
--- a/src/server/index.test.ts
+++ b/src/server/index.test.ts
@@ -52,7 +52,7 @@ jest.mock('socket.io', () => {
 });
 
 describe('new', () => {
-  test('accepts games with custom plugin APIs', () => {
+  test('processes custom-plugin and standard games through transport', () => {
     interface CustomPluginAPIs extends Record<string, unknown> {
       custom: { enabled: boolean };
     }
--- a/src/server/index.test.ts
+++ b/src/server/index.test.ts
@@ -15,7 +15,7 @@ import request from 'supertest';
 import { Server, createServerRunConfig, getPortFromServer } from '.';
 import type { KoaServer } from '.';
 import type { SocketIO } from './transport/socketio';
-import type { Game, StorageAPI } from '../types';
+import type { Game, Plugin, StorageAPI } from '../types';
 
 const game: Game = { seed: 0 };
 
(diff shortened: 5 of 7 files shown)


janmarsino98:
@Rupesh-ark @devill — the branch is rebased onto current main and CI/Coveralls are green. When you have a moment, could you please review? Thanks!

Rupesh-ark (changes requested):
Thanks @janmarsino98, the fix itself is correct. I verified it against a strict-mode consumer repro: with `Game[]` the exact error from #1050 appears on the `games:` line, and with `Game<any, any, any>[]` it compiles. Loosening `PluginAPIs` at the server boundary is safe since the server only runs games through `ProcessGameConfig` and never calls their plugin APIs, and a heterogeneous list can't be expressed generically anyway.

One real issue before this merges:

**The new test can't fail.** The root tsconfig has `"strict": false`, which disables `strictFunctionTypes`, so `setup` is checked bivariantly and the assignment in the test compiles with or without the fix. I confirmed this by reverting the `index.ts` change: `pnpm run ts` still passes with the test present. Jest can't catch it either, since babel strips types before running. So as written it documents the scenario but doesn't guard the regression.

Two changes I'd like:

1. Keep the test, but as runtime coverage: it does usefully verify that a mixed games list makes it through to `transport.init` with `processMove` attached. Please drop the "TypeScript regression test" framing from the PR description and test intent.
2. Add a guard that can actually fail. The integration harness (`scripts/integration.js`) already installs the packed tarball into a consumer fixture. A small `.ts` file there that imports `Server` from `boardgame.io/server` and passes it a custom-plugin game, checked with `tsc --noEmit` under a `"strict": true` tsconfig, reproduces the exact consumer setup from #1050 and will fail if this ever regresses. Happy for that to be this PR or a follow-up, whichever you prefer.

General note for regression tests: the one check missing from the validation list is confirming the test fails with the fix reverted. Worth making that a habit, especially when a tool wrote the test.

Nit, non-blocking: `Game<any, any>` is equivalent since `SetupData` already defaults to `any`. Fine either way.

janmarsino98:
@Rupesh-ark thanks for the review — addressed in `0a4f196`.

- Renamed the Jest case to make its runtime intent explicit: it verifies the custom-plugin and standard games are processed together through `transport.init`.
- Added a strict packed-consumer fixture using only `boardgame.io` and `boardgame.io/server` public APIs, and wired its typecheck into the integration run.
- Positive control passes with `Game<any, any, any>[]`.
- Negative control in an isolated worktree, reverting only `ServerOpts.games` to `Game[]`, fails at the `Server` call with `TS2322`: the `setup` context is incompatible because `custom` is missing.
- Build/test and integration are green on Node 24.x and 26.x; Coveralls is green at 100%.

@Rupesh-ark @devill, could you please take another look when convenient?

Rupesh-ark (approved):
LGTM

Issue #292, opened 2018-10-19
Illegal moves API


Izhaki:
Between the problem and proposed solution below, the former is more important - someone else might come up with a better solution.

# Background

Some moves are illegal. The logic behind an illegal move depends on any combination of these:

- `G`
- `ctx`
- Additional arguments passed

Note that it may be desired to know if a move is legal before actually attempting to dispatch it. Examples:

- A poker client disabling some buttons.
- A player getting an immediate feedback that a stone cannot be place in a particular position in a game of GO.

Also note that in the last example some *reason* might be useful ("Stone will have no liberties"). This is to highlight that just returning a boolean may not suffice.

# Currently

## Client

Here's an example from the docs:

```jsx
class TicTacToeBoard extends React.Component {
  onClick(id) {
    if (this.isActive(id)) {
      this.props.moves.clickCell(id);
      this.props.events.endTurn();
    }
  }

  isActive(id) {
    if (!this.props.isActive) return false;
    if (this.props.G.cells[id] !== null) return false;
    return true;
  }

  ...

}
```

But this doesn't prevent a hacker dispatching an illegal move to the server. So...

## Server

A move function that returns undefined denotes an illegal move:

```js
  moves: {
    clickCell(G, ctx, id) {
      if (G.cells[id] !== null) return undefined;
      ...
    },
  },
```

Note that you cannot invoke a move to check if it's illegal - if it is legal, it will be executed; so the check needs to live elsewhere.

In other words, the fact that one can return `undefined` from the move function fuses together two separate concerns:

- The execution of the move.
- The legal check.

# The issue

There's potential repetition here. Clean coders will extract the related logic into a function; something like:

```js
const isCellFree = (G, ctx, id) => G.cells[id] === null;
```

And will use this in both client (react) and game config.

This is a simple case. If such function needs to return a reason it may involve quite a few checks with some games.

# Proposed solution

I believe that the logic of weather a move is illegal should live as close as possible to the move definition itself.

So how about:

```js
  moves: {
    clickCell: {
      reduce: (G, ctx, id) => {
        const cells = [...G.cells];
        cells[id] = ctx.currentPlayer;
        return { ...G, cells };
      },
      isValid: (G, ctx, id) => G.cells[id] === null,
    },
  },
```

Notes:

- Internally, the framework will call `isValid` (if provided) before calling reduce.
- Perhaps allow both current and proposed versions (for cases where `isValid` is not needed).
[...]

PR #1274, opened 2026-07-18
feat: deliver move-rejection errors to the acting client (#723)

Rupesh-ark:
## Problem

When an action is rejected, the acting player receives no structured result: `INVALID_MOVE` silently rolls state back, and pre-dispatch failures such as stale state or an inactive player are also silent (#723, #592). Games work around this by duplicating validation on the client.

Addresses #723. Provides the transport foundation for a promise-based call-site API in #592. Related: #292.

### 1. Moves can attach a rejection reason

Moves may return `Invalid(payload)` from `boardgame.io/core`:

```ts
const recruit: Move = ({ G, playerID }, count) => {
  if (G.gold[playerID] < COST * count) {
    return Invalid({ code: 'NO_GOLD', message: 'Not enough gold' });
  }
  // ...
};
```

`return INVALID_MOVE` behaves exactly as before, with an `undefined` payload.

Invalid results use a registry symbol so separately built subpackages share the same runtime identity. The public `InvalidMoveResult` type also has a required private brand, so ordinary `{ payload: ... }` objects cannot be mistaken for invalid results by TypeScript.

The payload is extracted into transient error metadata before state is persisted or sent to other players.

### 2. The master returns a correlated action result

Each multiplayer action receives a client-local action ID. The master echoes it in a targeted `actionResult` containing either success or an `ActionError`.

This covers both reducer errors, including `Invalid(payload)` and plugin validation, and pre-dispatch errors such as stale state IDs, unavailable moves, inactive players, authentication failures, and missing matches.

The normal state update or rollback broadcast is unchanged. Only the acting client receives its action result and rejection payload.

### 3. Clients expose errors without duplicate or stale notifications

Multiplayer clients wait for the authoritative master result instead of surfacing the same optimistic rejection first. Results older than the client’s latest submitted action are ignored, so a delayed rejection cannot overwrite a newer action.

```ts
client.subscribe((state, error) => {
  if (error) {
    showToast(error.payload?.message ?? 'Invalid action');
  }
});
```

The second subscriber argument is a one-time rejection event: unrelated notifications pass `undefined`. `client.lastActionError` remains available for persistent inspection and is cleared when a subsequent action succeeds. React boards receive the same value as a `lastActionError` prop.

When a rejected move had been applied optimistically, the client requests a sync to restore the authoritative state. Without this, a client that pipelined moves ahead of a rejection stayed forked: rollback broadcasts carry a stateID below the optimistic local state and are ignored. This divergence predates this PR (reproducible on `main` whenever the master rejects an optimistically applied move); the correlated result channel is what makes the repair possible. The healing sync preserves `lastActionError` and the pending action ID, so a rejection survives the rollback it explains, and an action dispatched while the repair is in flight still receives its own result. An unsolicited sync, such as one following a reconnect, still clears both.

Single-player clients use the same public contract, reading their result directly from reducer transients.

## Backwards compatibility

- Old client + new server: old clients send no action ID, so the server preserves the previous silent-result behavior.
- New client + old server: the extra update argument is ignored and no action result is delivered.
- Persisted state, storage adapters, logs, and replays are unchanged; transient errors and action IDs are not persisted.
- Existing one-argument subscribers remain compatible.
- Custom transports remain source-compatible because the new `sendAction` argument is optional, but must forward the action ID and `actionResult` message to support rejection delivery.
- Plugin authors comparing move results with `result === INVALID_MOVE` should use the exported `isInvalidMoveResult()` to also recognise `Invalid(payload)`.

## Docs

Included:

- `Invalid(payload)` in the immutability guide.
- One-time rejection events in `subscribe((state, error))`.
- `client.lastActionError` and the React `lastActionError` board prop.

## Tests and coverage

Coverage includes:

- Reducer payload threading, including Immer mutation rollback.
- Stable invalid-result identity across separately built bundles.
- Master targeting and correlated success/error results.
- Pre-dispatch rejection results.
- Local, multiplayer, synchronous-transport, and stale-result client lifecycles.
- Error clearing after a later successful move or undo.
- Socket.IO action ID forwarding and action-result delivery.
- Legacy `INVALID_MOVE` behavior.
- Seeded interleaving fuzz over a virtual per-channel FIFO network (real client, master, and storage): exactly-once targeted results, convergence with the stored master state, ledger consistency, stale-result and error-event discipline; 25 seeds in CI, `STRESS_SEEDS=<n>` to soak. Directed races cover cell contention, pipelined rejection, result-before-broadcast skew, sync racing an in-flight result, reentrant dispatch, and the optimistic self-heal.
- Real Socket.IO end-to-end stress: contention bursts, cross-match isolation, two tabs for one player, old-client compatibility, and disconnect mid-flight.

## Follow-ups

- Strongly typed error payloads (`unknown` instead of `any`; see #732).
- Promise-based call-site API for #592, layered on the correlated action-result channel.

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -175,11 +175,12 @@ export class _ClientImpl<
   chatMessages: ChatMessage[];
   /**
    * Why this client’s most recent action was rejected, or undefined.
-   * Cleared when a subsequent action succeeds.
+   * Cleared when a subsequent action succeeds, or on reset / sync.
    */
   lastActionError?: ActionError;
   private nextActionID = 0;
   private latestActionID?: number;
+  private actionResultDeliveries = 0;
 
   constructor({
     game,
@@ -316,16 +317,33 @@ export class _ClientImpl<
      * Single-player clients read action errors directly from reducer
      * transients. Multiplayer clients wait for the master's correlated result.
      */
+    let pendingActionError: ActionError | undefined;
     const SubscriptionMiddleware =
       (store: Store) => (next: Dispatch<Action>) => (action: Action) => {
+        const deliveriesBefore = this.actionResultDeliveries;
         const result = next(action);
-        let actionError: ActionError | undefined;
-        if (!this.multiplayer && action.type !== Actions.STRIP_TRANSIENTS) {
+
+        // Notify after transient metadata has been stripped, so subscribers
+        // receive one rejection event paired with clean public state.
+        if (action.type === Actions.STRIP_TRANSIENTS) {
+          const actionError = pendingActionError;
+          pendingActionError = undefined;
+          if (actionError) this.notifySubscribers(actionError);
+          return result;
+        }
+
+        if (action.type === Actions.RESET) {
+          this.lastActionError = undefined;
+          this.latestActionID = undefined;
+        }
+
+        if (!this.multiplayer) {
           const transientError = (store.getState() as TransientState<G> | null)
             ?.transients?.error;
           if (transientError) {
             this.lastActionError = transientError;
-            actionError = transientError;
+            pendingActionError = transientError;
+            return result;
           } else if (
             action.type === Actions.MAKE_MOVE ||
             action.type === Actions.GAME_EVENT ||
@@ -335,7 +353,8 @@ export class _ClientImpl<
             this.lastActionError = undefined;
           }
         }
-        this.notifySubscribers(actionError);
+        if (this.actionResultDeliveries !== deliveriesBefore) return result;
+        this.notifySubscribers();
         return result;
       };
 
@@ -390,7 +409,10 @@ export class _ClientImpl<
 
   /** Handle the authoritative result of an action sent to the master. */
   private receiveActionResult(actionID: number, error?: ActionError): void {
+    // Only the latest submitted action owns lastActionError. Older results are
+    // stale once the player has attempted another action.
     if (actionID !== this.latestActionID) return;
+    this.actionResultDeliveries++;
     this.lastActionError = error;
     this.notifySubscribers(error);
   }
@@ -401,6 +423,10 @@ export class _ClientImpl<
     if (matchID !== this.matchID) return;
     switch (data.type) {
       case 'sync': {
+        // Transient errors are not persisted by the master. A sync establishes
+        // a fresh authoritative baseline and invalidates any in-flight result.
+        this.lastActionError = undefined;
+        this.latestActionID = undefined;
         const [, syncInfo] = data.args;
         const action = ActionCreators.sync(syncInfo);
         this.receiveMatchData(syncInfo.filteredMetadata);
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -181,6 +181,7 @@ export class _ClientImpl<
   private nextActionID = 0;
   private latestActionID?: number;
   private actionResultDeliveries = 0;
+  private latestAuthoritativeStateID = -1;
 
   constructor({
     game,
@@ -415,6 +416,14 @@ export class _ClientImpl<
     this.actionResultDeliveries++;
     this.lastActionError = error;
     this.notifySubscribers(error);
+    // Broadcasts older than the optimistic state are ignored, so a rejected
+    // optimistic move can only be rolled back by an explicit resync.
+    if (
+      error &&
+      this.store.getState()._stateID > this.latestAuthoritativeStateID
+    ) {
+      this.transport.requestSync();
+    }
   }
 
   /** Handle all incoming updates from a multiplayer transport. */
@@ -428,6 +437,7 @@ export class _ClientImpl<
         this.lastActionError = undefined;
         this.latestActionID = undefined;
         const [, syncInfo] = data.args;
+        this.latestAuthoritativeStateID = syncInfo.state._stateID;
         const action = ActionCreators.sync(syncInfo);
         this.receiveMatchData(syncInfo.filteredMetadata);
         this.store.dispatch(action);
@@ -435,6 +445,10 @@ export class _ClientImpl<
       }
       case 'update': {
         const [, state, deltalog] = data.args;
+        this.latestAuthoritativeStateID = Math.max(
+          this.latestAuthoritativeStateID,
+          state._stateID,
+        );
         const currentState = this.store.getState();
         if (state._stateID >= currentState._stateID) {
           const action = ActionCreators.update(state, deltalog);
@@ -444,6 +458,10 @@ export class _ClientImpl<
       }
       case 'patch': {
         const [, prevStateID, stateID, patch, deltalog] = data.args;
+        this.latestAuthoritativeStateID = Math.max(
+          this.latestAuthoritativeStateID,
+          stateID,
+        );
         const currentStateID = this.store.getState()._stateID;
         if (prevStateID !== currentStateID) break;
         const action = ActionCreators.patch(
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -178,12 +178,8 @@ export class _ClientImpl<
    * Cleared when a subsequent action succeeds.
    */
   lastActionError?: ActionError;
-  /**
-   * Count of action errors delivered via receiveActionError. Lets
-   * SubscriptionMiddleware detect a rejection that a synchronous transport
-   * delivered during `next(action)` and avoid clearing it on the same pass.
-   */
-  private actionErrorDeliveries = 0;
+  private nextActionID = 0;
+  private latestActionID?: number;
 
   constructor({
     game,
@@ -300,13 +296,16 @@ export class _ClientImpl<
     const TransportMiddleware =
       (store: Store) => (next: Dispatch<Action>) => (action: Action) => {
         const baseState = store.getState();
+        const shouldSend =
+          !('clientOnly' in action) && action.type !== Actions.STRIP_TRANSIENTS;
+        const actionID =
+          shouldSend && this.multiplayer
+            ? (this.latestActionID = ++this.nextActionID)
+            : undefined;
         const result = next(action);
 
-        if (
-          !('clientOnly' in action) &&
-          action.type !== Actions.STRIP_TRANSIENTS
-        ) {
-          this.transport.sendAction(baseState, action);
+        if (shouldSend) {
+          this.transport.sendAction(baseState, action, actionID);
         }
 
         return result;
@@ -314,31 +313,29 @@ export class _ClientImpl<
 
     /**
      * Middleware that intercepts actions and invokes the subscription callback.
-     * Also tracks action errors: the store still holds this action's transients
-     * here (TransientHandlingMiddleware strips them after we run), and the
-     * deliveries counter keeps the clear-on-success branch from wiping an
-     * error that a synchronous transport delivered during `next(action)`.
+     * Single-player clients read action errors directly from reducer
+     * transients. Multiplayer clients wait for the master's correlated result.
      */
     const SubscriptionMiddleware =
       (store: Store) => (next: Dispatch<Action>) => (action: Action) => {
-        const deliveriesBefore = this.actionErrorDeliveries;
         const result = next(action);
-        if (action.type !== Actions.STRIP_TRANSIENTS) {
+        let actionError: ActionError | undefined;
+        if (!this.multiplayer && action.type !== Actions.STRIP_TRANSIENTS) {
           const transientError = (store.getState() as TransientState<G> | null)
             ?.transients?.error;
           if (transientError) {
             this.lastActionError = transientError;
+            actionError = transientError;
           } else if (
-            (action.type === Actions.MAKE_MOVE ||
-              action.type === Actions.GAME_EVENT ||
-              action.type === Actions.UNDO ||
-              action.type === Actions.REDO) &&
-            this.actionErrorDeliveries === deliveriesBefore
+            action.type === Actions.MAKE_MOVE ||
+            action.type === Actions.GAME_EVENT ||
+            action.type === Actions.UNDO ||
+            action.type === Actions.REDO
           ) {
             this.lastActionError = undefined;
           }
         }
-        this.notifySubscribers();
+        this.notifySubscribers(actionError);
         return result;
       };
 
@@ -391,11 +388,11 @@ export class _ClientImpl<
     this.notifySubscribers();
   }
 
-  /** Handle an action error sent to this client by a multiplayer master. */
-  private receiveActionError(error: ActionError): void {
-    this.actionErrorDeliveries++;
+  /** Handle the authoritative result of an action sent to the master. */
+  private receiveActionResult(actionID: number, error?: ActionError): void {
+    if (actionID !== this.latestActionID) return;
     this.lastActionError = error;
-    this.notifySubscribers();
+    this.notifySubscribers(error);
   }
 
   /** Handle all incoming updates from a multiplayer transport. */
@@ -446,16 +443,16 @@ export class _ClientImpl<
         this.receiveChatMessage(chatMessage);
         break;
       }
-      case 'actionError': {
-        const [, error] = data.args;
-        this.receiveActionError(error);
+      case 'actionResult': {
+        const [, actionID, result] = data.args;
+        this.receiveActionResult(actionID, result.error);
         break;
       }
     }
   }
 
-  private notifySubscribers() {
-    this.subscribers.forEach((fn) => fn(this.getState(), this.lastActionError));
+  private notifySubscribers(actionError?: ActionError) {
+    this.subscribers.forEach((fn) => fn(this.getState(), actionError));
   }
 
   previewState(state: any) {
@@ -498,7 +495,7 @@ export class _ClientImpl<
     this.transport.subscribeToConnectionStatus(() => this.notifySubscribers());
 
     if (this._running || !this.multiplayer) {
-      fn(this.getState(), this.lastActionError);
+      fn(this.getState(), undefined);
     }
 
     // Return a handle that allows the caller to unsubscribe.
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -174,11 +174,16 @@ export class _ClientImpl<
   sendChatMessage: (message: any) => void;
   chatMessages: ChatMessage[];
   /**
-   * The most recent action error affecting this client, or undefined.
-   * Set when one of this client’s actions is rejected (locally or by
-   * the master) and cleared when a subsequent action succeeds.
+   * Why this client’s most recent action was rejected, or undefined.
+   * Cleared when a subsequent action succeeds.
    */
   lastActionError?: ActionError;
+  /**
+   * Count of action errors delivered via receiveActionError. Lets
+   * SubscriptionMiddleware detect a rejection that a synchronous transport
+   * delivered during `next(action)` and avoid clearing it on the same pass.
+   */
+  private actionErrorDeliveries = 0;
 
   constructor({
     game,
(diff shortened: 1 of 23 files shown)


src/plugins/plugin-immer.ts
```diff
@@ -20,16 +21,16 @@ const ImmerPlugin: Plugin = {
...
     (context, ...args) => {
-      let isInvalid = false;
+      let invalidResult: typeof INVALID_MOVE | InvalidMoveResult | undefined;
       const newG = produce(context.G, (G) => {
         const result = move({ ...context, G }, ...args);
-        if (result === INVALID_MOVE) {
-          isInvalid = true;
+        if (isInvalidMoveResult(result)) {
+          invalidResult = result;
           return;
```
devill, line 45: `Invalid(...)` is built inside the `produce` recipe, so a payload that references `G` seems to
escape holding revoked drafts:

```ts
moves: { play: ({ G }) => Invalid({ card: G.deck[0] }) }
// TypeError: Cannot perform 'Object.prototype.toString' on a proxy that has been revoked
```

Under `NODE_ENV=production` it's quieter — `plugin-serializable` is skipped, dispatch succeeds, and
the revoked proxy sits in `lastActionError.payload` until something reads it, single-player
included.

As far as I can tell the boundary is: unsafe if the payload references an object or array reached
through `G`; primitives from `G` and freshly-built objects are fine.

One way to handle it, snapshotting drafts before they escape:

```ts
import { produce, current, isDraft } from 'immer';
import { isInvalidMoveResult, Invalid, INVALID_MOVE } from '../core/constants';

const snapshot = (value: unknown): unknown => {
  if (isDraft(value)) return current(value as any);
  if (Array.isArray(value)) return value.map((item) => snapshot(item));
  if (value && typeof value === 'object')
    return Object.fromEntries(
      Object.entries(value).map(([k, v]) => [k, snapshot(v)]),
    );
  return value;
};

// in fnWrap:
if (isInvalidMoveResult(result)) {
  invalidResult =
    result === INVALID_MOVE ? result : Invalid(snapshot(result.payload));
  return;
}
```

With that, `Invalid({ card: G.deck[0], n: G.n })` gives `{"card":{"id":"ace"},"n":7}` and the suite
stays green. A test pinning it would be good if you go this route.

Small related thing: `Invalid(new Error(...))` trips `Move state is not JSON-serialiazable.`, which
points at `G` rather than the payload — might be worth a note on the JSDoc at
`src/core/constants.ts:21-26` about what can go in a payload.

src/client/client.ts
```diff
@@ -284,25 +296,46 @@ export class _ClientImpl<
...
+          } else if (
+            action.type === Actions.MAKE_MOVE ||
+            action.type === Actions.GAME_EVENT ||
+            action.type === Actions.UNDO ||
+            action.type === Actions.REDO
+          ) {
+            this.lastActionError = undefined;
+          }
+        }
+        this.notifySubscribers(actionError);
```
devill: `TransientHandlingMiddleware` (`src/core/reducer.ts:241-249`) dispatches `stripTransients()` after
`next()`, which re-enters this middleware and notifies a second time. For one rejected move I see:

```
[ {type: 'action/invalid_move', payload: {...}}, undefined ]
```

`if (error) showToast(...)` is fine with that, but `setError(error)` lands on `undefined`.
Multiplayer notifies once, so the two modes end up with slightly different contracts.

Suggested fix — skip the notify on the strip. I also folded in clearing on `RESET`, since
`reset()` currently leaves the previous error in place on a new game, and it's the same conditional:

```suggestion
            action.type === Actions.REDO ||
            action.type === Actions.RESET
          ) {
            this.lastActionError = undefined;
          }
        }
        if (action.type === Actions.STRIP_TRANSIENTS) return result;
        this.notifySubscribers(actionError);
```

Sequence becomes `[error]`, `reset()` clears, suite stays green.

Heads-up on the test at `src/client/client.test.ts:686-698` — it filters `error !== undefined`
before asserting the length, so it passes either way. Asserting the whole sequence would pin the
behaviour you want.

devill (commented):
Thanks for this — this is a genuinely useful feature.

Fair warning: since this was a large PR I used Claude (Opus and Fable) to review. 
I've confirmed the findings to the best of my ability, but still, if anything
below reads as confidently wrong, it probably is; push back. We also found two 
pre-existing issues, noted here, since they are small fixes, but I'll approve 
even if those are not fixed. 

Having said that I have questions around two potential issues, and some smaller things.

1. `Invalid(payload)` throws when the payload references an object or array read out of `G`
2. In single-player, the error handed to `subscribe` is followed by a trailing `undefined` in the
   same dispatch, so any consumer that *stores* the argument (`setError(error)`) ends up with
   nothing.

I've put a suggested fix under each — both worked when I tried them here, but you'll know better
whether they fit what you had in mind.

One small thing: the description mentions an `actionError` message and lists pre-dispatch errors
as a follow-up, but the code looks like it ships `actionResult` with a correlation ID and already
handles the pre-dispatch cases. Did the design change along the way? Just want to make sure I'm
reading the intent right.

**Process request**: Agents tend to commit multiple things in one commit, which makes it hard to review. After the
implementation pass, please ask the subagent to re-cut the branch into commits that each do one
thing. It makes the review easier, and since it also serves as an agentic review pass, it can help
catch issues before the PR. For this one I did that locally — happy to share if it helps.

---

## A few smaller things

**A move that throws leaves the caller without a result.** *(pre-existing — `onUpdate` has never
had a `try`/`catch`, and `src/server/transport/socketio.ts:206` doesn't catch the rejection
either.)* Not something this PR broke, but it becomes more visible now that a client can be
waiting on a result: nothing is sent and it waits indefinitely. Wrapping the body and sending
`ActionErrorType.ActionInvalid` on a throw would close the loop, if you feel like picking it up —
equally fine as a follow-up.

**Success is reported just before it's persisted.** `src/master/master.ts:360` sends the result
before `storageAPI.setState` at `:379` — though `sendAll` already broadcast pre-persist on `main`,
so this follows the existing pattern rather than introducing anything. Mentioning it only in case
you'd rather the new result wait for the write.

**Multiplayer doesn't clear the error on sync.** `src/client/client.ts:392-396` — only
`receiveActionResult` writes `lastActionError`, so if that frame is lost (or the server is older)
it stays until the next result. Clearing it in the `sync` case at `:410` would cover reconnects.

**`Invalid()` from `turn.onMove`.** *(pre-existing — `src/core/flow.ts` is untouched here.)* The
hook's return is assigned straight to `G` at `:751`, so it becomes state rather than an error.
Already true of `INVALID_MOVE`, but the new API makes it a more tempting thing to try — a
"moves only" line in the docs would probably save someone an afternoon.

**Two rapid moves.** `src/client/client.ts:393` keeps only the newest action, so if move 1 is
rejected and move 2 succeeds, the rejection isn't surfaced. "Latest wins" is a reasonable choice —
is that the intent? If so a short comment would help, since `lastActionError` reads like it never
drops one.

**Error codes in the docs.** `Client.md` lists the `action/*` codes, but the master also sends
`update/unauthorized_action` and `update/match_not_found`, so consumers will see `ErrorType`
(`src/types.ts:32`) rather than just `ActionErrorType`. Worth mentioning in the doc.

---

## Tests worth adding

- `master.ts:182` (`actionID === undefined`) is what keeps old clients working — nothing covers it.
- Of the new pre-dispatch result sites, only `StaleStateId` is tested.
- The `lastActionError` board prop (`src/client/react.tsx:187`).
- `Invalid()` with no payload, and non-object payloads.

Rupesh-ark:
Refreshed this against current `main` and recut the branch into four focused commits.

The July review findings are addressed:

- `Invalid(...)` snapshots Immer drafts before they escape, with regression coverage.
- Single-player subscribers emit one rejection notification, and reset/sync clearing is covered.
- The docs now describe moves-only use, latest-result semantics, and both `action/*` and `update/*` errors.
- Compatibility, React props, pre-dispatch results, optimistic resync, Socket.IO targeting, and concurrency races have coverage.

tbsvttr (commented):
Reviewed as an outside contributor with no merge rights — input for @Rupesh-ark and @devill, not an approval.

## What I ran

On `rupesh-ark/invalid-move-payload` (4 commits on top of `5e9a2c9`): lint, `tsc --noEmit`, the full suite (45 suites / 966 tests) and the build. All green. Everything below comes from throwaway probes run against this branch and deleted again; no tracked file was touched.

I re-checked the two July findings first.

**Finding A (Immer drafts, `src/plugins/plugin-immer.ts`) — fixed.** `snapshotDrafts()` now runs over `result.payload` inside the produce recipe. @devill's exact repro `moves: { play: ({ G }) => Invalid({ card: G.deck[0] }) }` yields `payload === {"card":{"id":"ace","details":{"suit":"spades"}}}`, `JSON.stringify` succeeds, and it behaves the same under `NODE_ENV=production` where plugin-serializable is a no-op. Deeper nesting (`{a:{b:[{c:G.deck[0]}]}}`, `[[G.deck[0]]]`) and a draft behind an own enumerable getter are snapshotted too. I could not get a revoked proxy out through a plain object or an array.

One thing worth knowing rather than fixing: `current()` returns the base object verbatim for an unmodified draft, so `client.lastActionError.payload.card === client.getState().G.deck[0]` is literally `true`. That is harmless only because immer's auto-freeze is unconditional and nothing here calls `setAutoFreeze(false)`.

**Finding B (double notification, `src/client/client.ts`) — fixed, and the weak test is gone.** A single-player client with `moves: { bad: () => Invalid({reason:'nope'}) }` now produces exactly one subscriber call carrying the error, with no trailing `undefined`. The `filter(error !== undefined)` @devill flagged has been replaced by `expect(fn.mock.calls.map(([, error]) => error)).toEqual([expectedError])`, which does fail on `[error, undefined]`. `reset()`, `undo()`, `redo()` and a later successful move all clear the error.

The two modes still differ in call *shape* — single player gives `[error]`, multiplayer against a synchronous `Local()` master gives `[undefined, error]` — but both deliver exactly one error event, so `if (error) …` and `setError(error)` are correct in both. Not filing that.

Both findings below share one root cause: `receiveActionResult` fires `requestSync()` at `client.ts:425` right after publishing the error, and the `sync` handler unconditionally wipes it at `client.ts:437-438`.

---

## 1. The repair sync clears `lastActionError` before a React board can render it

`client.ts:417-426` sets `lastActionError`, notifies subscribers, then calls `requestSync()`. `client.ts:437` clears it again. On `Local()` there are no awaits in the path — `LocalMaster` has no auth and `InMemory` is synchronous — so the whole heal runs *inside* `client.moves.X()`, and the field is already `undefined` when the call returns.

React only ever reads the persistent field: `react.tsx:141` subscribes with `() => this.forceUpdate()`, discarding the error argument, and `react.tsx:187` passes `this.client.lastActionError` down. So for the one class of rejection the heal exists for — a move the client applies optimistically and the master rejects — the board prop is never set.

```js
const game = {
  setup: () => ({ secret: 'x', played: 0 }),
  playerView: ({ G }) => ({ ...G, secret: null }),
  moves: {
    play: ({ G }) => { if (G.secret === 'x') return Invalid({ reason: 'secret blocks' }); G.played++; },
    bad:  () => Invalid({ reason: 'nope' }),   // control: rejected on both sides
  },
};

// A) plain client on Local(), subscriber pushes [state._stateID, error?.payload]
c.moves.play();
// observed  [[1, {reason:'secret blocks'}], [1, undefined], [0, undefined]]
// observed  c.lastActionError immediately after the call returns:  undefined

// B) React client, same game, act(() => props.moves.play()), recording every render
// observed  ["no-error"]
// expected  at least one render with props.lastActionError set

// C) control, act(() => props.moves.bad())
// observed  [{reason:'nope'}]        // works
```

State itself is repaired correctly (`_stateID` back to 0), so this is a lost signal rather than a divergence — the player sees the move revert with no explanation.

The caveat *is* documented, but only in the Plain JS tab (`Client.md:205`), and it points readers at the `subscribe` error argument, which React boards cannot reach — `BoardProps` exposes no client and no subscribe. Meanwhile the React tab this PR adds at `Client.md:413-423` gives a `useEffect(..., [props.lastActionError])` toast recipe with no caveat, and `immutability.md:97-99` says the payload can be read from "the `lastActionError` board prop" and "remains set until a later action succeeds". For this class it never appears at all.

Not pre-existing: `git show origin/main:src/client/client.ts | grep -c lastActionError` → `0`.

Either the repair sync should not invalidate the error it was dispatched to repair, or the React wrapper needs to surface the one-time event — the subscriber callback already receives it and `react.tsx:141` throws it away.

## 2. A move submitted while a repair sync is in flight has its rejection swallowed

`receiveActionResult` drops anything not matching `latestActionID` (`client.ts:415`), and the sync handler zeroes `latestActionID` (`client.ts:438`). Since the heal fires a sync on every optimistically-diverged rejection, there is a one-RTT window in which a further action's `actionResult` arrives *after* the sync — channels are FIFO — and is discarded. No subscriber event, `lastActionError` stays `undefined`.

Reproduced end to end against a real `Master` + `InMemory` + `getFilterPlayerView` behind a FIFO network with explicit flushes:

```js
client.moves.m(); flushServer(); await tick; flushClient();   // result 1
client.moves.m(); flushServer(); await tick; flushClient();   // sync, then result 2

// after 1:  events [.., {reason:'server says no'}]   lastActionError set
// master rejects move 2:  invalid stateID, was=[1], expected=[0]   (master.ts:324)
// after 2:  events [.., {reason:'server says no'}, undefined, undefined, undefined]
//           lastActionError undefined,  stateID 0,  G {"n":0}
```

Scope, so this isn't over-read: it needs a **server-only** rejection (secret-state validation, stale stateID, server-side inactive-player or auth). For a move that is `Invalid()`/`INVALID_MOVE` on both sides the client's own reducer rejects it too, `_stateID` never advances, `requestSync()` is never called, and two rapid clicks both report cleanly. With three rapid moves the one-time event for the newest action does still fire.

Same root cause as finding 1 — one fix covers both.

---

## Coverage notes, not defects

- `multiplayer-fuzz.test.ts:640-647` already encodes the invariant finding 2 violates (it allows a sync *after* the result to clear the error; here the sync arrives *before* it). 256 seeds never generate that interleaving — worth a deterministic case rather than more seeds.
- `client.test.ts:770` uses `toHaveBeenCalledWith(expect.anything(), error)`, which would not catch a double delivery on the multiplayer path. The strict `mock.calls.map(...)` form used in the single-player and synchronous-transport tests would.
- No React test covers `props.lastActionError` for a master-rejected optimistic move; one would have caught finding 1.

## What I went after and could not break

- **`Invalid()` identity.** A 16-case accept/reject matrix, including a cross-realm object built in a `vm` context, a null-prototype carrier, string-named `"Symbol(...)"` keys, a same-description `Symbol()`, and JSON round-tripped data — wire data cannot forge a rejection. The registry symbol earns its place: the build emits three physical copies of the constants module, and a game returning `core.Invalid()` from `dist/cjs/core.js` played through a `Client` from `dist/boardgameio.js` works.
- **Falsy payloads.** `Invalid(undefined | null | 0 | '' | false | NaN)` and bare `INVALID_MOVE` all keep the key present, so "rejected with no payload" stays distinguishable from "not rejected" in both modes.
- **Transient stripping.** After a rejected move, no `transients` key and no payload substring in persisted state, the stored log, the `sendAll` broadcast, or anything handed to subscribers. Replaying the log from `initialState` reproduces `G` and `_stateID` exactly, and rollback is reference-identical.
- **Master return paths.** All 14 exits from `onUpdate` send exactly one `actionResult` with the right type, after the broadcast. No path leaves a client waiting.
- **Targeting.** Against the real server `TransportAPI` + `InMemoryPubSub` with two sockets, the acting socket got `["update","actionResult"]` and the other `["update"]`, with no trace of the hidden payload. Two clients acting concurrently both use local `actionID` 1 and receive only their own result.
- **Old client + new server.** A 4-argument `onUpdate` produces zero `send` calls across rejected, stale, inactive-player, malformed and successful actions. A transport subclass with the old 2-argument `sendAction` type-checks and runs.
- **No resync loop.** SYNC is `clientOnly` and re-arms the guard immediately; five consecutive diverging rejections produce exactly six syncs, linear.
- **Stale-result arithmetic.** IDs are pre-incremented, monotonic and never reused. Duplicate, out-of-order, error-then-success, success-then-error and future-ID results were all walked; none can clear or overwrite a newer error.

Two operational notes, neither a defect: the acting client now takes one extra frame and one extra subscriber notification per *successful* action, and on the socket.io server every `sync` runs `removeClient`/`addClient`, writing match metadata and broadcasting `matchData` to everyone — so each diverging rejection now costs that too.

Rupesh-ark:
@tbsvttr thank you, this is an exceptional review. Both findings are real and both are fixed in 86e00ab.

You were right that they share a root cause, and one change closes both. `receiveActionResult` now marks the sync it requests for itself, and the sync handler keeps `lastActionError` and `latestActionID` when that mark is set. An unsolicited sync, for example after a reconnect, still clears both as before.

**Finding 1.** With the repair no longer erasing the error, the persistent field is still set when React renders, so `props.lastActionError` arrives for the master-rejected optimistic case. I took your first option rather than surfacing the one-time event through the wrapper, since it also keeps the plain JS and React contracts identical. New test in `react.test.tsx` using your secret-state repro shape; it fails on the parent commit.

**Finding 2.** Preserving `latestActionID` across the repair closes the one-RTT window. Deterministic case added at `multiplayer-fuzz.test.ts` (`a rejection arriving behind a repair sync is still reported`), driving a real `Master` + `InMemory` with the sync flushed ahead of the second result, plus a unit-level version in `client.test.ts`.

**Fuzz invariant.** You were right that `multiplayer-fuzz.test.ts:640-647` encoded the rule this violated. It now replays the preserve/clear decision over the delivery timeline instead of assuming any sync after a result clears, with the harness recording which syncs the client asked for. Soaked at `STRESS_SEEDS=400`, green.

**Other notes.** Tightened the assertion you flagged in `client.test.ts` to the strict `mock.calls.map(...)` form. Corrected the `Client.md` caveat, which described the old clearing behaviour and pointed React users at `subscribe`.

On your two operational notes: both stand and neither changes with this fix. The extra frame per successful action is inherent to waiting for the authoritative result, and the `removeClient`/`addClient` churn per sync is worth its own issue against the socket.io transport rather than more scope here.

Full suite, lint, typecheck and build are green on `d5212d9`.

tbsvttr:
Re-checked `d5212d9`. Both findings are fixed, and I confirmed the new tests are real guards rather than tests written against the new behaviour — I took each one back to `ea039cf` with the old `client.ts` and they all fail there:

- `client.test.ts` → `the repair sync keeps the rejection that requested it`, `a move sent while the repair sync is in flight still reports` (2 failed)
- `react.test.tsx` → `board receives lastActionError for a master-rejected optimistic move` (1 failed)
- `multiplayer-fuzz.test.ts` → `a rejection arriving behind a repair sync is still reported` (1 failed)

My own repros now come out clean. Finding 1, plain client on `Local()`, secret-state game:

```
events              [[1,'secret blocks'], [1,undefined], [0,undefined]]
lastActionError     {reason:'secret blocks'}     // was undefined
_stateID after      0                            // still repaired
React board renders ['secret blocks']            // was ['no-error']
control (both-sides rejection) ['nope']          // unchanged
```

Full suite green here too: 45 suites / 971 tests.

## One case the boolean does not cover

`healSyncPending` is a boolean consumed by the next sync, whichever sync that is. Two optimistically-diverged rejections inside one RTT request two repairs, the first sync consumes the flag, and the second takes the unsolicited branch and clears an error that is still current:

```js
client.moves.valid();                                    // action 1
sendToClient({type:'actionResult', args:['A',1,{error:first}]});
client.moves.valid();                                    // action 2, before either sync lands
sendToClient({type:'actionResult', args:['A',2,{error:second}]});
sendToClient({type:'sync', args:['A',{state}]});         // repair 1
sendToClient({type:'sync', args:['A',{state}]});         // repair 2

// requestSync calls  2
// before syncs   lastActionError {reason:'second'}
// after sync 1   lastActionError {reason:'second'}   correct
// after sync 2   lastActionError undefined           expected {reason:'second'}
// subscriber events  ['first','second']              both fired, so the one-time signal is intact
```

Same shape as finding 1 — the persistent field and therefore the React prop are lost — just needing two server-only rejections in flight rather than one. Counting outstanding repairs instead of flagging one would close it:

```ts
private healSyncsPending = 0;
// in receiveActionResult, before requestSync():
this.healSyncsPending++;
// in the sync handler:
if (this.healSyncsPending > 0) this.healSyncsPending--;
else { this.lastActionError = undefined; this.latestActionID = undefined; }
```

with `RESET` zeroing it rather than setting `false`.

Worth being straight about the limit of that too: a counter still cannot tell *which* sync is which, so an unsolicited sync landing while a repair is outstanding is still attributed to the repair. Correlating syncs properly would mean an ID on the request, which looks like more than this PR should carry. The counter at least makes the common case — a player clicking faster than the round trip — behave the same as the single-rejection case you have just fixed.

Agreed on both operational notes, including that the `removeClient`/`addClient` churn per sync belongs in its own issue against the socket.io transport.

PR #1276, opened 2026-07-19
Fix flaky server tests: bind test servers to 127.0.0.1

devill:
## Problem

`src/server/api.test.ts` failed ~8% of local runs with `Parse Error: Expected HTTP/`, wrong statuses, or truncated bodies — a different test failing each time.

Root cause: supertest creates a server per request with host-less `listen(0)`, which binds the **wildcard** address. The kernel may hand out an ephemeral port that another local process already holds on `127.0.0.1` (observed here with IntelliJ's built-in web server and Ollama — both binds succeed and coexist). supertest then dials `127.0.0.1:port`, and the kernel routes the connection to the most-specific bind: the foreign process. Depending on who answers, you get their 404 page, an unrelated 200, or non-HTTP bytes. Nothing ever fails loudly, which is what made this a heisenbug — and why CI (no IDE listening on localhost) never saw it.

## Fix

The invariant is *bind address == dial address*:

- **api.test.ts**: one suite-level `http.Server` bound explicitly to `127.0.0.1`, delegating to the current test's Koa callback. supertest receives the already-listening server, so it never binds anything itself. (Note: the bind must be awaited — `listen(port, host)` resolves the host asynchronously, and supertest silently re-listens on the wildcard when handed a non-listening server.) Also drops a leftover `jest.setTimeout(2_000_000_000)`.
- **index.test.ts** had the same hazard plus fixed ports (8888, 9999, 57890/57891 — the last two inside the macOS ephemeral range, so any transient outgoing connection can occupy them). Tests now use ephemeral ports bound to `127.0.0.1`, enabled by two small `Server.run` changes:
  - a new optional `host` in the run config, passed to all listen calls (documented in `docs/documentation/api/Server.md`);
  - `apiPort: 0` now means "ephemeral API port" — previously the falsy check silently ignored it (`null`/`undefined` behave as before).

## Verification

- Reproduced pre-fix at 2/20 and 3/40 suite runs; instrumentation correlated failing requests to the foreign listeners by socket port.
- Post-fix: 110 consecutive clean runs across the two suites.
- Full suite green (42 suites / 883 tests), lint clean, `src/server/index.ts` stays at 100% coverage.

#### Checklist

https://claude.ai/code/session_01KNK7XYPXjxHpKJUCL52kVt

### DIFF
--- a/docs/documentation/api/Server.md
+++ b/docs/documentation/api/Server.md
@@ -119,6 +119,10 @@ boardgame.io `port`.
 - `apiCallback`: Called when the Koa server is ready. Only applicable if
 `apiPort` is specified.
 
+The run config also accepts a `host` alongside `port`, which binds both the
+game server and the Lobby API server to that host. It defaults to all
+interfaces.
+
 #### With HTTPS
 
 ```js
--- a/src/server/api.test.ts
+++ b/src/server/api.test.ts
@@ -26,8 +26,6 @@ import { InMemory } from './db/inmemory';
 import { Origins } from './cors';
 import type { Game, Server } from '../types';
 
-jest.setTimeout(2_000_000_000);
-
 beforeEach(() => {
   dateMock.clear();
 });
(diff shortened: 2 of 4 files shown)


Rupesh-ark (approved):
LGTM!

PR #1277, opened 2026-07-20
docs(changelog): classify v0.43.3 serializability check as breaking

### LINKED ISSUE #1187: Mention breaking change for version 0.43.3 in changelog
It seems like the [check for non-serializable states](https://github.com/boardgameio/boardgame.io/pull/896) was introduced in v0.43.3.

This should be mentioned as a breaking change in the changelog.

I stumbled upon this when updating boardgame.io for my game (which is client-side only).

Rupesh-ark:
## Summary

- classify the v0.43.3 non-serializable-state enforcement as a breaking change
- keep the LobbyClient exception-detail entry under features

Closes #1187

## Verification

- `pnpm exec prettier --check docs/documentation/CHANGELOG.md`
- `git diff --check`

### DIFF
--- a/docs/documentation/CHANGELOG.md
+++ b/docs/documentation/CHANGELOG.md
@@ -362,9 +362,12 @@ See [the Server reference page](https://boardgame.io/documentation/#/api/Server)
 
 ### v0.43.3
 
-#### Features
+#### Breaking Changes
 
 * [[01c522c](https://github.com/boardgameio/boardgame.io/commit/01c522c)] Throw error in development if non-serializable state is used in a move ([#896](https://github.com/boardgameio/boardgame.io/pull/896))
+
+#### Features
+
 * [[ccc9ada](https://github.com/boardgameio/boardgame.io/commit/ccc9ada)] Add details to exceptions raised in LobbyClient ([#898](https://github.com/boardgameio/boardgame.io/pull/898))
 
 #### Bugfixes
--- a/docs/documentation/CHANGELOG.md
+++ b/docs/documentation/CHANGELOG.md
@@ -364,6 +364,8 @@ See [the Server reference page](https://boardgame.io/documentation/#/api/Server)
 
 #### Breaking Changes
 
+Putting non-serializable values in `G` — class instances, `Map`/`Set`, `Date`, functions — now throws in development. The check runs on the state a move returns or mutates, and also on state produced by `setup` and by turn/phase hooks. Convert such values to plain objects or arrays before storing them in `G`. Production builds are unaffected.
+
 * [[01c522c](https://github.com/boardgameio/boardgame.io/commit/01c522c)] Throw error in development if non-serializable state is used in a move ([#896](https://github.com/boardgameio/boardgame.io/pull/896))
 
 #### Features


docs/documentation/CHANGELOG.md
```diff
@@ -362,9 +362,12 @@ See [the Server reference page](https://boardgame.io/documentation/#/api/Server)
 
 ### v0.43.3
 
-#### Features
+#### Breaking Changes
 
 * [[01c522c](https://github.com/boardgameio/boardgame.io/commit/01c522c)] Throw error in development if non-serializable state is used in a move ([#896](https://github.com/boardgameio/boardgame.io/pull/896))
```
devill, line 369: All 13 other `Breaking Changes` sections in this changelog open with a sentence describing the break before the commit list (v0.45.0, v0.40.0 and v0.38.0 are good examples). Without it, a reader hitting this section mid-upgrade has to open PR #896 to find out whether it affects them.

```suggestion
#### Breaking Changes

Moves that return non-serializable state — class instances, `Map`/`Set`, functions, `Date` — now throw in development. Convert such values to plain objects or arrays before storing them in `G`.

* [[01c522c](https://github.com/boardgameio/boardgame.io/commit/01c522c)] Throw error in development if non-serializable state is used in a move ([#896](https://github.com/boardgameio/boardgame.io/pull/896))
```

(Wording taken from `src/plugins/plugin-serializable.ts` — `isSerializable` accepts only primitives, plain objects and arrays, and the throw is gated on `NODE_ENV !== 'production'`.)

devill (changes requested):
Thanks for picking this up — the reclassification itself is right, and splitting the two v0.43.3 entries into separate sections is exactly the correct shape.

One thing to add before this lands: a line saying what actually breaks. Every other `Breaking Changes` section in this file has one, and it's the part a reader upgrading through v0.43.3 is looking for. Suggestion inline.

docs/documentation/CHANGELOG.md
Rupesh-ark: Applied in 92280e2 — your structure and value list, with the scope widened after checking the plugin.

Two things the "moves that return" phrasing misses:

- `FnWrap` composes serializable *outside* immer (`CORE_PLUGINS = [Immer, Random, Log, Serializable]`, reduced in order), so it sees the produced `G`. A move that only mutates — `G.m = new Map()`, no return — throws as well.
- `FnWrap` is applied to hooks (`flow.ts:87`) as well as moves (`game.ts:93`), so state from `setup` and turn/phase hooks is checked too, at client creation.

Confirmed each case throws `Move state is not JSON-serialiazable.`: mutated `Map`, `Date`, class instance, a `turn.onBegin` returning a `Set`, and a `setup` returning a `Set`.

Final text:

> Putting non-serializable values in `G` — class instances, `Map`/`Set`, `Date`, functions — now throws in development. The check runs on the state a move returns or mutates, and also on state produced by `setup` and by turn/phase hooks. Convert such values to plain objects or arrays before storing them in `G`. Production builds are unaffected.

Prettier clean, branch updated to main.

devill (approved):

PR #1278, opened 2026-07-20
fix(lobby): broadcast player metadata updates

### LINKED ISSUE #877: Lobby.updatePlayer does not propagate changes
It looks like there is no mechanism to send an update when the player data changes via the lobby. Calling updatePlayer() does not cause any updates to the clients.

As a workaround, I discovered that the changes do propagate when a player connects/disconnects so now I just execute:
```
    ...
    lobbyClient.updatePlayer(gameName, matchID, { playerID, credentials, newName });
    this.client.stop();
    this.client.start();
    ...
```

### LINKED ISSUE #1079: Problem with the leave game
### Hello guys, i have a problem with a game. When i win, my opponent   leaves   from the game, then if he wants come back and join to the last game, matchData return only  *id* and *isConnected*  for the second player
```javascripts
[{"id": 0, "isConnected": true, "name": "FirstPlayer"}, {"id": 1, "isConnected": true}]
```
#### but actually show 
```javascripts
[{"id": 0, "isConnected": false, "name": "FirstPlayer"}, {"id": 1, "isConnected": true,"name": "SecondPlayer"}]
```
#### After that  the game is unplayable, a lot of bugs come out
**Q1:**  How to add *name* to the matchData when its return?
**Q2** Is there another way to solve the problem?

Rupesh-ark:
## Changes

- publish credential-stripped match data after successful lobby joins and player metadata updates
- publish refreshed match data after slot or game leave operations when the match remains
- skip publication when the final player leaves and the match is deleted

Closes #877
Closes #1079

### DIFF
--- a/src/server/api.test.ts
+++ b/src/server/api.test.ts
@@ -99,6 +99,17 @@ describe('.configureRouter', () => {
     return app;
   }
 
+  function createMatchTransport() {
+    const transportAPI = { send: jest.fn(), sendAll: jest.fn() };
+    const transport = {
+      createTransportAPI: jest.fn(() => transportAPI),
+      getMatchQueue: jest.fn(() => ({
+        add: <T>(task: () => PromiseLike<T>) => Promise.resolve(task()),
+      })),
+    };
+    return { transport, transportAPI };
+  }
+
   // A single suite-level server bound explicitly to 127.0.0.1. Letting
   // supertest create a server per request would bind the wildcard address,
   // where the kernel may allocate an ephemeral port that another local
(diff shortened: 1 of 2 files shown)


src/server/api.test.ts
```diff
@@ -1172,11 +1223,13 @@ describe('.configureRouter', () => {
                   };
                 },
               });
-              const app = createApiServer({ db, auth, games });
+              const { transport, transportAPI } = createMatchTransport();
+              const app = createApiServer({ db, auth, games, transport });
```
devill, line 1227: This might be worthy of an extract method to avoid code duplication

devill (approved):
The small refactor suggestion is not worthy of blocking the PR, but please consider it.

PR #1280, opened 2026-07-20
fix: pass the game's ai options to Debug Panel and Local bots

Rupesh-ark:
The Debug Panel constructed its bots with only `client.game.ai.enumerate`, dropping any other options defined on `game.ai`, reported for `objectives` in #7 (https://github.com/boardgameio/boardgame.io/issues/7#issuecomment-815012624), and the same applied to `iterations` and `playoutDepth` even though the tutorial says the 1000-iteration default "can be configured". `Local({ bots })` built its bots the same narrow way, on the path where bots actually play.

Changes:
- `AI.svelte` now spreads `client.game.ai` into the bot constructor (both on initial mount and when switching bots), with `game` and `iterationCallback` listed after the spread so they can't be overridden. This is the fix delucis suggested in the thread.
- `LocalMaster` does the same for `Local({ bots })`, with `game` and `seed` after the spread. Without it the two paths disagree silently — a bot tuned in the panel plays differently in the real game, which is harder to spot than the original bug (thanks @devill for catching it).
- `Game['ai']` gains an index signature so TypeScript games can actually declare these extra options. Typing them precisely would mean importing MCTSBot types into `types.ts` (circular), so the extra keys stay `unknown`. Worth being explicit about the cost: this also turns off excess-property checking for `ai`, so a typo like `iteratons: 42` used to be a `TS2353` error and now compiles and is dropped silently by the bot's destructuring. Nothing in the current code catches that; a dev-only check on the MCTSBot side would, and covers third-party bots too — follow-up, not this PR.
- New debug-panel test asserting a game with `ai: { iterations: 42, playoutDepth: 7 }` shows those values in the AI tab's option sliders (previously the 1000/50 defaults), plus a Local transport test asserting the bot constructor receives `iterations`, `playoutDepth` and `objectives`.

## Verification

- Each test fails without its own fix: the panel test reads `1000` instead of `42`; the transport test loses all three keys.
- Full suite: 43 suites, 903 tests passing. `tsc --noEmit`, ESLint and Prettier clean on the changed files.

Related: #7 (stays open — this covers the debug-panel and local-play paths from that thread)

### DIFF
--- a/src/client/debug/ai/AI.svelte
+++ b/src/client/debug/ai/AI.svelte
@@ -44,8 +44,8 @@
   let bot;
   if (client.game.ai) {
     bot = new MCTSBot({
+      ...client.game.ai,
       game: client.game,
-      enumerate: client.game.ai.enumerate,
       iterationCallback,
     });
     bot.setOpt('async', true);
@@ -57,8 +57,8 @@
   function ChangeBot() {
     const botConstructor = bots[selectedBot];
     bot = new botConstructor({
+      ...client.game.ai,
       game: client.game,
-      enumerate: client.game.ai.enumerate,
       iterationCallback,
     });
     bot.setOpt('async', true);
--- a/src/client/debug/tests/debug.test.ts
+++ b/src/client/debug/tests/debug.test.ts
@@ -39,6 +39,31 @@ test('switching panels', async () => {
   client.stop();
 });
 
+test('AI panel bot uses the game’s ai options', async () => {
+  const client = Client({
+    game: {
+      moves: { clickCell: () => {} },
+      ai: {
+        enumerate: () => [{ move: 'clickCell', args: [0] }],
+        iterations: 42,
+        playoutDepth: 7,
+      },
+    },
+  });
+  client.start();
+
+  await fireEvent.click(screen.getByText('AI'));
+
+  const iterations = screen.getByLabelText('iterations') as HTMLInputElement;
+  const playoutDepth = screen.getByLabelText(
+    'playoutDepth',
+  ) as HTMLInputElement;
+  expect(iterations.value).toBe('42');
+  expect(playoutDepth.value).toBe('7');
+
+  client.stop();
+});
+
 test('visibility toggle', async () => {
   const client = Client({ game: {} });
   client.start();
--- a/src/client/transport/local.test.ts
+++ b/src/client/transport/local.test.ts
@@ -47,6 +47,33 @@ describe('bots', () => {
     expect(client.getState().ctx.turn).toBe(3);
   });
 
+  test('bots are constructed with the game’s ai options', () => {
+    const botOptions = [];
+    class SpyBot extends RandomBot {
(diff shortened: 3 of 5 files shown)


devill (commented):
## Local AI play seems not to be fixed by this

`Local({ bots })` builds its bots the old way, with only `{ game, enumerate, seed }`:

https://github.com/boardgameio/boardgame.io/blob/a370d416430b731c902c383151ac91df1f45c21c/src/client/transport/local.ts#L73-L77

So after this PR, `ai: { objectives, iterations, playoutDepth }` changes bot behaviour in the Debug Panel but is still silently ignored when bots actually play. Given `objectives` is most of what #7 was asking for, someone will tune their bot in the panel, see it work, and find it playing differently in the real game — which is a harder bug to spot than the one you're fixing.

The same spread covers it:

```js
        initializedBots[playerID] = new bot({
          ...game.ai,
          game,
          seed: game.seed,
        });
```

(`enumerate` arrives via the spread; `game`/`seed` stay after it so they can't be overridden.)

Happy for it to be a follow-up PR if you'd rather keep this one debug-panel-only — I'd just not want the two paths to disagree silently in the meantime.

## On your flagged judgment call

I think you made the right call — `Game` is the extension point; a bot is a plugin. If `types.ts` had to name MCTSBot to type this, the bot would stop being a plugin.

The widened type doesn't buy any safety either — a static check can't catch this. A game author never calls the bot constructor, so their object literal never meets `MCTSBot`'s parameter type; and `AI.svelte`, the one place that does construct from `game.ai`, isn't in the typechecked program at all (`tsc --listFiles` sees zero `.svelte` files).

It could be argued that the check belongs on the MCTSBot side instead, at runtime. The bot already declares every option precisely (`src/ai/mcts-bot.ts:60-80`), but those types are erased at build time, so nothing actually inspects what arrives — an unknown key is dropped by the destructuring, and `iterations: '42'` silently becomes the 1000 default via the `typeof iterations === 'number' ? iterations : 1000` fallback. A dev-only complaint at that point would give the author a real error, keep the core ignorant of bots, and cover third-party bots too — same shape as `plugin-serializable`'s `NODE_ENV !== 'production'` throw.

Nothing for this PR either way — the spread is the fix and it's right. If you think it's worth the hassle, you can open an issue for it.

Rupesh-ark:
Folded the `Local` fix into this PR rather than a follow-up, you're right that the silent divergence is worse than the original bug, and it's the same three lines.

`19e2d8a`: `LocalMaster` spreads `game.ai` with `game`/`seed` after it, exactly as you wrote it. Test added in `local.test.ts` — a spy bot subclass captures its constructor options and asserts `iterations`, `playoutDepth` and `objectives` arrive. It fails without the fix (all three keys missing); the panel test fails the same way (`1000` instead of `42`). Full suite 43 suites / 903 tests, tsc + lint clean, branch updated to main.

devill (approved):

PR #1287, opened 2026-07-26
fix(build): compile dynamic imports in the jest babel env

devill:
## Problem

`babel.config.js` runs `@babel/preset-env` with `modules: false` and, for the `test` env, adds only `@babel/plugin-transform-modules-commonjs`. That plugin rewrites static `import`/`export`, but it leaves `import()` expressions alone — so under Jest (which runs CommonJS) any dependency containing a dynamic import falls through to Node's ESM loader and throws:

```
TypeError: A dynamic import callback was invoked without --experimental-vm-modules
```

This is latent on `main` today. It becomes visible in #1286, where `@testing-library/svelte` 5.4.2 moved its wrapper scaffold behind a dynamic import — 8 debug-panel tests fail on both Node 24 and 26.

## Fix

Add `@babel/plugin-transform-dynamic-import` alongside the CommonJS transform in the `test` env, so `import()` compiles to `Promise.resolve().then(() => require(...))`. Pinned to the `^7` line to match the rest of the Babel toolchain (the plugin's latest is `8.x`, which peer-depends on `@babel/core@^8`).

This is deliberately *not* `NODE_OPTIONS=--experimental-vm-modules`: that would switch the whole suite to Jest's experimental ESM runtime, a far bigger change with much wider blast radius than the actual defect.

## Test

`babel.config.test.js` asserts a dynamic import resolves inside the Jest environment. Verified red before the config change and green after.

## Verification

Run locally against this branch:

- `pnpm run test:coverage` — 43 suites / 891 tests pass, coverage unchanged at 100%
- `pnpm run build` — pass
- `pnpm run ts` — pass
- `pnpm run test:integration` — pass

Also cherry-verified on top of #1286's branch: all 884 tests pass there with this change applied, including the 8 that currently fail.

## Note

The commit is unsigned — the SSH signing key was passphrase-locked and unavailable. The squash merge itself will be signed by GitHub.

https://claude.ai/code/session_01BRtvKfpVqgRdDrcY4x6zVL


Rupesh-ark (commented):
LGTM, since it's mostly updates, imma approve

devill:
Please submit the review so this can be merged :)

Rupesh-ark (approved):

PR #1292, opened 2026-07-26
docs(agents): note that pnpm settings live in pnpm-workspace.yaml

devill:
`overrides` and dependency pins moved to `pnpm-workspace.yaml` in #1273.

If a `pnpm` block also exists in `package.json`, pnpm reads `package.json` and ignores the workspace file entirely — every existing pin is silently dropped, with no warning. Worth writing down so it isn't rediscovered the hard way.


Rupesh-ark (approved):
LGTM

PR #1295, opened 2026-07-26
docs: ask for one commit per concern, and route agents to CONTRIBUTING.md

devill:
Two related gaps in the contributor docs.

**`CONTRIBUTING.md` said nothing about commit structure.** Bundled commits are slow to review, and the request to re-cut a branch into single-purpose commits currently has to be made by hand on each PR. Writing it down makes it the repo's expectation instead of a recurring ask. Re-cutting also works as a review pass in its own right, which matters most on agent-authored branches.

**`AGENTS.md` never referenced `CONTRIBUTING.md`.** It described the toolchain in detail but not how changes are meant to land, so an agent had no reason to look for the branch, testing, or PR rules. It now points at `CONTRIBUTING.md` and summarises it in a sentence.


Rupesh-ark (approved):

Rupesh-ark:
Merging this, as the change you proposed makes sense.

PR #1296, opened 2026-07-27
fix(log): preserve metadata from lifecycle hooks

### LINKED ISSUE #1228: Log plugin does not work on hooks, only on moves

janmarsino98:
## Summary

- attach log metadata from lifecycle hooks to the move or event that triggered them
- handle hooks reached through queued and automatic events, including turn, phase, and game endings
- preserve synthetic event arguments and player IDs while stripping credentials
- document hook metadata behaviour and add regression coverage

Fixes #1228.
Builds on the direct-event support added in #1267.

## Testing

- `corepack pnpm exec jest src/core/flow.test.ts src/plugins/plugin-log.test.ts --runInBand` (150 tests)
- `corepack pnpm exec eslint src/core/flow.ts src/core/reducer.ts src/plugins/plugin-log.ts src/plugins/plugin-log.test.ts`
- `corepack pnpm run ts`
- `corepack pnpm run build`

#### Checklist

- [x] Regression coverage was added for all changed paths. The full pre-push coverage command could not start locally because the Windows environment exposes pnpm through Corepack while the hook invokes `pnpm` directly; CI will run the full coverage suite.

### DIFF
--- a/docs/documentation/debugging.md
+++ b/docs/documentation/debugging.md
@@ -36,8 +36,8 @@ const client = Client({
 
 ### Custom metadata in game logs
 
-It can sometimes be helpful to surface some metadata during a move.
-You can do this by using the log plugin. For example,
+It can sometimes be helpful to surface metadata during a move or a lifecycle
+hook. You can do this by using the log plugin. For example,
 
 ```js
 const move = ({ log }) => {
@@ -46,7 +46,9 @@ const move = ({ log }) => {
 ```
 
 This metadata is stored in the `log` client property and displayed
-in the Log section of the debug panel.
+in the Log section of the debug panel. Metadata set in `onBegin`, `onEnd`, or
+`onMove` is attached to the log entry for the move or event that triggered the
+hook. If multiple hooks update the same entry, the last metadata set is used.
 
 ### Redux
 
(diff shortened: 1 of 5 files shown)


devill (commented):
Fixes #1228 — metadata from a phase `onBegin` now lands on the triggering entry. Three things
first:

- **Coverage is red** (−0.2%, 99.821%), so the "coverage added for all changed paths" box isn't
  true. Every gap is new dead code: `flow.ts:544,864,885,891` and `reducer.ts:183-186`
  (`clearLogMetadata` is unreachable now). Delete rather than test.
- **Undeclared change:** an event dispatched inside a move now gets its own log entry.
  `events.endGame({ winner: '0' })` in a move gives `["MOVE:win"]` on `main` and
  `["MOVE:win","EVENT:endGame"]` here. Not in the description; all 906 tests pass either way.
- **This replaces #1267 rather than building on it** — the ~86 lines dropped from `reducer.ts`
  are your `107604e`. Worth saying so, it's a second pass over the same ground.

PR #1298, opened 2026-07-30
Prioritize explicit turn endings

### LINKED ISSUE #924: moveLimit overrides the next parameter in endTurn
To reproduce, create a game with this in its declaration:
```
	turn: 
	{
		moveLimit: 1,
	},

```

And then in a move, fire the endTurn event with a `next` argument:
```
		ctx.events?.endTurn?.( { next: '2' } );
```

Play will proceed to the next player in the play order instead of the next player specified in endTurn. I would guess the endTurn event fired from the move is actually doing nothing, but I don't know for sure. If I remove the moveLimit from the game it works as expected, so maybe the bug here is that endTurn doesn't report any kind of error to indicate that it's not going to do anything?

janmarsino98:
Fixes #924

An explicit `endTurn` or `pass` queued by a move now takes precedence over the automatic `maxMoves` / legacy `moveLimit` turn end. This preserves arguments such as `endTurn({ next: '2' })` instead of advancing through the default play order.

The automatic turn end previously ran before the queued event was flushed. By the time that event was processed, its captured turn no longer matched and its `next` argument was discarded.

Tests:
- focused regression suites: 142 tests passed
- `pnpm run lint`
- `pnpm test`: 43 suites passed, 903 tests passed, 1 todo
- `pnpm run test:coverage`
- `pnpm run ts`
- `pnpm run build`

### DIFF
--- a/src/core/flow.test.ts
+++ b/src/core/flow.test.ts
@@ -427,6 +427,21 @@ describe('turn', () => {
         expect(state.ctx.turn).toBe(2);
         expect(state.ctx.currentPlayer).toBe('1');
       });
+
+      test('explicit endTurn takes precedence over the automatic turn end', () => {
+        const game: Game = {
+          moves: {
+            endTurn: ({ events }) => events.endTurn({ next: '2' }),
+          },
+          turn: { maxMoves: 1 },
+        };
+        const client = Client({ game, numPlayers: 3 });
+
+        client.moves.endTurn();
+
+        expect(client.getState().ctx.turn).toBe(2);
+        expect(client.getState().ctx.currentPlayer).toBe('2');
+      });
     });
 
     describe('with phases', () => {
--- a/src/core/flow.ts
+++ b/src/core/flow.ts
@@ -276,7 +276,12 @@ export function Flow({
       }
 
       // Check if we should end the turn.
-      if ([OnMove, UpdateStage, UpdateActivePlayers].includes(fn)) {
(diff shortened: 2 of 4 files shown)


devill (commented):
Fixes #924's repro, but it also switches off `turn.endIf`.

`Events.dispatch` is never cleared, so `isTurnEndPending()` stays true after the queued `endTurn`
has been consumed — every later `OnMove` / `UpdateStage` / `UpdateActivePlayers` in that action
skips `ShouldEndTurn`, which gates `endIf` and not just `maxMoves`. No `maxMoves` here; passes on
`main`, `Received: 2` on this branch:

```ts
const client = Client({
  game: {
    setup: () => ({}),
    turn: {
      onBegin: ({ ctx, events }) => {
        if (ctx.turn === 2) events.setActivePlayers({ all: Stage.NULL });
      },
      endIf: ({ ctx }) => ctx.turn === 2,
    },
    moves: { A: ({ events }) => { events.endTurn(); } },
  },
  numPlayers: 3,
});
client.moves.A();
expect(client.getState().ctx.turn).toBe(3);
```

Could you scope `isTurnEndPending()` to the entries `update()` hasn't reached yet, and add this
as a regression test?

Worth settling first: `flow.ts:280` is the only `_private` use outside `src/plugins/`, and the
`?? false` fallback silently restores the old behaviour if that api is ever absent.

PR #1309, opened 2026-07-31
Select the configured active player

### LINKED ISSUE #1038: AI framework automatically picks first active player
When stages were introduced (https://github.com/boardgameio/boardgame.io/commit/da2f0eaadfe53701fa62370578bcb1c9ee5a1a7d)

the Step function (back then in client.js) received the following update:

```js
      let playerID = ctx.currentPlayer;
      if (ctx.stage) {
        playerID = Object.keys(ctx.stage)[0];
      }
```

This results in the `Step`-Function always picking the first player that is currently active, instead of picking the `currentPlayer`, or, a player of the developers choice.

What's the reasoning behind this? Wouldn't it make more sense to pick the currentPlayer, or allow the caller to declare for which player (probably the bot player) the step should be executed?

In my use case this currently makes it impossible for me to make the bot player player 2, as it will not execute any moves, as player 1 is also able to make moves in the enemy turn (but mostly decides not to do it).

janmarsino98:
Fixes #1038

`Step` now accepts an explicit active player. When `Simulate` receives a map of player controllers, it selects an active player that has a matching configuration and stops cleanly when none is available.

Previously, both paths always selected the first key in `ctx.activePlayers`. That ignored the player requested by the caller and could choose an entry missing from the supplied controller map.

Tests:
- focused regression suite: 23 tests passed
- `pnpm run lint`
- `pnpm test`: 43 suites passed, 904 tests passed, 1 todo
- `pnpm run test:coverage`: 100% statements, branches and lines
- `pnpm run ts`
- `pnpm run build`

### DIFF
--- a/src/ai/ai.test.ts
+++ b/src/ai/ai.test.ts
@@ -253,6 +253,19 @@ describe('Simulate', () => {
 
     expect(endState.G.movedBy).toBe('2');
   });
+
+  test('stops when no bot is configured for the active player', async () => {
+    const state = InitializeGame({ game: TicTacToe });
+
+    const result = await Simulate({
+      game: TicTacToe,
+      bots: {},
+      state,
+      depth: 1,
+    });
+
+    expect(result).toEqual({ state, metadata: null });
+  });
 });
 
 describe('Bot', () => {
--- a/src/ai/ai.test.ts
+++ b/src/ai/ai.test.ts
@@ -140,6 +140,33 @@ describe('Step', () => {
     await Step(client, bot);
     expect(client.getState().G).toEqual({ moved: true });
   });
+
+  test('uses a requested active player', async () => {
+    const enumerate = jest.fn((_G, _ctx, playerID) => [
+      makeMove('A', undefined, playerID),
+    ]);
+    const client = Client({
+      numPlayers: 3,
+      game: {
+        setup: () => ({ movedBy: null }),
+        moves: {
+          A: ({ G, playerID }) => {
+            G.movedBy = playerID;
+          },
+        },
+        turn: {
+          activePlayers: { all: Stage.NULL },
+        },
+        ai: { enumerate },
+      },
+    });
+
+    const bot = new RandomBot({ enumerate });
+    await Step(client, bot, '2');
+
+    expect(enumerate.mock.calls[0][2]).toBe('2');
+    expect(client.getState().G.movedBy).toBe('2');
+  });
 });
 
 describe('Simulate', () => {
@@ -196,6 +223,35 @@ describe('Simulate', () => {
       depth: 1,
     });
     expect(endState.ctx.gameover).not.toBe(undefined);
+  });
+
+  test('selects an active player with a configured bot', async () => {
+    const game = ProcessGameConfig({
+      setup: () => ({ movedBy: null }),
+      moves: {
+        A: ({ G, playerID }) => {
+          G.movedBy = playerID;
+        },
+      },
+      turn: {
+        activePlayers: { all: Stage.NULL },
+      },
+      endIf: ({ G }) => G.movedBy !== null,
+    });
+    const bot = new RandomBot({
+      seed: 'test',
+      enumerate: (_G, _ctx, playerID) => [makeMove('A', undefined, playerID)],
+    });
+
+    const state = InitializeGame({ game, numPlayers: 3 });
+    const { state: endState } = await Simulate({
+      game,
+      bots: { '2': bot },
+      state,
+      depth: 1,
+    });
+
+    expect(endState.G.movedBy).toBe('2');
   });
 });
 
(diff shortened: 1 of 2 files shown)


src/ai/ai.test.ts
```diff
@@ -197,6 +224,48 @@ describe('Simulate', () => {
...
+
+    const result = await Simulate({
+      game: TicTacToe,
+      bots: {},
+      state,
+      depth: 1,
+    });
+
+    expect(result).toEqual({ state, metadata: null });
+  });
```
devill, line 268: **This test doesn't reach the branch it's for.** `TicTacToe` sets only `minMoves`/`maxMoves`, so `ctx.activePlayers` is `null` and the test exits through the `state.ctx.currentPlayer` fallback — `activePlayers.find(...) === undefined` never runs. Branch coverage doesn't catch it because `.find` isn't a branch. (Adding `?? activePlayers[0]` to that `find` leaves all 23 tests green.)

```suggestion
  test('stops when no bot is configured for the active player', async () => {
    const game = ProcessGameConfig({
      setup: () => ({ movedBy: null }),
      moves: {
        A: ({ G, playerID }) => {
          G.movedBy = playerID;
        },
      },
      turn: { activePlayers: { all: Stage.NULL } },
    });
    const state = InitializeGame({ game, numPlayers: 3 });

    const result = await Simulate({ game, bots: {}, state, depth: 1 });

    expect(result).toEqual({ state, metadata: null });
  });
```

src/ai/ai.ts
```diff
@@ -63,12 +80,10 @@ export async function Simulate({
...
   let iter = 0;
   while (state.ctx.gameover === undefined && iter < depth) {
-    let playerID = state.ctx.currentPlayer;
-    if (state.ctx.activePlayers) {
-      playerID = Object.keys(state.ctx.activePlayers)[0];
-    }
+    const playerID = getBotPlayer(state, bots);
 
     const bot = bots instanceof Bot ? bots : bots[playerID];
+    if (!bot) break;
```
devill: **`Simulate` now stops silently where it used to throw.** `Simulate({ game, bots: { '1': bot }, state })` on a game whose current player is `'0'` returns the untouched initial state — indistinguishable from a finished simulation.

```suggestion
    const bot = bots instanceof Bot ? bots : bots[playerID];
    if (!bot) {
      logging.info(`no bot configured for player ${playerID}, stopping`);
      break;
    }
```

with `import * as logging from '../core/logger';` at the top.

Two nits while you're in here: `getBotPlayer` can return `undefined`, so `: PlayerID | undefined` on the signature; and `Simulate`'s JSDoc still calls `bots` "an array of bots" — now that its keys decide who moves, that line is worth correcting.

devill (commented):
**"Fixes #1038" closes an issue this only partly covers**

@delucis' comment on #1038 asks for three things: let the developer control who plays, consolidate the three copies of the selection logic (`GetBotPlayer` in the Local transport, `Step`, `Simulate`), and possibly allow selecting more than one active player. This delivers the first for `Step` and takes the three copies down to two — `Simulate` still has no way for a caller to name the player, and `src/client/transport/local.ts:28` keeps its own copy. (They agree today: both resolve to the lowest player ID present in both `activePlayers` and `bots`.)

Keeping the PR narrow is the right call. Could you switch the body to `Refs #1038` and add a "Remaining limitations" section, the way you did on #1267? Otherwise merging quietly drops the rest of the issue.

src/ai/ai.ts
devill: Following on from this one — what makes the guard awkward is that `bots` is `Bot | Record<PlayerID, Bot>`, told apart by `bots instanceof Bot` at the point of use. That's a prototype check on an abstract class rather than on the data, and routing `Step` through the shared helper spreads it to a second call site.

I've pushed `devill/simulate-bots-map` (`603dd4e`, on top of your `545d64e`) which normalises once at the entry point instead: a single `Bot` becomes a map covering every seat, and everything downstream works on a plain `Record<PlayerID, Bot>`. **The public signature is unchanged** — no caller or test needed touching. One file, +19/−16, 904 tests green, coverage still 100%.

Would you fold it in? Cherry-pick or merge, whichever suits. Happy to drop it if you'd rather keep this PR to the selection fix.

janmarsino98: Thanks — this is a cleaner approach. I've folded 603dd4e into the PR.

devill (commented):
Thanks for folding in `603dd4e`. Three points still open from last round:

- Body still says `Fixes #1038` — needs `Refs`, plus the "Remaining limitations" section
  (`Simulate`'s missing caller-supplied player; the third copy at
  `src/client/transport/local.ts:28`).
- `ai.test.ts:257` still initialises with `TicTacToe`, so the test misses its branch. Confirmed:
  adding `?? Object.keys(state.ctx.activePlayers)[0]` to the `find` leaves all 23 green.
- `ai.ts:89` is still a bare `break`.

The suggestions from last time still apply as-is.

PR #1311, opened 2026-08-01
build(integration): bump vite and vitest, and put the directory in Dependabot's scope

devill:
Closes the four open Dependabot alerts against `integration/`:

| Alert | Package | Severity |
| --- | --- | --- |
| Vitest UI server allows arbitrary file read and execute | vitest | critical |
| `server.fs.deny` bypass on Windows alternate paths | vite | high |
| Path traversal in optimized deps `.map` handling | vite | medium |
| launch-editor NTLMv2 hash disclosure via UNC paths on Windows | vite | medium |

**Why no Dependabot PR existed.** `.github/dependabot.yml` only had entries for `/` and github-actions, so `integration/package.json` was never scanned. Its dependencies had drifted five majors behind and the alerts had nothing to fix them. The second commit adds an entry for the directory, with its own group so the scaffold's updates stay separate from the library's.

**Why current rather than the minimum patched versions.** The alerts are cleared by vite 6.4.3 and vitest 3.2.6, but landing there would just have Dependabot open a stack of major-version PRs as soon as the directory came into scope. `@vitejs/plugin-react` moves to v6 because that is the first release peering with vite 8.

Verified with `pnpm run test:integration`: typecheck, vitest, `vite build`, the CJS and ESM node smoke tests, publint and attw all pass. `pnpm test` (43 suites / 901 tests) and `pnpm run lint` are green.


Rupesh-ark (commented):
LGTM

Rupesh-ark (approved):

PR #1312, opened 2026-08-01
chore(deps): bump nanoid to 6 and p-queue to 9

devill:
Supersedes #1301 and #1306, which both fail because the new majors are ESM-only and nothing in the repo was set up for that. They land together because they need the same jest change and p-queue drags in `p-timeout`, which is ESM-only too.

**Jest.** `transformIgnorePatterns` leaves `node_modules` alone apart from an allowlist, so every suite reaching nanoid or p-queue died on `Jest encountered an unexpected token`. The three packages join the list and babel converts them like the ESM packages already there.

**The p-queue unwrap goes away.** `socketio.ts` carried an interop shim because p-queue 6 shipped CommonJS with `exports.default`, which made `new PQueue()` throw `PQueue is not a constructor` in the ESM server build. v9 is native ESM with a real default export, so both builds get the class directly. I checked this against the built artefacts, not just the source — constructing a queue from `dist/cjs/server.js` and `dist/esm/server.js`.

**The smoke tests get a queue.** They only ever loaded `boardgame.io/server`, so nothing verified that the p-queue default import survives bundling and the unwrap could have been wrong in either direction without CI noticing. With the new assertions in place, pairing them with p-queue 6 reproduces `TypeError: PQueue is not a constructor`; with p-queue 9 they pass.

**`engines.node` moves to `>=22.12`.** `dist/cjs/*.js` `require()`s both packages, and `require()` of an ES module was only unflagged in Node 22.12 — the old `>=22` promised support for 22.0–22.11, where CommonJS consumers would now hit `ERR_REQUIRE_ESM`.

Verified locally: `pnpm run lint`, `pnpm test` (43 suites / 901 tests), `pnpm run build`, and `pnpm run test:integration` end to end.


Rupesh-ark (approved):
LGTM

PR #1313, opened 2026-08-01
chore(deps-dev): remove tempy

devill:
Closes #1303 by removing the dependency rather than upgrading it.

**Why #1303 can't be merged.** tempy 3 breaks `scripts/changelog.js` twice over: it is ESM-only with a top-level await, so `require('tempy')` throws `ERR_REQUIRE_ASYNC_MODULE`, and `file()` was renamed to `temporaryFile()`. Neither shows up in CI, because nothing there runs the changelog script — it is invoked by hand after `npm version`. The break would land at the next release.

**Why the dependency can go.** tempy was used for exactly one thing: a scratch path to prepend the new release notes to `docs/documentation/CHANGELOG.md`. The script wrote the notes to that temp file, appended the existing changelog to it, then copied the result back. Concatenating in memory and writing once does the same job in one line and needs no temp file at all.

Verified by running `pnpm run changelog` before and after the refactor and diffing the generated `CHANGELOG.md` — byte-identical. `pnpm test` (43 suites / 901 tests) and `pnpm run lint` are green.


Rupesh-ark (dismissed):
LGTM

Rupesh-ark (approved):
Looks good. This keeps the changelog script CommonJS compatible and removes an unnecessary dependency.

PR #1325, opened 2026-08-04
Migrate ESLint to flat config so ESLint 9 can land

### LINKED ISSUE #1324: Migrate to ESLint flat config, so ESLint 9 can land
`pnpm run lint` still runs ESLint 8.57.1, which npm now reports as unsupported on every install (`WARN deprecated eslint@8.57.1: This version is no longer supported`). The config is `.eslintrc` plus `.eslintignore`, and ESLint 9 removed both in favour of `eslint.config.js`.

That is why the Dependabot bump cannot land — #1302 fails CI with:

```
Oops! Something went wrong! :(

ESLint: 9.8.0

ESLint couldn't find an eslint.config.(js|mjs|cjs) file.
```

### Why this is worth doing beyond the one stuck pull request

Staying on ESLint 8 silently freezes the plugins around it, because Dependabot can only offer versions whose peer ranges resolve:

* **`eslint-plugin-unicorn` is stuck at 56.0.1 while 72.0.0 is current.** Recent releases peer `eslint >=10.4`, so no upgrade can even be proposed while ESLint is 8. This is not visible as a stale pull request — Dependabot simply stays silent.
* **`eslint-plugin-react@7.37.5` is already the latest release and peers `^3 || ^4 || ^5 || ^6 || ^7 || ^8 || ^9.7`.** So ESLint 9 is the realistic target here; ESLint 10 stays out of reach until that plugin supports it.

### Scope

Replace `.eslintrc` and `.eslintignore` with a single `eslint.config.js`, porting the existing `extends`, `plugins`, `parser`, `env`, `rules` and `settings`.

The shareable configs currently extended all ship flat-config variants at the versions already installed, so nothing needs upgrading first to start:

| extends | flat equivalent available |
| --- | --- |
| `eslint:recommended` | `@eslint/js` → `js.configs.recommended` |
| `plugin:jest/recommended` | `configs['flat/recommended']` |
| `plugin:unicorn/recommended` | `configs['flat/recommended']` |
| `plugin:react/recommended` | `configs.flat.recommended` |
| `plugin:@typescript-eslint/recommended` | `configs['flat/recommended']` |
| `plugin:prettier/recommended` | `eslint-plugin-prettier/recommended` |

`@eslint/js` is currently only present transitively and would need promoting to a direct devDependency.

`env: {node, browser, es6}` becomes `languageOptions.globals`, and the seven `.eslintignore` entries (`examples/`, `dist/`, `node_modules/`, `coverage/`, `npm/`, `docs/`, `integration/`) become an `ignores` block.

### Done when

* `.eslintrc` and `.eslintignore` are gone, replaced by `eslint.config.js`.
* `pnpm run lint` and `pnpm test` pass on ESLint 9 with no new warnings, and the rule set is unchanged — the same violations are reported before and after.
* #1302 merges, or is closed in favour of the pull request that does the migration.
* The `ignore` rule for ESLint majors in `.github/dependabot.yml` is removed (added in #1323 to stop the unmergeable bump blocking the weekly grouped update).

devill:
Closes #1324.

`pnpm run lint` still ran ESLint 8.57.1, which is no longer supported, and #1302 (the Dependabot bump to ESLint 9) cannot land because ESLint 9 removed `.eslintrc` and `.eslintignore`.

This replaces both with a single `eslint.config.js`, porting the existing `extends`, `plugins`, `parser`, `env`, `rules` and `settings` to their flat-config equivalents at the plugin versions already installed — nothing needed upgrading first. `eslint` moves to `^9`, `@eslint/js` is promoted from a transitive dependency to a direct devDependency, and `globals` is added.

### Behaviour is preserved

`pnpm run lint` reports zero problems both before and after, over the same files:

| | files linted | problems |
| --- | --- | --- |
| ESLint 8.57.1 + `.eslintrc` | 127 (19 `.js`, 99 `.ts`, 9 `.tsx`) | 0 |
| ESLint 9.39.5 + `eslint.config.js` | 129 (20 `.js`, 1 `.cjs`, 99 `.ts`, 9 `.tsx`) | 0 |

No file loses coverage. The two added are `eslint.config.js` itself and `jest-svelte-transformer.cjs` — flat config lints `.cjs` by default, where ESLint 8's `eslint .` did not.

Three deliberate adjustments keep the reported problem set unchanged:

- Flat config lints dotfiles that `.eslintrc` skipped by default, so the jsdom stub `.empty_module.js` is added to `ignores`.
- Flat config defaults `reportUnusedDisableDirectives` to `warn`; it is pinned back to `off`.
- `/* eslint-env es2020 */` is no longer supported, so it is dropped from `jest.setup.js` — those globals now come from `languageOptions.globals`.

### Testing

`pnpm run lint` is green, and `pnpm test` passes 43 suites / 903 tests (1 todo).


Rupesh-ark (approved):
LGTM

